PR #20, opened 2015-12-25
added ClientMessage and Bits

bilalyasar:
client message is written in TypeScript

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,8 @@ hs_err_pid*
 typings/
 npm-debug.log
 
+**/typings/**/*.d.ts
+**/typings/tsd.cached.json
+
+# TypeScript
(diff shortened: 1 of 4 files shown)


asimarslan:
I think we shouldn't commit the compiled codes.

bilalyasar:
yes, updated.

mustafaiman:
I guess `.map` files should be omitted too?

gAmUssA:
let's follow M$ directory structure. `d.ts` files should be placed in `typings` folder.

Here is a reference project https://github.com/Microsoft/vscode/tree/master/src

bilalyasar:
i updated PR.

mustafaiman:
:+1:

PR #23, opened 2015-12-30
initial config api

mustafaiman:
Order as 
```
export class ClientConfig { //
export class NetworkConfig { //
```
Causes use before declaration error in compiled JS code at `jshint` stage although I think it is not an issue in TypeScript. Anyway I re-ordered the classes so that it passes without an error now.


bilalyasar:
:+1:

PR #24, opened 2016-01-04
Map API

mustafaiman:
API design with `put`, `get`, `remove`, `clear`, `size` and `containsKey` operations.
All the operations return promises, therefore they are designed as asynchronous operations.

Part of https://github.com/hazelcast-incubator/hazelcast-nodejs-client/issues/18

### DIFF
--- a/src/proxy/BaseProxy.ts
+++ b/src/proxy/BaseProxy.ts
@@ -1,4 +1,5 @@
 import {SerializationService} from '../serialization/SerializationService';
+import {Data} from '../serialization/Data';
 export class BaseProxy {
 
     protected client : any;
@@ -12,11 +13,11 @@ export class BaseProxy {
         this.serviceName = serviceName;
     }
 
-    protected toData(object : any) : Buffer {
+    protected toData(object : any) : Data {
         return this.client.serializationService.toData(object);
     }
 
-    protected toObject(buffer : Buffer) : any {
-        return this.client.serializationService.toObject(buffer);
+    protected toObject(data : Data) : any {
+        return this.client.serializationService.toObject(data);
     }
 }
--- a/src/proxy/BaseProxy.ts
+++ b/src/proxy/BaseProxy.ts
@@ -1,0 +1,22 @@
+import {SerializationService} from '../serialization/SerializationService';
+export class BaseProxy {
+
+    protected client : any;
+    private name : string;
+    private serviceName : string;
+
+    //TODO client should be of type {real client type}
+    constructor(client : { serializationService : SerializationService}, serviceName: string, name: string) {
+        this.client = client;
+        this.name = name;
+        this.serviceName = serviceName;
+    }
+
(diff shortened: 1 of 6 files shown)


src/proxy/BaseProxy.ts
```diff
@@ -0,0 +1,22 @@
...
+    private serviceName : string;
+
+    //TODO client should be of type {real client type}
+    constructor(client : { serializationService : SerializationService}, serviceName: string, name: string) {
+        this.client = client;
+        this.name = name;
+        this.serviceName = serviceName;
+    }
+
+    protected toData(object : any) : Buffer {
```
GhostInAMachine: What's the purpose of this method?

mustafaiman: All proxies will extend from this class. This is to serialize the given object before sending it to Hazelcast server. This is only trivial for now because we will use JSON serialization. However this abstraction will be neccesaary when we aim for Hazelcast serialization.

GhostInAMachine: Okay, I see. Looks good to me.

asimarslan:
I think we should consider the other forms of each method like putWithTTL. Instead of a new func name, we may consider optional parameters. WDYT?
ex:
`put(key: K, value: V, ttl?:long) : Promise<V>;`

mustafaiman:
Sure, I can update this PR with optional value of TTL for `put` operation. Also, `remove` can take optional `value` argument.

mustafaiman:
@asimarslan I updated the PR with optional parameters for put and remove operations.
As a side note, typescript compiler does not fail if you do not implement a method with optional parameters. All TS cares about is there is at least one implementation for a method name.
Is this ok to merge now?

src/serialization/SerializationService.ts
```diff
@@ -0,0 +1,11 @@
+export class SerializationService {
+    toData(object: any) : Buffer {
```
asimarslan: We need an interface called Data in this service. Its basicly a buffer wrapper and it is already defined in hazelcast.

mustafaiman: I added `Data` type. I also realized that `BaseProxy` should work with `Data` rather than `Buffer` and fixed it too.

src/proxy/BaseProxy.ts
```diff
@@ -0,0 +1,23 @@
...
+        this.client = client;
+        this.name = name;
+        this.serviceName = serviceName;
+    }
+
+    protected toData(object : any) : Data {
+        return this.client.serializationService.toData(object);
+    }
+
+    protected toObject(buffer : Data) : any {
```
asimarslan: paramter name `data` instead of `buffer` will be better

mustafaiman: sure. changed it.

asimarslan:
:+1:

PR #27, opened 2016-01-20
name conversion byte array to buffer

mustafaiman:
`Buffer` is the byte array for Node.js. Changed method names.


asimarslan:
@mustafaiman Please rebase and merge this. it's a trivial rename

PR #28, opened 2016-01-22
Implemented ConnectionManager and InvocationService

GhostInAMachine:


src/ClientMessage.ts
```diff
@@ -1,206 +1,189 @@
...
- |                       CorrelationId                           |
- |                                                               |
- +---------------------------------------------------------------+
- |                        PartitionId                            |
- +-----------------------------+---------------------------------+
- |        Data Offset          |                                 |
- +-----------------------------+                                 |
- |                      Message Payload Data                    ...
- |                                                              ...
- */
```
asimarslan: This is an important doc, why do you remove it?

src/ClientMessage.ts
```diff
@@ -1,206 +1,189 @@
...
+        this.cursor += BitsUtil.BYTE_SIZE_IN_BYTES;
     }
 
-    appendInt(val: number) {
-        this.buffer.writeInt32LE(val, this.getWriteOffset());
-        this.writeIndex += BitsUtil.INT_SIZE_IN_BYTES;
-    }
-
-    appendLong(value: any) {
+    appendLong(value:any) {
```
asimarslan: If we use bit size in type name then this should be Int64

src/ClientMessage.ts
GhostInAMachine: What is the point of duplicating the information already present in the specification?

src/ClientMessage.ts
GhostInAMachine: I thought about it, but I guess this is a better name since we are using a wrapper for 64-bit integer values, which are not supported natively in javascript.

src/ClientMessage.ts
GhostInAMachine: Plus, the numbers above seem to mark the field length. If so, then this ASCII diagram is outdated: Correlation ID occupies 64 bits now instead of 32.

src/ClientMessage.ts
```diff
@@ -1,206 +1,189 @@
...
 
 import Long = require('long');
 import {BitsUtil} from './BitsUtil';
 
 class ClientMessage {
-    buffer:  Buffer;
-    writeIndex:  number;
-    readIndex:  number;
 
-    constructor(payload_size:  number) {
```
mustafaiman: why don't you use constructor?

src/ClientMessage.ts
```diff
@@ -1,206 +1,189 @@
-///<reference path="../typings/node/node.d.ts" />
```
mustafaiman: Aren't these lines needed for compilation?

src/codec/AuthDecoder.ts
```diff
@@ -0,0 +1,35 @@
+import ClientMessage = require("../ClientMessage");
```
mustafaiman: I think `AuthDecoder` and `AuthEncoder` classes will be replaced by `ClientAuthenticationCodec` in the future. So I suggest combining these two classes into one with name `ClientAuthenticationCodec` with methods `encodeRequest` and `decodeResponse` methods. So generated codec can fit very easily.

src/invocation/ClientConnectionManager.ts
```diff
@@ -0,0 +1,56 @@
...
+class ClientConnectionManager {
+
+    private groupConfig:GroupConfig;
+    private ownerConnection:ClientConnection;
+    private ownerConnectionPicker:OwnerConnectionPicker;
+    private invocationService:InvocationService;
+
+    private ready = Q.defer<ClientConnectionManager>();
+
+    constructor(addresses:Address[], groupConfig:GroupConfig, invocationService:InvocationService) {
```
mustafaiman: `adresses` should be retrieved from `ClientNetworkConfig`. We use only addresses for now but other properties such as `connectionAttemptLimit` can be used later.

src/invocation/OwnerConnectionPicker.ts
```diff
@@ -0,0 +1,34 @@
...
+        if (index >= this.addresses.length) {
+            this.ready.reject("Unable to connect to any of the following addresses " + this.addresses);
+        }
+
+        var currentAddress = this.addresses[index];
+        var clientConnection = new ClientConnection(currentAddress);
+        clientConnection.connect().then((connection:ClientConnection) => {
+            this.ready.resolve(connection);
+        }).catch(() => {
+            this.tryAddress(index + 1);
```
mustafaiman: this method trying next address until it runs out of addresses does not feel right. Maybe we can change method signature to something like `tryAddresses(startingIndex: number)`. WDYT?

src/invocation/ClientConnectionManager.ts
GhostInAMachine: Addresses will be provided by higher level component from ClientNetworkConfig. There is no point in passing the entire object here

mustafaiman:
There are lots of style issues in this PR. When I run `npm test`, `tslint` fails with numerous errors. Please correct them.

src/codec/AuthDecoder.ts
GhostInAMachine: Yes, this is just temporary solution until codec generator is done

src/ClientMessage.ts
GhostInAMachine: No, they are not. "require" calls are sufficient.

src/ClientMessage.ts
asimarslan: I can think of some benefits like;
1. This is source code documentation not spec. So we are showing what we implement. Spec allows to add custom data in a client message using the Data-offset field so the actual implementation may vary. The documentation should display what is actually implemented. 
2. Spec itself may change and add more fields to the client message. So the documentation will help the developers to see what is actually implemented.
3. Documented code is better than non-documented code and give more confidence to community I believe.

For the second comment that the doc is outdated; It is not. Correlation Id occupy two lines which has 32bit each.

Note: We borrowed the above notation from Martin Thompson's Aeron project. 
https://github.com/real-logic/Aeron/blob/master/aeron-client/src/main/java/uk/co/real_logic/aeron/protocol/HeaderFlyweight.java

src/ClientMessage.ts
GhostInAMachine: It appears that Typescript does not support multiple constructors. We need to have a constructor that accepts the buffer for reading received messages.

src/invocation/OwnerConnectionPicker.ts
GhostInAMachine: Can you describe how your suggested method is supposed to operate?

src/invocation/ClientConnectionManager.ts
GhostInAMachine: Disregard my previous comment, you are right

src/invocation/OwnerConnectionPicker.ts
mustafaiman: I was just suggesting to change the name of the method and its argument.

src/ClientMessage.ts
mustafaiman: Are you sure? Because when I disable tslint and run `npm test` on this pull request, it fails with
```
src/ClientMessage.ts(3,23): error TS2307: Cannot find module 'long'.
src/ClientMessage.ts(8,20): error TS2304: Cannot find name 'Buffer'.
src/ClientMessage.ts(13,26): error TS2304: Cannot find name 'Buffer'.
src/ClientMessage.ts(24,24): error TS2304: Cannot find name 'Buffer'.
src/ClientMessage.ts(28,17): error TS2304: Cannot find name 'Buffer'.
src/ClientMessage.ts(125,25): error TS2304: Cannot find name 'Buffer'.
src/ClientMessage.ts(178,21): error TS2304: Cannot find name 'Buffer'.
src/ClientMessage.ts(181,26): error TS2304: Cannot find name 'Buffer'.
src/invocation/ClientConnection.ts(1,22): error TS2307: Cannot find module 'net'.
src/invocation/ClientConnection.ts(20,30): error TS2304: Cannot find name 'Buffer'.
src/invocation/ClientConnection.ts(34,20): error TS2304: Cannot find name 'Buffer'.
src/invocation/ClientConnectionManager.ts(32,61): error TS2304: Cannot find name 'Buffer'.
src/invocation/InvocationService.ts(19,37): error TS2304: Cannot find name 'Buffer'.
...
```
When I put `///<reference ...` comments back, it compiles.

src/ClientMessage.ts
mustafaiman: Sure, I get it now.

src/invocation/OwnerConnectionPicker.ts
```diff
@@ -0,0 +1,34 @@
...
+    constructor(addresses: Address[]) {
+        this.addresses = addresses;
+    }
+
+    public pick(): Q.Promise<ClientConnection> {
+        this.tryAddress(0);
+        return this.ready.promise;
+    }
+
+    private tryAddress(index: number) {
```
bilalyasar: i think we can change name of this method as @mustafaiman said.

src/ClientMessage.ts
bilalyasar: i talked with mustafa, we can remove files attribute from gulp config file

bilalyasar:
other than minor comments, :+1:

src/ClientMessage.ts
```diff
@@ -26,53 +24,65 @@ import Long = require('long');
...
     }
 
     getCorrelationId(): number {
         return this.buffer.readUInt32LE(BitsUtil.CORRELATION_ID_FIELD_OFFSET);
     }
 
-    setCorrelationId(val: number) {
-        this.buffer.writeUInt32LE(val, BitsUtil.CORRELATION_ID_FIELD_OFFSET);
+    setCorrelationId(value: number) {
+        this.buffer.writeUInt32LE(value, BitsUtil.CORRELATION_ID_FIELD_OFFSET);
```
asimarslan: correlationId is an int64 value, is this correct ?

src/ClientMessage.ts
```diff
@@ -26,53 +24,65 @@ import Long = require('long');
...
+    constructor(buffer: Buffer) {
+        this.buffer = buffer;
+    }
+
+    getBuffer(): Buffer {
+        return this.buffer;
     }
 
     getCorrelationId(): number {
         return this.buffer.readUInt32LE(BitsUtil.CORRELATION_ID_FIELD_OFFSET);
```
asimarslan: correlation should be Long and returned accordingly

src/ClientMessage.ts
```diff
@@ -26,53 +24,65 @@ import Long = require('long');
...
 
-    constructor(payload_size:  number) {
-        this.buffer = new Buffer(payload_size);
-        this.buffer.fill(0, 0, BitsUtil.HEADER_SIZE + payload_size);
-        this.setDataOffset(BitsUtil.HEADER_SIZE);
-        this.setVersion(BitsUtil.VERSION);
-        this.writeIndex = 0;
-        this.readIndex = 0;
+    private buffer: Buffer;
+    private cursor: number = BitsUtil.HEADER_SIZE;
```
asimarslan: As far as I see, there is another cursor in BufferWrapper class. One of them may be redundent

mustafaiman: BufferWrapper.js is from old code. It is not used anymore.

GhostInAMachine: Mustafa is correct

asimarslan:
:+1:

mustafaiman:
:+1:

PR #31, opened 2016-02-03
added codecs

bilalyasar:


src/codec/ClientCreateProxyCodec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+
+
+
+static calculateSize(name, serviceName, target){
+    // Calculates the request payload size
+    var dataSize = 0;
+    dataSize += BitsUtil.calculateSizeStr(name);
+    dataSize += BitsUtil.calculateSizeStr(serviceName);
+    dataSize += BitsUtil.calculateSizeCom.hazelcast.nio.address(target);
```
mustafaiman: I guess this should be `BitsUtil.calculateSize(target).

src/codec/ClientCreateProxyCodec.ts
```diff
@@ -0,0 +1,49 @@
...
+    return dataSize;
+}
+
+static encodeRequest(name, serviceName, target){
+    // Encode request into clientMessage
+    var payloadSize;
+    var clientMessage = new ClientMessage(payloadSize=this.calculateSize(name, serviceName, target));
```
mustafaiman: There is no `ClientMessage(size: number)`. This should be `newClientMessage(...)`

asimarslan:
@bilalyasar , This PR is very big and github cannot display it. Can you split it into some logical parts like Map, Queue List, Set. Can make sure we can review it here.

asimarslan:
And do not add any codec that Python client do not have like Cache, MapReduce. Not all distributed object supported by non-java clients

bilalyasar:
@asimarslan , ok will split them.

PR #32, opened 2016-02-05
added map codecs

bilalyasar:


src/codec/MapAddEntryListenerCodec.ts
```diff
@@ -0,0 +1,78 @@
...
+    return dataSize;
+}
+
+static encodeRequest(name, includeValue, listenerFlags, localOnly){
+    // Encode request into clientMessage
+    var payloadSize;
+    var clientMessage = new ClientMessage(payloadSize=this.calculateSize(name, includeValue, listenerFlags, localOnly));
+    clientMessage.setMessageType(REQUEST_TYPE);
+    clientMessage.setRetryable(RETRYABLE);
+    clientMessage.appendStr(name);
```
asimarslan: ClientMessage api is not compatible with the generated code. Please validate them. 
Example: appendStr --> appendString  etc

src/codec/MapPutCodec.ts
```diff
@@ -0,0 +1,57 @@
+import {ClientMessage} from "../ClientMessage";
```
mustafaiman: ClientMessage is imported like `import ClientMessage = require('..ClientMessage');`

src/codec/MapAddEntryListenerCodec.ts
```diff
@@ -0,0 +1,78 @@
...
+    dataSize += BitsUtil.BOOLEAN_SIZE_IN_BYTES;
+    dataSize += BitsUtil.INT_SIZE_IN_BYTES;
+    dataSize += BitsUtil.BOOLEAN_SIZE_IN_BYTES;
+    return dataSize;
+}
+
+static encodeRequest(name, includeValue, listenerFlags, localOnly){
+    // Encode request into clientMessage
+    var payloadSize;
+    var clientMessage = new ClientMessage(payloadSize=this.calculateSize(name, includeValue, listenerFlags, localOnly));
```
mustafaiman: should be `ClientMessage.newClientMessage( ... )`

src/codec/MapPutCodec.ts
```diff
@@ -0,0 +1,57 @@
...
+    clientMessage.appendLong(ttl);
+    clientMessage.updateFrameLength();
+    return clientMessage;
+}
+
+static decodeResponse(clientMessage){
+    // Decode response from client message
+    var parameters;
+
+if(clientMessage.readBool() === true){
```
mustafaiman: this part is generated wrong. should be `clientMessage.readBoolean() === false`

src/codec/MapPutCodec.ts
```diff
@@ -0,0 +1,59 @@
...
+    clientMessage.updateFrameLength();
+    return clientMessage;
+}
+
+static decodeResponse(clientMessage : ClientMessage){
+    // Decode response from client message
+    var parameters :any ;
+
+if(clientMessage.readBoolean() === true){
+    parameters['response'] = clientMessage.readBuffer();
```
mustafaiman: after discussion we agreed to return deserialized objects from codecs. see https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/protocol/codec/map_put_codec.py#L42

So decodeResponse should take a function as a parameter and deserialize the object before returning.

src/codec/MapPutCodec.ts
```diff
@@ -0,0 +1,59 @@
...
+    clientMessage.appendBuffer(value);
+    clientMessage.appendLong(threadId);
+    clientMessage.appendLong(ttl);
+    clientMessage.updateFrameLength();
+    return clientMessage;
+}
+
+static decodeResponse(clientMessage : ClientMessage){
+    // Decode response from client message
+    var parameters :any ;
```
mustafaiman: Should initialize this like `parameters: any = {}`. Otherwise `parameters['response'] = ...` fails.

src/codec/MapPutCodec.ts
```diff
@@ -0,0 +1,59 @@
...
+    var dataSize : number = 0;
+    dataSize += Utils.calculateSizeString(name);
+    dataSize += Utils.calculateSizeBuffer(key);
+    dataSize += Utils.calculateSizeBuffer(value);
+    dataSize += BitsUtil.LONG_SIZE_IN_BYTES;
+    dataSize += BitsUtil.LONG_SIZE_IN_BYTES;
+    return dataSize;
+}
+
+static encodeRequest(name : string, key : Buffer, value : Buffer, threadId : any, ttl : any){
```
mustafaiman: I dont know if it makes much of a difference but threadId and ttl can be declared as `number`.

src/codec/MapPutCodec.ts
```diff
@@ -0,0 +1,59 @@
...
+import {MapMessageType} from './MapMessageType';
+
+var REQUEST_TYPE = MapMessageType.MAP_PUT;
+var RESPONSE_TYPE = 105;
+var RETRYABLE = false;
+
+
+export class MapPutCodec{
+
+constructor() {
```
mustafaiman: We do not need constructors.

src/codec/MapPutCodec.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+
+
+
+static calculateSize(name : string , key : Buffer , value : Buffer , threadId : any , ttl : any ){
+    // Calculates the request payload size
+    var dataSize : number = 0;
+    dataSize += Utils.calculateSizeString(name);
```
mustafaiman: It is better if we use `BitsUtil.calculateSizeString` and move that method to BitsUtil

src/codec/MapAddEntryListenerCodec.ts
```diff
@@ -0,0 +1,82 @@
...
+        clientMessage.setRetryable(RETRYABLE);
+        clientMessage.appendString(name);
+        clientMessage.appendBoolean(includeValue);
+        clientMessage.appendInt32(listenerFlags);
+        clientMessage.appendBoolean(localOnly);
+        clientMessage.updateFrameLength();
+        return clientMessage;
+    }
+
+    static decodeResponse(clientMessage:ClientMessage, toObjectFunction:(data:Data) => any) {
```
mustafaiman: `toObjectFunction` is not used here. Also, in many other places in this pr.

asimarslan: not all messages have data to deserialize. but the function signature should be same

mustafaiman:
:+1:

src/codec/MapClearCodec.ts
```diff
@@ -0,0 +1,41 @@
...
+        var clientMessage = ClientMessage.newClientMessage(this.calculateSize(name));
+        clientMessage.setMessageType(REQUEST_TYPE);
+        clientMessage.setRetryable(RETRYABLE);
+        clientMessage.appendString(name);
+        clientMessage.updateFrameLength();
+        return clientMessage;
+    }
+
+    static decodeResponse(clientMessage:ClientMessage, toObjectFunction:(data:Data) => any) {
+        // Decode response from client message
```
asimarslan: There are some decodeResponse functions that should be removed like this one.
We still put a comment for it to show that it is there but removed as no one will call it.
```
//static decodeResponse(clientMessage:ClientMessage, toObjectFunction:(data:Data) => any)
```

src/codec/MapEntrySetCodec.ts
```diff
@@ -0,0 +1,49 @@
...
+        // Decode response from client message
+        var parameters:any = {};
+        var responseSize = clientMessage.readInt32();
+        var response:any = [];
+        for (var responseIndex = 0; responseIndex <= responseSize; responseIndex++) {
+            var responseItem = clientMessage.readMapEntry();
+
+            response.push(responseItem)
+        }
+        parameters['response'] = response;
```
asimarslan: you need to deserialize the result here. You need something like a custom List/array implementation that deserialize the item if accessed so you won't deserialize everything

asimarslan:
:+1:

PR #33, opened 2016-02-09
Map proxy

mustafaiman:
Adds HazelcastClient class and connection initiation logic.
Adds map proxy creation.
Adds Map#put operation.
Adds map put test.
Partition number is hard-coded 271. This will be fixed.
Intoduces basic JsonSerialization to use in map put operation.
Removes old tests and proxy codes.
I fixed generated codecs and used them. They can be replaced by generated codecs when generated codecs are ready.


src/proxy/BaseProxy.ts
```diff
@@ -1,24 +1,31 @@
...
-    //TODO client should be of type {real client type}
-    constructor(client: { serializationService : SerializationService}, serviceName: string, name: string) {
+    constructor(client: HazelcastClient, serviceName: string, name: string) {
         this.client = client;
         this.name = name;
         this.serviceName = serviceName;
     }
 
+    protected invokeWithPartitionId(clientMessage: ClientMessage, key: any): Q.Promise<ClientMessage> {
+        var partitionId: number = Math.abs(this.client.getSerializationService().toData(key).getPartitionHash()) % 271;
```
asimarslan: PartitionService should handle this. Maybe just an empty one can be better to start with.

mustafaiman: I will abstract this one to a dummy partition service for this pr.

src/HazelcastClient.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.proxyManager = new ProxyManager(this);
+        this.partitionService = new PartitionService(this);
+        this.connectionManager = new ClientConnectionManager(this.config.networkConfig,
+            this.config.groupConfig, this.invocationService);
+    }
+
+    private init(): Q.Promise<HazelcastClient> {
+        var deferred = Q.defer<HazelcastClient>();
+
+        Q.all([
```
GhostInAMachine: Why are you using "all" here?

mustafaiman: Client should wait for all services to be initialised. You can add other services here in future.

src/proxy/BaseProxy.ts
```diff
@@ -1,24 +1,49 @@
...
+        var deferred: Q.Deferred<T> = Q.defer<T>();
+        var toObject = this.toObject.bind(this);
+        promise.then(function(clientMessage: ClientMessage) {
+            var parameters: any = codec.decodeResponse(clientMessage, toObject);
+            deferred.resolve(parameters.response);
+        });
+        return deferred.promise;
+    }
+
+    protected encodeInvokeOnKey<T>(codec: any, key: any, ...codecArguments: any[]): Q.Promise<T> {
```
GhostInAMachine: What is the point of having both invokeOnKey and invokeOnPartition? Key is mapped to partition key anyway.

Also, I think request should be encoded outside of the scope of this method

src/proxy/BaseProxy.ts
```diff
@@ -1,24 +1,49 @@
...
+        var partitionId: number = this.client.getPartitionService().getPartitionId(key);
+        return this.encodeInvokeOnPartition<T>(codec, partitionId, ...codecArguments);
+    }
+
+    protected encodeInvokeOnPartition<T>(codec: any, partitionId: number, ...codecArguments: any[]): Q.Promise<T> {
```
GhostInAMachine: Why can't you encode the request at the proxy level and pass the raw bytes here? Passing codec class with arguments here confuses me really.

src/proxy/BaseProxy.ts
mustafaiman: Do you suggest to calculate partition id of a key in proxy code? In that case, we have to put the same calculation code to each method of map(get, put, remove, etc).

GhostInAMachine: No, encode the request at proxy level (e.g. MapProxy), but calculate the partition ID here

mustafaiman: Where encoding requests happen is another issue. I do not think this is related with comment you wrote here.

GhostInAMachine: Sorry, I indeed replied to wrong comment. Basically, you can inline the invokeOnPartition method

mustafaiman: There are operations which are not assigned with a key but are assigned with a partition. You can see these methods exist in other clients, too.

src/proxy/BaseProxy.ts
mustafaiman: We discussed this in Slack channel. I thought we agreed to do it this way.

GhostInAMachine: With whom did you discuss?

src/codec/MapPutCodec.ts
```diff
@@ -0,0 +1,47 @@
...
+import {Data} from '../serialization/Data';
+
+var REQUEST_TYPE = 0x0101;
+var RESPONSE_TYPE = 105;
+
+export class MapPutCodec{
+    static calculateSize(name: string, key: Data, value: Data, threadId: number, ttl: number){
+        // Calculates the request payload size
+        var dataSize = 0;
+        dataSize += SizeUtil.getStringSize(name);
```
mustafaiman: I know this is not related to codec work but can we move `SizeUtil.getStringSize` to `BitsUtil` and change its name to calculateSizeString with this PR.

src/HazelcastClient.ts
GhostInAMachine: OK

src/proxy/BaseProxy.ts
GhostInAMachine: OK

src/proxy/BaseProxy.ts
GhostInAMachine: Issue resolved in Slack

src/codec/MapPutCodec.ts
GhostInAMachine: Yeah, I seen no harm in that. SizeUtil was a temporary class of mine while I was waiting for auto-generated codecs to arrive

mustafaiman: Actually, I was trying to comment on Codec PR about this :) Thanks for approval anyway:)

GhostInAMachine:
Looks good

asimarslan:
I couldn't see the Map proxy optional parameters like put ttl etc.. Why did they removed?

mustafaiman:
@asimarslan They are not removed. Since they are optional parameters in `IMap` Typescript does not require to implement all possible signatures of those methods. I implemented only basic put operation for this PR. I will add other functionality in another PR.

asimarslan:
TypeScript have default values for optional parameters. As our codecs requires all parameters, it can be easier to put these into function signature instead of into codec parameter list.
For example put :
```javascript
function put(key: any, value:any, ttl = -1) {
      ..... MapPutCodec, keyData, valueData, ttl, ...
}
```

mustafaiman:
good point. done.

asimarslan:
:+1:

PR #35, opened 2016-02-23
Implemented smart routing

GhostInAMachine:


src/proxy/BaseProxy.ts
```diff
@@ -25,8 +25,8 @@ export class BaseProxy {
         return deferred.promise;
     }
 
-    protected encodeInvokeOnKey<T>(codec: any, key: any, ...codecArguments: any[]): Q.Promise<T> {
-        var partitionId: number = this.client.getPartitionService().getPartitionId(key);
+    protected encodeInvokeOnKey<T>(codec: any, partitionKey: any, ...codecArguments: any[]): Q.Promise<T> {
+        var partitionId: number = this.client.getPartitionService().getPartitionId(partitionKey);
```
mustafaiman: This is actually not `partitionKey` but `key`. I mean the key from key-value pair. Please revert this naming change.

GhostInAMachine: This does not have to be a key from key-value pair. In non-distributed data structures (List, Set) the structure name is used to determine the partition ID.

asimarslan:
@GhostInAMachine Is this PR also implements, Cluster and partition service ?

GhostInAMachine:
@asimarslan Yeah, it does

src/invocation/InvocationService.ts
```diff
@@ -24,13 +25,16 @@ class InvocationService {
         return deferred.promise;
     }
 
-    public invokeOnPartition(clientMessage: ClientMessage, partitionId: number): Q.Promise<ClientMessage> {
+    invokeOnPartition(clientMessage: ClientMessage, partitionId: number): Q.Promise<ClientMessage> {
         clientMessage.setPartitionId(partitionId);
-        var connection: ClientConnection = this.client.getConnectionManager().getOwnerConnection();
-        return this.invokeOnConnection(connection, clientMessage);
+        var address: Address = this.client.getPartitionService().getAddressForPartition(partitionId);
+        return this.client.getConnectionManager().getOrConnect(address)
```
mustafaiman: invokeOnPartition does not imply sending the request to specific member that holds the partition.I think in our terms `invokeOnPartition` means just setting partitionId accordingly (See https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/invocation.py#L139 ). 
Sending the request to a specific connection or owner connection should be determined later (See https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/invocation.py#L128 )
With these changes, we remove DummyClient option. Am I wrong?

src/proxy/BaseProxy.ts
mustafaiman: ok

src/invocation/InvocationService.ts
GhostInAMachine: I will add the the switch in next commit, stay tuned

mustafaiman: Can you add the switch in this PR? We already have a dummy client but this PR removes the dummy behaviour. We would like to introduce smart client as an addition, not as a replacement.

src/invocation/InvocationService.ts
```diff
@@ -24,13 +25,16 @@ class InvocationService {
...
+            .then<ClientMessage>((connection: ClientConnection) => {
+                return this.invokeOnConnection(connection, clientMessage);
+            });
     }
 
-    public processResponse(buffer: Buffer) {
+    processResponse(buffer: Buffer) {
         var clientMessage = new ClientMessage(buffer);
         var correlationId = clientMessage.getCorrelationId().toNumber();
         this.pending[correlationId].resolve(clientMessage);
```
mustafaiman: We should check if `clientMessage` is an event by checking event flag. If it is not an event, we should remove that entry with `correlationId` from `this.pending`.

PR #40, opened 2016-03-09
Error handling

GhostInAMachine:


src/HazelcastClient.ts
```diff
@@ -44,8 +44,8 @@ class HazelcastClient {
             return this.partitionService.initialize();
         }).then(() => {
             deferred.resolve(this);
-        }).catch(() => {
-            deferred.reject('Client failed to start');
+        }).catch((e) => {
+            deferred.reject(e);
```
mustafaiman: Propagating the error to upper level is already default behaviour. This `.catch` is unnecessary. Correct me if I am wrong.

GhostInAMachine: Before the change in case of the error you would only get a failure message without any explanation.

mustafaiman: I see now. Ok.

mustafaiman:
:+1:

asimarslan:
:+1:

PR #41, opened 2016-03-09
adds membership-listener

mustafaiman:
Adds membership listener.
Membership listener is initiated when owner connection is established.
PartitionService.refresh is fired when member list changes.

### DIFF
--- a/src/invocation/InvocationService.ts
+++ b/src/invocation/InvocationService.ts
@@ -6,6 +6,7 @@ import HazelcastClient = require('../HazelcastClient');
 import {Data} from '../serialization/Data';
 import Address = require('../Address');
 import ExceptionCodec = require('../codec/ExceptionCodec');
+import {BitsUtil} from '../BitsUtil';
 
 class InvocationService {
 
@@ -13,6 +14,7 @@ class InvocationService {
 
     private correlationCounter = 0;
     private pending: {[id: number]: Q.Deferred<ClientMessage>} = {};
+    private eventHandlers: {[id: number]: (...args: any[]) => any} = {};
     private client: HazelcastClient;
     private smartRoutingEnabled: boolean;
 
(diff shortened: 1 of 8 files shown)


src/invocation/InvocationService.ts
```diff
@@ -13,6 +14,7 @@ class InvocationService {
 
     private correlationCounter = 0;
     private pending: {[id: number]: Q.Deferred<ClientMessage>} = {};
+    private eventHandlers: {[id: number]: (...args: any[]) => any} = {};
```
GhostInAMachine: I thought we agreed that events should be handled in a separate service

mustafaiman: I thought there is no point in moving a single event handler list to another service. This way nodejs resembles other clients' architecture.

GhostInAMachine: I still don't like it, but let's leave it as is for the time being

src/invocation/InvocationService.ts
```diff
@@ -21,12 +23,16 @@ class InvocationService {
         this.smartRoutingEnabled = hazelcastClient.getConfig().networkConfig.smartRouting;
     }
 
-    invokeOnConnection(connection: ClientConnection, request: ClientMessage): Q.Promise<ClientMessage> {
+    invokeOnConnection(connection: ClientConnection, request: ClientMessage, handler: (...args: any[]) => any = null)
```
GhostInAMachine: Passing handler function argument here looks really ugly to me. 

@gAmUssA What do you think?

mustafaiman: actually I will refactor this part and rebase this PR if we merge https://github.com/hazelcast-incubator/hazelcast-nodejs-client/pull/42 . When we merge that PR we can create ListenerInvocation and attach the handler to Invocation like in https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/invocation.py#L60

gAmUssA: I'm ok with a typed handler.

GhostInAMachine:
👍

PR #43, opened 2016-03-10
heartbeat service

mustafaiman:
Implements Heartbeat mechanism.
Adds options for `hazelcast.client.heartbeat.interval` and `hazelcast.client.heartbeat.timeout` to config.
Detects unresponsive nodes based on Heartbeat but doesn't do anything about it.

### DIFF
--- a/src/Heartbeat.ts
+++ b/src/Heartbeat.ts
@@ -1,0 +1,68 @@
+import {ClientPingCodec} from './codec/ClientPingCodec';
+import HazelcastClient = require('./HazelcastClient');
+import ClientConnection = require('./invocation/ClientConnection');
+
+var PROPERTY_HEARTBEAT_INTERVAL: string = 'hazelcast.client.heartbeat.interval';
+var PROPERTY_HEARTBEAT_TIMEOUT: string = 'hazelcast.client.heartbeat.timeout';
+
+class Heartbeat {
+
+    private client: HazelcastClient;
+    private heartbeatTimeout: number;
+    private heartbeatInterval: number;
+
+    //Actually it is a NodeJS.Timer. Another typing file that comes with a module we use causes TSD to see
+    //return type of setTimeout as number. Because of this we defined timer property as `any` type.
+    private timer: any;
+
+    constructor(client: HazelcastClient) {
+        this.client = client;
+        this.heartbeatInterval = this.client.getConfig().properties[PROPERTY_HEARTBEAT_INTERVAL];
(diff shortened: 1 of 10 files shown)


mustafaiman:
Connection retry does not work because of https://github.com/hazelcast-incubator/hazelcast-nodejs-client/issues/44
Heartbeat works.

src/Heartbeat.ts
```diff
@@ -0,0 +1,85 @@
...
+        }
+        this.timer = setTimeout(this.heartbeatFunction.bind(this), this.heartbeatInterval);
+    }
+
+    private onHeartbeatStopped(connection: ClientConnection) {
+        connection.heartbeating = false;
+        console.log('heartbeat stopped on ' + connection.address);
+        this.listeners.forEach((listener) => {
+            if (listener.hasOwnProperty('onHeartbeatStopped')) {
+                Q.fcall(listener.onHeartbeatStopped.bind(this), connection);
```
GhostInAMachine: Q.fcall creates a promise, but it is not used anywhere

src/invocation/ClientConnectionManager.ts
```diff
@@ -51,13 +51,16 @@ class ClientConnectionManager {
             var callback = (authenticated: boolean) => {
                 if (authenticated) {
                     result.resolve(connection);
-                    this.establishedConnections[addressIndex] = connection;
+                    this.establishedConnections[connection.address.toString()] = connection;
```
GhostInAMachine: Is there a reason for this change?

src/Heartbeat.ts
mustafaiman: Maybe I should change it with setImmediate() ?

src/invocation/ClientConnectionManager.ts
mustafaiman: Yes, if user gives `localhost` as server address we connect to localhost but then we refer the same connection with its IP address returned by server after authentication. It causes complexity in listener events. One example is heartbeat stop. Owner connection stops heartbeating but later connection manager cannot find right listener to notify because the connection is referred with name `localhost` instead of its IP.

GhostInAMachine: OK

src/Heartbeat.ts
GhostInAMachine: Yeah, you can do that

GhostInAMachine:
Okay, looks good 👍

src/Heartbeat.ts
GhostInAMachine: Probably

PR #50, opened 2016-03-21
Invocation retry

GhostInAMachine:
You can test retrying by setting up a map interceptor which would throw an exception like NullPointerException and then adding a relevant exception code to isRetryable method in RemoteException class

### DIFF
--- a/src/invocation/InvocationService.ts
+++ b/src/invocation/InvocationService.ts
@@ -134,7 +134,7 @@ export class InvocationService {
         if (messageType === EXCEPTION_MESSAGE_TYPE) {
             var remoteException = ExceptionCodec.decodeResponse(clientMessage);
             var boundToConnection = pendingInvocation.connection;
-            var deadlineExceeded = pendingInvocation.deadline.getTime() > new Date().getTime();
+            var deadlineExceeded = new Date().getTime() > pendingInvocation.deadline.getTime();
             var shouldRetry = !boundToConnection && !deadlineExceeded && remoteException.isRetryable();
 
             if (shouldRetry) {
--- a/src/invocation/InvocationService.ts
+++ b/src/invocation/InvocationService.ts
@@ -8,18 +8,26 @@ import Address = require('../Address');
 import ExceptionCodec = require('../codec/ExceptionCodec');
 import {BitsUtil} from '../BitsUtil';
 
-export interface Invocation {
+var EXCEPTION_MESSAGE_TYPE = 109;
+var INVOCATION_TIMEOUT = 120000;
+var INVOCATION_RETRY_DELAY = 1000;
+
+class Invocation {
+
+    constructor(request: ClientMessage) {
+        this.request = request;
+    }
+
     request: ClientMessage;
-    partitionId?: number;
-    address?: Address;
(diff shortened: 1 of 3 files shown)


src/invocation/InvocationService.ts
```diff
@@ -106,19 +124,35 @@ export class InvocationService {
         var messageType = clientMessage.getMessageType();
 
         if (clientMessage.hasFlags(BitsUtil.LISTENER_FLAG)) {
-            Q.fcall(this.eventHandlers[correlationId].handler, clientMessage);
```
mustafaiman: I put there a fcall because it puts the handler at the end of the event loop. If we call the handler directly, it is called immediately. We should break execution logic into parts as best as we can. You may use `setImmediate` instead of `Q.fcall` if unused promise bothers you.

src/invocation/InvocationService.ts
```diff
@@ -8,19 +8,27 @@ import Address = require('../Address');
 import ExceptionCodec = require('../codec/ExceptionCodec');
 import {BitsUtil} from '../BitsUtil';
 
-export interface Invocation {
+var EXCEPTION_MESSAGE_TYPE = 109;
+var INVOCATION_TIMEOUT = 120000;
+var INVOCATION_RETRY_DELAY = 1000;
+
+class Invocation {
```
mustafaiman: If we do not `export Invocation` no other class can create invocation by itself and call `InvocationService.invoke` directly. Do we want that? I think `Invocation` should be exported.

src/invocation/InvocationService.ts
```diff
@@ -8,19 +8,27 @@ import Address = require('../Address');
...
+
     request: ClientMessage;
-    partitionId?: number;
-    address?: Address;
-    connection?: ClientConnection;
-    deferred?: Q.Deferred<ClientMessage>;
-    handler?: (...args: any[]) => any;
+    partitionId: number;
+    address: Address;
+    deadline: Date = new Date(new Date().getTime() + INVOCATION_TIMEOUT);
```
mustafaiman: I see we retry if an exception comes from server but how do we handle the situation where no response comes at all? I did not see any code handling such situation.

GhostInAMachine: Timing out will be added in a different PR

src/invocation/InvocationService.ts
GhostInAMachine: Why should any other component have a direct access to invoke?

mustafaiman: Maybe it is not needed. Ok let's leave it unexported.

src/invocation/InvocationService.ts
mustafaiman: Ok. I created an issue so we do not forget it. https://github.com/hazelcast/hazelcast-nodejs-client/issues/52

src/invocation/InvocationService.ts
GhostInAMachine: @mustafaiman Latest commit wraps the callback with setImmediate

mustafaiman:
:+1: Rebase please

Issue #44, opened 2016-03-15
Invocation does not reject on closed connection


mustafaiman:
https://github.com/hazelcast-incubator/hazelcast-nodejs-client/blob/master/src/invocation/ClientConnection.ts#L41does not provide a method to know send operation succeeds of fails.

mustafaiman:
Solved via https://github.com/hazelcast/hazelcast-nodejs-client/pull/51

PR #51, opened 2016-03-24
Connection write error handling

mustafaiman:
In scenario of two members M1, M2 and client C;

- C authenticates to M1,
- C loses connection to M1,
- C authenticates to M2,
- M1 comes back to cluster,
- C loses connection to M2,
- C is not able to connect to M1

This PR solves this problem. Client is able to connect to any cluster member alive.

### DIFF
--- a/src/invocation/ClusterService.ts
+++ b/src/invocation/ClusterService.ts
@@ -23,17 +23,16 @@ class ClusterService {
         this.members = [];
     }
 
-    start(): Q.Promise<ClusterService> {
+    start(): Q.Promise<void> {
         this.initHeartbeatListener();
         this.initConnectionListener();
         return this.connectToCluster();
     }
 
-    connectToCluster() {
-        var deferred = Q.defer<ClusterService>();
+    connectToCluster(): Q.Promise<void> {
+        var deferred = Q.defer<void>();
         var attemptLimit = this.client.getConfig().networkConfig.connectionAttemptLimit;
         var attemptPeriod = this.client.getConfig().networkConfig.connectionAttemptPeriod;
-        var attempt = 1;
 
         this.tryAddressIndex(0, attemptLimit, attemptPeriod, deferred);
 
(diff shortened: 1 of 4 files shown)


mustafaiman:
Solves https://github.com/hazelcast/hazelcast-nodejs-client/issues/44

src/invocation/ClusterService.ts
```diff
@@ -72,26 +72,27 @@ class ClusterService {
...
-                if (index === this.addresses.length) {
-                    attemptLimit = attemptLimit - 1;
-                    if (attemptLimit === 0) {
-                        var error = new Error('Unable to connect to any of the following addresses ' + this.addresses);
-                        deferred.reject(error);
-                        return;
-                    } else {
-                        setTimeout(this.tryAddressIndex(0, attemptLimit, attemptPeriod, deferred), attemptPeriod);
-                    }
+            if (this.addresses.length <= index) {
```
GhostInAMachine: Can you explain what exactly is happening here?

mustafaiman: Baiscally, it tries to connect all the addresses until it finally finds an owner connection. If it is unable to do so in a whole pass over availbale addresses, it waits `attemptPeriod` milliseconds and tries again. It does this process `attemptLimit` times and fails if it still cannot connect

GhostInAMachine: But what does this commit fix?

mustafaiman: I cannot remember if this was fixing anything :D . Maybe cosmetic changes. I find this one more readable. AttempLimit logic is in `if` part and iterating over connections is in `else` part separately. These were a little mixed before.

GhostInAMachine: Commit name says "ClusterService connnect to cluster fix"

mustafaiman: I did not notice that you commented on the commit itself, not the PR. Commit name is misleading. The fixing part is the second commit actually by adding feedback about success of connection.write. Let me squash the commits and fix the messages.

mustafaiman: I squashed two commits and wrote correct commit message.

GhostInAMachine:
👍

src/invocation/InvocationService.ts
```diff
@@ -96,7 +96,9 @@ export class InvocationService {
             this.eventHandlers[correlationId] = invocation;
         }
         this.pending[correlationId] = invocation;
-        connection.write(invocation.request.getBuffer());
+        connection.write(invocation.request.getBuffer()).catch((e) => {
+            invocation.deferred.reject(e);
```
bilalyasar: :+1:

bilalyasar:
LGTM

PR #57, opened 2016-03-28
added nvmrc with usage of stable Node.js version

gAmUssA:


mustafaiman:
Let's say a person uses `nvm`. She clones the repo and she does not have latest stable version installed on her computer. Does nvm automatically downloads the latest stable when she tries to run this project? What if she does not want to install the latest version?

gAmUssA:
@mustafaiman A developer needs to run `nvm use` to tell `nvm` to download stable node version. If she for some reasons don't want to use stable she can tell to use a particular version with `nvm use v5.4.1`. 
Or even use already installed (not with `nvm`) node binaries with `nvm use system`. 
I will provide setup instructions in `README.adoc` later

mustafaiman:
:+1:

Issue #59, opened 2016-03-29
npm install shouldn't run the tests


gAmUssA:
subj

PR #60, opened 2016-03-29
removes ts build from npm install stage

mustafaiman:
npm install should not do anything related with Typescript compilation and tests. End user will have only javascript sources.
fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/59


gAmUssA:
:+1:

PR #64, opened 2016-04-01
Owner connection lost/reconnection fix + Remote controller test

mustafaiman:
When client lost connection it only tries to connect one of the addresses in config. With this PR it tries to find a new owner among all cluster members whether or not they were listed in config.

PR also adds some remote controller tests

Adds client.shutdown method


GhostInAMachine:
👍

PR #66, opened 2016-04-03
connection manager extends EventEmitter

mustafaiman:
This is proper Node.js way of adding listeners to an event.


GhostInAMachine:
👍

Issue #56, opened 2016-03-25
Missing methods


mustafaiman:
- getMembers() in ClusterService
- getDistributedObjects() in HazelcastInstance
- getLocalEndpoint()

mustafaiman:
fixed via https://github.com/hazelcast/hazelcast-nodejs-client/pull/80

PR #67, opened 2016-04-04
HazelcastClient.getDistributedObjects() function

mustafaiman:
Depends on https://github.com/hazelcast/hazelcast-nodejs-client/pull/64 (for testing)
Part of https://github.com/hazelcast/hazelcast-nodejs-client/issues/56

### DIFF
--- a/src/HazelcastClient.ts
+++ b/src/HazelcastClient.ts
@@ -11,6 +11,9 @@ import ClusterService = require('./invocation/ClusterService');
 import Heartbeat = require('./Heartbeat');
 import {LoggingService} from './LoggingService';
 import {LifecycleService, LifecycleEvent} from './LifecycleService';
+import {ClientGetDistributedObjectsCodec} from './codec/ClientGetDistributedObjectsCodec';
+import {DistributedObject} from './DistributedObject';
+import defer = Q.defer;
 
 class HazelcastClient {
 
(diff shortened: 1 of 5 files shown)


src/HazelcastClient.ts
```diff
@@ -68,6 +71,22 @@ class HazelcastClient {
...
+    getDistributedObjects(): Q.Promise<DistributedObject[]> {
+        var deferred = Q.defer<DistributedObject[]>();
+        var clientMessage = ClientGetDistributedObjectsCodec.encodeRequest();
+        var toObjectFunc = this.serializationService.toObject.bind(this);
+        var proxyManager = this.proxyManager;
+        this.invocationService.invokeOnRandomTarget(clientMessage).then(function(resp) {
+            var objectsInfoList = ClientGetDistributedObjectsCodec.decodeResponse(resp, toObjectFunc).response;
+            var proxies: DistributedObject[] = [];
+            for (var i = 0; i < objectsInfoList.size(); i++)  {
+                proxies.push(proxyManager.getOrCreateProxy(objectsInfoList.get(i)[1], objectsInfoList.get(i)[0]));
```
GhostInAMachine: Since you have fetched the distributed objects from the server, the proxies already exist. Why recreate them?

mustafaiman: getOrCreateProxy does not create a new proxy if it already exists.
All proxies may not be present in this client. Another client may have created a proxy at the server.

GhostInAMachine: But you still make an additional network call, no?

mustafaiman: If a proxy is already present, proxyManager directly returns that. It does not make a new network call. See https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/proxy/ProxyManager.ts#L29

GhostInAMachine: What I'm trying to say is that you already have all necessary data to create proxies from the response to GetDistributedObjects request. There is no point in making a CreateProxy request per each distributed object.

mustafaiman: I see your point now. Thanks. I'll fix that

GhostInAMachine: Since you are at it anyway, can you extract a variable from an expression "objectsInfoList.get(i)" so that it does not repeat

mustafaiman: sure

mustafaiman:
updated pr @GhostInAMachine

PR #68, opened 2016-04-05
DistributedObjectListener

mustafaiman:
Introduces `ListenerService`
Adds `addDistributedObjectListener` and `removeDistributedObjectListener` methods to `ProxyManager` and `HazelcastClient`. These methods take a function as a parameter. This function is called with `objectName`, `serviceName` and `eventType` parameters in this order when an event occurs. `eventType` is either `created` or `destroyed`.

Adds tests.

Fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/55

### DIFF
--- a/src/HazelcastClient.ts
+++ b/src/HazelcastClient.ts
@@ -1,5 +1,5 @@
 import {SerializationService} from './serialization/SerializationService';
-import {InvocationService} from './invocation/InvocationService';
+import {InvocationService, ListenerService} from './invocation/InvocationService';
 import ClientConnectionManager = require('./invocation/ClientConnectionManager');
 import {ClientConfig} from './Config';
 import ProxyManager = require('./proxy/ProxyManager');
@@ -15,6 +15,7 @@ import {ClientGetDistributedObjectsCodec} from './codec/ClientGetDistributedObje
 import {DistributedObject} from './DistributedObject';
 import defer = Q.defer;
 import {ClientInfo} from './ClientInfo';
+import ClientMessage = require('./ClientMessage');
 
 class HazelcastClient {
 
@@ -22,6 +23,7 @@ class HazelcastClient {
     private loggingService: LoggingService;
     private serializationService: SerializationService;
     private invocationService: InvocationService;
+    private listenerService: ListenerService;
     private connectionManager: ClientConnectionManager;
     private partitionService: PartitionService;
     private clusterService: ClusterService;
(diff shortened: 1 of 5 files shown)


src/HazelcastClient.ts
```diff
@@ -68,6 +71,22 @@ class HazelcastClient {
         return deferred.promise;
     }
 
+    getDistributedObjects(): Q.Promise<DistributedObject[]> {
+        var deferred = Q.defer<DistributedObject[]>();
+        var clientMessage = ClientGetDistributedObjectsCodec.encodeRequest();
+        var toObjectFunc = this.serializationService.toObject.bind(this);
+        var proxyManager = this.proxyManager;
+        this.invocationService.invokeOnRandomTarget(clientMessage).then(function(resp) {
```
GhostInAMachine: I think this duplicates the commit from another PR

mustafaiman: yes. no need to make same comments here. I'll fix both PRs

src/invocation/InvocationService.ts
```diff
@@ -162,3 +170,44 @@ export class InvocationService {
     }
 }
 
+export class ListenerService {
```
GhostInAMachine: As far as I can see it is not used yet, is it?

mustafaiman: Do you mean using ListenerService? It is used in https://github.com/hazelcast/hazelcast-nodejs-client/pull/68/files#diff-5371374fe67ebe5287b4903ca93e33ceR73 . It will be used a lot more in newer releases.

src/invocation/InvocationService.ts
```diff
@@ -162,3 +170,44 @@ export class InvocationService {
...
+        this.internalEventEmitter = new EventEmitter();
+        this.internalEventEmitter.setMaxListeners(0);
+    }
+
+    registerListener(codec: any, handler: any): Q.Promise<string> {
+        var deferred = Q.defer<string>();
+        var invocation = new Invocation(codec.encodeRequest(true));
+        invocation.handler = handler;
+        var listenerIdToCorrelation = this.listenerIdToCorrelation;
+        this.client.getInvocationService().invoke(invocation).then(function(responseMessage) {
```
GhostInAMachine: If you use a "fat arrow" syntax (=>) then you don't need to bother with "this" reference reassigning

src/invocation/InvocationService.ts
GhostInAMachine: So this is strictly for invocations related to Listener registration?

mustafaiman: Yes

PR #70, opened 2016-04-06
lifecycle-service

mustafaiman:
fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/54

### DIFF
--- a/src/Config.ts
+++ b/src/Config.ts
@@ -39,8 +39,16 @@ export interface LifecycleListener {
     (event: string): void;
 }
 
-export interface ListenerConfigs {
-    lifecycle?: LifecycleListener[];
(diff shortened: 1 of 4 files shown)


src/Config.ts
```diff
@@ -35,8 +35,12 @@ export class GlobalSerializerConfig {
...
 }
 
-export class ListenerConfig {
-    //TO-DO
+export interface LifecycleListener {
+    (event: string): void;
+}
+
+export interface ListenerConfigs {
+    lifecycle?: LifecycleListener[];
```
GhostInAMachine: Would look better if there was an "addLifecycleListener" method

mustafaiman: makes sense

PR #75, opened 2016-04-07
removes unused lodash dependency

mustafaiman:
We do not use this dependency. I do not know why it is there.


gAmUssA:
👍

gAmUssA:
I we can go ahead and merge

PR #76, opened 2016-04-07
Client UUID and Owner UUID are correctly set to secondary connections

GhostInAMachine:


src/invocation/ClientConnectionManager.ts
```diff
@@ -95,14 +96,10 @@ class ClientConnectionManager extends EventEmitter {
...
     }
 
-    private authenticate(connection: ClientConnection): Q.Promise<boolean> {
-        var name = this.client.getConfig().groupConfig.name;
-        var password = this.client.getConfig().groupConfig.password;
-        var invocationService = this.client.getInvocationService();
-
-        var authenticator = new ConnectionAuthenticator(connection, invocationService, name, password);
+    private authenticate(connection: ClientConnection, ownerConnection: boolean): Q.Promise<boolean> {
+        var authenticator = new ConnectionAuthenticator(connection, this.client);
```
mustafaiman: Isn't it better if you pass `ownerConnection` to constructor and then call `authenticator.authenticate()`. I think passing the argument to .authenticate creates the impression that the same authenticator will be used with different argument later although it will not.
But not a big deal.

mustafaiman:
:+1: with minor comment.

src/invocation/ClientConnectionManager.ts
GhostInAMachine: You have to create a constructor variable, and then a class field -> too much typing without appreciable benefit :)

PR #79, opened 2016-04-10
heartbeat tests and change in Address class

mustafaiman:
Two tests for Heartbeat Service.

I removed Address.toString method because implementation should not rely on any method of Address class. Users may provide Address objects created in Javascript to configuration and they may not provide a toString function in their object. Then client would behave unpredictably. I have put a static utility function instead of instance method.

### DIFF
--- a/src/invocation/ClientConnectionManager.ts
+++ b/src/invocation/ClientConnectionManager.ts
@@ -27,7 +27,7 @@ class ClientConnectionManager extends EventEmitter {
     }
 
     getOrConnect(address: Address, ownerConnection: boolean = false): Q.Promise<ClientConnection> {
-        var addressIndex = address.host + ':' + address.port;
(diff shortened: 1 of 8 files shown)


src/invocation/ClientConnectionManager.ts
```diff
@@ -27,7 +27,7 @@ class ClientConnectionManager extends EventEmitter {
     }
 
     getOrConnect(address: Address, ownerConnection: boolean = false): Q.Promise<ClientConnection> {
-        var addressIndex = address.toString();
+        var addressIndex = address.host + ':' + address.port;
```
GhostInAMachine: For consistency, please use encodeToString here as well

mustafaiman: good catch, thanks

PR #80, opened 2016-04-11
adds getLocalEndpoint

mustafaiman:


GhostInAMachine:
What purpose does this serve?

mustafaiman:
It is useful info for users to diagnose problems, logging.

src/invocation/ClientConnection.ts
```diff
@@ -19,6 +19,10 @@ class ClientConnection {
         this.lastRead = 0;
     }
 
+    getLocalAddress() {
+        return new Address(this.socket.localAddress, this.socket.localPort);
```
GhostInAMachine: Local address points to the address of the network interface from which the outbound connection is made, right?

mustafaiman: Yes

Issue #85, opened 2016-04-14
generate API documentation from comments


gAmUssA:
http://typedoc.io
http://typedoc.io/api/  - looks like an output of a tool
http://typedoc.io/guides/installation.html#task-runners - gulp integration

PR #86, opened 2016-04-18
Docs

mustafaiman:
Adds API documentation, uses Typedoc.
When you run `npm run typedoc` it creates html documentation using stylized comments on top of functions, properties and classes. Documentation is located under `docs` folder. It is added to `.gitignore` but NOT added to `.npmignore`. So docs folder will not be in github repository but when we `npm publish` the package, it will be added to npm repo. When somebody pulls the repo through npm, they will have `docs`.

Fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/85

### DIFF
--- a/src/invocation/InvocationService.ts
+++ b/src/invocation/InvocationService.ts
@@ -14,23 +14,52 @@ var EXCEPTION_MESSAGE_TYPE = 109;
 var INVOCATION_TIMEOUT = 120000;
 var INVOCATION_RETRY_DELAY = 1000;
 
+/**
+ * A request to be sent to a hazelcast node.
+ */
 export class Invocation {
 
     constructor(request: ClientMessage) {
         this.request = request;
     }
 
+    /**
+     * Representatiton of the request in binary form.
+     */
     request: ClientMessage;
+    /**
+     * Partition id of the request. If request is not bound to a specific partition, should be set to -1.
+     */
     partitionId: number;
+    /**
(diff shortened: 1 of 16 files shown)


src/invocation/InvocationService.ts
```diff
@@ -14,23 +14,52 @@ var EXCEPTION_MESSAGE_TYPE = 109;
...
     constructor(request: ClientMessage) {
         this.request = request;
     }
 
+    /**
+     * Representatiton of the request in binary form.
+     */
     request: ClientMessage;
+    /**
+     * Partition id of the request. If request is not bounded to a specific partition, should be set to -1.
```
GhostInAMachine: Bind (present) -> bound (past participle), not bounded

src/invocation/InvocationService.ts
```diff
@@ -14,23 +14,52 @@ var EXCEPTION_MESSAGE_TYPE = 109;
...
     connection: ClientConnection;
+
+    /**
+     * Promise managing object.
+     */
     deferred: Q.Deferred<ClientMessage>;
+
+    /**
+     * If this is an event listener registration, handler should be set to the function to be called on events.
+     * If this is not an event listener, should be set to null.
```
GhostInAMachine: Replace "If this is not an event listener" with otherwise

GhostInAMachine:
👍

PR #89, opened 2016-04-20
removes gulp from build process, uses npm scripts instead

mustafaiman:
## Motivation of Removing Gulp
I do not want to use Gulp in our builds and tests for three reasons,
- We need to introduce a different Gulp plugin for each single tool. This brings a lot of developer dependencies, which is not a good thing by itself. In addition to that, each plugin is separately maintained from their main projects and plugins are often outdated. To modify the build pipeline, one should now about using `gulp`, `the other tool` and `gulp plugin of the tool`. This is too much work. Also each piece of dependencies can have bugs.
- Gulp often does not behave nicely. It often swallows errors thrown by the tools in pipeline. This is a big problem for continuous integration. Not only continuous integration but we also suffer understanding the behaviour most of the time. e.g. `gulp-mocha` does not show the output when mocha fails because of a file not found. I often have to run the steps (that gulp was supposed to be automating for me) by hand to hunt errors.
- `npm` makes using command line interface of theses tools very easy. All you need to do is put a tool in `scripts` section of `package.json` and run `npm run <script-name>`. Also we do not need to install any tools globally by using developer dependencies and package.json scripts.

I also remove `jshint` tool from dependencies with this PR. Since JS code is generated from TS, there is no point in linting JS files. We did not use it anyway.

I changed a rule in tslint.json because old rule was removed from updated version of tslint. My change introduces the same rule with new syntax.

## New Build Commands
After a clean checkout, commands to be used are:
```
npm install (install dependencies)
npm run compile (this automatically installs typings and compiles all ts sources)
npm test (run mocha tests, without coverage)
```
Note that tslint is not a part of this pipeline. After merging this PR, we can set Cloudbees to run tslint at the server on each pull request. If you want to run `tslint` locally, run:
```
npm run lint
```
## Other Useful Commands
Code coverage
```
npm run coverage
```
Generate API Docs
```
npm run generate-docs
```


hasancelik:
verify

hasancelik:
verify

hasancelik:
verify

GhostInAMachine:
👍

PR #91, opened 2016-04-21
adds simple map benchmark with put get and remove

mustafaiman:
Map put/get/remove benchmark. 
- Put %40
- Get %40
- Remove %20

Same test as https://github.com/hazelcast/hazelcast-python-client/blob/master/benchmarks/map_async_bench.py


mustafaiman:
verify

mustafaiman:
verify

Issue #18, opened 2015-12-23
Map Proxy Implementation


asimarslan:
A documented Map API, covering async methods.
Map API implementation proxy.

PR #92, opened 2016-04-25
Map Proxy API + Implementation

mustafaiman:
fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/18. Note that this API does not include `executeOnKeys`, `addInterceptor` and Predicates related methods. These will come with Serialization implementation.


mustafaiman:
verify

src/proxy/Map.ts
```diff
@@ -56,4 +124,269 @@ export class Map<K, V> extends BaseProxy implements IMap<K, V> {
     isEmpty(): Q.Promise<boolean> {
         return this.encodeInvokeOnRandomTarget<boolean>(MapIsEmptyCodec);
     }
+
+    getAll(keys: K[]): Q.Promise<any[]> {
+        var partitionService = this.client.getPartitionService();
+        var partitionsToKeys: {[id: string]: any} = {};
+        var key: K;
+        for (var i in keys) {
+            key = keys[i];
```
GhostInAMachine: Why declare outside of the loop and then reassign within it?

src/proxy/Map.ts
```diff
@@ -56,4 +124,269 @@ export class Map<K, V> extends BaseProxy implements IMap<K, V> {
     isEmpty(): Q.Promise<boolean> {
         return this.encodeInvokeOnRandomTarget<boolean>(MapIsEmptyCodec);
     }
+
+    getAll(keys: K[]): Q.Promise<any[]> {
+        var partitionService = this.client.getPartitionService();
+        var partitionsToKeys: {[id: string]: any} = {};
+        var key: K;
```
GhostInAMachine: There is similar logic of partition ID mapping in putAll, can you extract into separate method?

src/proxy/Map.ts
mustafaiman: Putting it inside for method does not make its scope restricted to for loop. So I intentionally put them outside. This way it is less likely that someone mistakenly introduces the same variable later in the same method expecting it to be null. Matter of taste.

GhostInAMachine: If you use _let variableName = value_ it will be restricted to the loop scope.

mustafaiman: Since we use Typescript compiler target as `ES5` all `let`s are transpiled to `var`. So this would be worse. You expect it to be block scoped and get function scoped. Since Node 4 do not support majority of ES6 features we have to stick with ES5 and `var`s.

GhostInAMachine: OK

mustafaiman: There is requests from community to make Typescript support selectively turning on and off certain features: https://github.com/Microsoft/TypeScript/issues/4389 . I hope they support this feature and we can use cool ES6 features that Node already supports.

src/proxy/Map.ts
mustafaiman: I looked at it but I do not see an easy way to extract these operations to a common method. If you look closely, putAll operation is aware that it has pairs and it should use first element of the pair as key. Whereas, getAll opearation does not have key value pairs and uses elements of the array as keys directly. 

Also putAll has to do `partitionsToKeys[pId].push({key: keyData, val: this.toData(pair[1])})` different than getAll: `partitionsToKeys[pId].push(keyData);`.
I
 appreciate trying to re-use code but I do not see an easy way to achieve it in here.

GhostInAMachine: OK

jackspaniel:
Any idea when this will be published to npm?

mustafaiman:
Hello @jackspaniel . This is already merged to master so it will be available with `0.3` release. 0.3 is planned to be released early June.

PR #106, opened 2016-06-01
extend set proxy from PartitionSpecificProxy

mustafaiman:
This change removes redundant encodeInvokeOnKey calls with `this.name` as parameter.


mustafaiman:
Minor change, merging myself since it is necessary to release 0.3

PR #108, opened 2016-06-03
List implementation

GhostInAMachine:
Interface fixes

### DIFF
--- a/src/IList.ts
+++ b/src/IList.ts
@@ -1,43 +1,46 @@
+import Promise = Q.Promise;
 import {DistributedObject} from './DistributedObject';
 import {ItemListener} from './core/ItemListener';
+
 export interface IList<E> extends DistributedObject {
+
     /**
-     * Appends the specified element at the end of the list.
-     * @param element
-     * @return `true` if this list has changed.
+     * Appends the specified element to the end of the list.
+     * @param element - element to be added
+     * @return `true` if this list has changed as a result of this operation, 'false' otherwise.
      */
     add(element: E): Promise<boolean>;
 
     /**
-     * Inserts the specified element at specified index. Shifts the following
-     * elements by one position.
-     * @param index
-     * @param element
+     * Inserts the specified element at the specified index.
+     * Shifts the subsequent elements to the right.
+     * @param index position at which the element should be inserted
+     * @param element element to be inserted
      */
     addAt(index: number, element: E): Promise<void>;
 
     /**
-     * Appends all specified elements at the end of this list.
-     * @param elements
-     * @return `true` if thi list changed.
+     * Appends all elements in the specified array to the end of this list,
+     * @param elements array to be appended
+     * @return `true` if this list has changed as a result of this operation, 'false' otherwise.
      */
     addAll(elements: E[]): Promise<boolean>;
 
     /**
-     * Inserts all elements at specified index. Shifts the following
-     * elements by one position.
-     * @param index
-     * @param elements
-     * @return `true` if thi list changed.
+     * Inserts all elements in the specified array at specified index, keeping the order of the array.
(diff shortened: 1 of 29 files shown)


src/IList.ts
```diff
@@ -69,78 +71,78 @@ export interface IList<E> extends DistributedObject {
...
      * @param element
      */
     lastIndexOf(element: E): Promise<number>;
 
     /**
      * Removes the given element from this list.
-     * @param element
-     * @return `true` if this list changed, `false` otherwise.
+     * @param element element to be removed
+     * @return `true` if this list has changed as a result of this operation, 'false' otherwise.
```
mustafaiman: code pieces should be surrounded with backtics ( \` ), not single quotes ( \' ).
There are multiple occurences of it in this PR.
It is not a big problem, can be corrected with direct commit later.

src/proxy/List.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    removeItemListener(registrationId: string): Q.Promise<boolean> {
+        return this.client.getListenerService().deregisterListener(
+            ListRemoveListenerCodec.encodeRequest(this.name, registrationId),
+            ListRemoveListenerCodec.decodeResponse
+        );
+    }
+
+    private serializeList(input: Array<any>): Array<Data> {
```
mustafaiman: I think serializeList can take only `Array<E>`. There is no need to allow `Array<any>`.

test/ListProxyTest.js
```diff
@@ -0,0 +1,266 @@
...
+    it("gets item at index", function () {
+        var input = [1, 2, 3];
+        return listInstance.addAll(input).then(function () {
+            return listInstance.get(1).then(function (result) {
+                expect(result).to.equal(2);
+            });
+        });
+    });
+
+    it("removes item at index", function () {
```
mustafaiman: This same test can be and should be written as
```
    it("removes item at index", function () {
        var input = [1, 2, 3];
        return listInstance.addAll(input).then(function () {
            return listInstance.removeAt(1);
        }).then(function (removed) {
            expect(removed).to.equal(2);
            return listInstance.toArray();
        }).then(function(arr) {
            expect(arr).to.deep.equal([1, 3]);
        });
    });
```
It is much easier to read. Also this is why we use promises for. Current test misses the point in using promises.

test/ListProxyTest.js
```diff
@@ -0,0 +1,266 @@
...
+                expect(replaced).to.equal(2);
+            });
+        }).then(function () {
+            return listInstance.toArray().then(function (all) {
+                expect(all).to.deep.equal([1, 3]);
+            });
+        });
+    });
+
+    it("replaces item at index", function () {
```
mustafaiman: This test is also hard to read.

test/ListProxyTest.js
```diff
@@ -0,0 +1,266 @@
...
+            });
+        }).then(function () {
+            return listInstance.containsAll([3, 4]).then(function (contains) {
+                expect(contains).to.be.false;
+            });
+        });
+    });
+
+
+    it("is empty", function () {
```
mustafaiman: I think this test should be divided into two different tests as it has two different scenarios.

test/ListProxyTest.js
```diff
@@ -0,0 +1,266 @@
...
+                } else {
+                    done(new Error("Expected 1, got " + item))
+                }
+            }
+        };
+        return listInstance.addItemListener(listener, true).then(function () {
+            listInstance.add(1);
+        })
+    });
+
```
mustafaiman: Can you add another test where `includeValue` for ItemListener is `false`. Then branch coverage for this PR will be complete.

mustafaiman:
👍 (once green)

PR #110, opened 2016-06-06
fix: runs shutdown procedure when connection to cluster is lost.

mustafaiman:


mustafaiman:
Trivial fix. Merging myself.

PR #113, opened 2016-06-13
Lock API

GhostInAMachine:


src/proxy/ILock.ts
```diff
@@ -0,0 +1,73 @@
...
+    lock(): Promise<void>;
+
+    /**
+     * Leases the lock for a specified period of time, waiting indefinitely if necessary.
+     * The returned promise will be resolved as soon as the lock becomes available.
+     * The lock will be held for the specified amount of time and then released automatically.
+     *
+     * @param leaseMillis period of time in milliseconds for which the lock should be held.
+     */
+    lease(leaseMillis: number): Promise<void>;
```
mustafaiman: Why do we have `lease` as a seperate method? We can put `leaseMillis` into `lock()` method as an optional argument. This is also how other clients do it.

src/proxy/ILock.ts
```diff
@@ -0,0 +1,73 @@
...
+     * @returns `true` if lock was obtained right away, `false` otherwise.
+     */
+    tryLock(): Promise<boolean>;
+
+    /**
+     * Tries to acquire the lock, waiting for the specified amount of time.
+     * @param waitMillis period of time in milliseconds to wait for the lock to become available.
+     * @returns `true` if the lock was obtained within the specified period, `false` otherwise.
+     */
+    waitLock(waitMillis: number): Promise<boolean>;
```
mustafaiman: I think this is "with lease" version of `tryLock`. This can be an optional argument in `tryLock`.

GhostInAMachine: TypeScript does not support method overloading

src/proxy/ILock.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Tries to lease the lock for `leaseMillis`, waiting for the lock to become available for `waitMillis`.
+     * @param waitMillis period of time in milliseconds to wait for the lock to become available.
+     * @param leaseMillis period of time in milliseconds for which the lock should be held.
+     * @returns `true` if the lock was obtained within the specified period, `false` otherwise.
+     */
+    waitLease(waitMillis: number, leaseMillis: number): Promise<boolean>;
```
mustafaiman: Having a lot of methods with different names is confusing. We should make our API similar to Hazelcast's other clients as much as possible. See https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/proxy/lock.py#L29 and https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/core/ILock.java#L90

src/proxy/ILock.ts
mustafaiman: But it supports optional arguments in interfaces and default values in implementation. Those two methods should be aggreagated: `tryLock(waitMillis?: number)`.

mustafaiman: and also waitLease: `tryLock(waitMillis?: number, leaseMillis?: number)`

src/proxy/ILock.ts
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Forcefully unlocks the lock.
+     * Usually, the same client has to call `unlock` the same amount of times
+     * as the amount of times it has called `lock`, otherwise the lock will remain in locked state.
+     * This method will disregard the acquire count and release the lock immediately.
+     */
+    forceUnlock(): Promise<void>;
+
+    /**
+     * Checks if this lock is in the locked state.
```
mustafaiman: Please add a `@return` with relevant information to doc.

src/proxy/ILock.ts
```diff
@@ -0,0 +1,73 @@
...
+     */
+    isLocked(): Promise<boolean>;
+
+    /**
+     * Checks if this lock was put into the locked state by this client.
+     */
+    isLockedByThisClient(): Promise<boolean>;
+
+    /**
+     * Returns the number of times this lock was acquired by its owner,
```
mustafaiman: `@returns`

src/proxy/ILock.ts
```diff
@@ -0,0 +1,73 @@
...
+     * i.e. how many times the `unlock` method should be invoked for the lock to become free.
+     */
+    getLockCount(): Promise<number>;
+
+    /**
+     * Returns the number of milliseconds in which the lease for this lock will expire.
```
mustafaiman: `@returns`

PR #114, opened 2016-06-14
Lock Implementation

GhostInAMachine:


mustafaiman:
Again failed with generated file. Please add a newline at the end of that file. I'll fix this in code-generator later.

test/LockProxyTest.js
```diff
@@ -0,0 +1,180 @@
...
+        lockOne = clientOne.getLock('test');
+        lockTwo = clientTwo.getLock('test');
+    });
+
+    afterEach(function () {
+        return Q.all([lockOne.destroy(), lockTwo.destroy()]);
+    });
+
+    after(function () {
+        clientOne.shutdown();
```
mustafaiman: Is `clientTwo.shutdown` missing here?

test/LockProxyTest.js
```diff
@@ -0,0 +1,180 @@
...
+            return lockOne.isLocked();
+        }).then(function (locked) {
+            expect(locked).to.be.true;
+            return lockTwo.isLocked();
+        }).then(function (locked) {
+            expect(locked).to.be.true;
+        });
+    });
+
+    it("correctly reports remaining lease time", function () {
```
mustafaiman: Is this test reliable? What if more than 100 milliseconds pass before `getRemainingLeaseTime` returns? This may cause irrelevant failures in PR builder in future.

Maybe we can call `getRemainingLeaseTime` two times one after the other. Then assert that first response is greater than the second.

mustafaiman:
👍 with minor comments. Nice tests 💯

PR #115, opened 2016-06-16
adds queue and set code samples

bounaykut:
poll, peek and add functions for QUEUE

contains, add, remove and size functions for SET

add(Object),  add(Object, index), remove(Object) and remove(index) functions for LIST

### DIFF
--- a/code_samples/list.js
+++ b/code_samples/list.js
@@ -7,8 +7,8 @@ var addPerson = function(list, val){
 };
 
 var addPersonTo = function(list, val, index){
-    return   list.add(val, index).then(function(previousVal){
-        console.log('Added value to the ' + JSON.stringify(index) + 'th index: ' + val);
+    return   list.add(val, index).then(function(){
+        console.log('Added value to the ' + index + 'th index: ' + val);
     });
 };
 
--- a/code_samples/list.js
+++ b/code_samples/list.js
@@ -1,0 +1,59 @@
+var Client = require('../.').Client;
+
+var addPerson = function(list, val){
+  return   list.add(val).then(function(){
+        console.log('Added value: ' + val);
+  });
+};
+
+var addPersonTo = function(list, val, index){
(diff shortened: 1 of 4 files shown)


code_samples/queue.js
```diff
@@ -0,0 +1,61 @@
+var Client = require('../.').Client;
+
+var insertPerson = function (queue, val) {
+    return queue.add(val).then(function(previousVal) {
+        console.log('Added value: ' + JSON.stringify(val) + ',  previous value: ' + JSON.stringify(previousVal));
```
mustafaiman: Can you print raw values instead of `JSON.stringify`. Like `console.log('Added value: ' + val + ', prev....)`. Since they are all strings there is no need for JSON.

code_samples/map.js
```diff
@@ -13,7 +13,7 @@ var removePerson = function (map, key) {
 
 var getPerson = function (map, key) {
     return map.get(key).then(function(val) {
-        console.log('Person with id ' + key + ': ' + JSON.stringify(val));
+        console.log('Person with id ' + JSON.stringify(key) + ': ' + JSON.stringify(val));
```
mustafaiman: Why did you add `JSON.stringify` here? `key` is already a number. There is no need to `stringify` it before printing to console. We `stringify` `val` because it is an `object` and not printable nicely.

code_samples/list.js
```diff
@@ -0,0 +1,59 @@
...
+
+var addPerson = function(list, val){
+  return   list.add(val).then(function(){
+        console.log('Added value: ' + val);
+  });
+};
+
+var addPersonTo = function(list, val, index){
+    return   list.add(val, index).then(function(previousVal){
+        console.log('Added value to the ' + JSON.stringify(index) + 'th index: ' + val);
```
mustafaiman: `index` is a number. There is no need to `stringify` it.

mustafaiman:
👍 with minor comments. Nice samples!

PR #120, opened 2016-06-21
Removes unneccessary promise creation

mustafaiman:
Removes unneccessary promise in BaseProxy.createPromise. Brings noticable performance increase.

### DIFF
--- a/src/proxy/BaseProxy.ts
+++ b/src/proxy/BaseProxy.ts
@@ -20,19 +20,12 @@ export class BaseProxy {
     }
 
     private createPromise<T>(codec: any, promise: Q.Promise<ClientMessage>): Q.Promise<T> {
-        var deferred: Q.Deferred<T> = Q.defer<T>();
         var toObject = this.toObject.bind(this);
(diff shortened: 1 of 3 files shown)


src/proxy/BaseProxy.ts
```diff
@@ -20,19 +20,15 @@ export class BaseProxy {
...
-        var deferred: Q.Deferred<T> = Q.defer<T>();
         var toObject = this.toObject.bind(this);
-        promise.then(function(clientMessage: ClientMessage) {
-            if (codec.hasOwnProperty('decodeResponse')) {
+        return promise.then(function(clientMessage: ClientMessage) {
+            if (codec.decodeResponse) {
                 var parameters: any = codec.decodeResponse(clientMessage, toObject);
-                deferred.resolve(parameters.response);
+                return parameters.response;
             } else {
```
GhostInAMachine: I think `else` branch can be removed here

mustafaiman: Right

PR #122, opened 2016-06-22
adds ClassDefinitionTest

bounaykut:
Test units for getFieldById, hasField and getFieldType and equals methods was added.


mustafaiman:
verify

Issue #124, opened 2016-06-23
[Test-Failure] Lock Proxy acquires lock before timeout is exceeded


mustafaiman:
https://hazelcast-l337.ci.cloudbees.com/view/Clients/job/NodeJS-pr-builder/102/console

> 17:24:27   1) Lock Proxy acquires lock before timeout is exceeded:
17:24:27      AssertionError: expected 2 to be above 1000
17:24:27       at test/LockProxyTest.js:90:35
17:24:27       at _fulfilled (node_modules/q/q.js:834:54)
17:24:27       at self.promiseDispatch.done (node_modules/q/q.js:863:30)
17:24:27       at Promise.promise.promiseDispatch (node_modules/q/q.js:796:13)
17:24:27       at node_modules/q/q.js:604:44
17:24:27       at runSingle (node_modules/q/q.js:137:13)
17:24:27       at flush (node_modules/q/q.js:125:13)

mustafaiman:
This test fails a lot in PR builder.

PR #125, opened 2016-06-23
Move to bluebird

mustafaiman:
I changed from Q to Bluebird. It should not affect end users as we still return Promises.
I advise doing a clean install at your local after this PR is merged:
```
npm run clean
rm -r node_modules
npm install
npm run compile
```

### DIFF
--- a/src/HazelcastClient.ts
+++ b/src/HazelcastClient.ts
@@ -1,7 +1,7 @@
 import {SerializationService, SerializationServiceV1} from './serialization/SerializationService';
 import {InvocationService, ListenerService} from './invocation/InvocationService';
 import {ClientConfig} from './Config';
-import * as Q from 'q';
+import * as Promise from 'bluebird';
 import {IMap} from './proxy/IMap';
 import {ISet} from './proxy/ISet';
 import {LoggingService} from './logging/LoggingService';
@@ -38,7 +38,7 @@ export default class HazelcastClient {
      * @param config Default {@link ClientConfig} is used when this parameter is absent.
      * @returns {Q.Promise<HazelcastClient>}
      */
-    public static newHazelcastClient(config?: ClientConfig): Q.Promise<HazelcastClient> {
+    public static newHazelcastClient(config?: ClientConfig): Promise<HazelcastClient> {
         var client: HazelcastClient = new HazelcastClient(config);
         return client.init();
     }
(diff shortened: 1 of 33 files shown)


mustafaiman:
Unrelated test failure: https://github.com/hazelcast/hazelcast-nodejs-client/issues/124

mustafaiman:
verify

mustafaiman:
@GhostInAMachine review please

src/HazelcastClient.ts
```diff
@@ -61,8 +61,8 @@ export default class HazelcastClient {
         this.heartbeat = new Heartbeat(this);
     }
 
-    private init(): Q.Promise<HazelcastClient> {
-        var deferred = Q.defer<HazelcastClient>();
+    private init(): Promise<HazelcastClient> {
+        var deferred = Promise.defer<HazelcastClient>();
```
GhostInAMachine: This is deprecated. 
http://bluebirdjs.com/docs/api/deferred-migration.html

mustafaiman:
I currently removed some deferreds but not all of them. I'll change all of them but I would like to tell my concerns about removing deferreds;
- Deferreds suits our invocation system best. We store deferreds in `pendingQueue` and resolve when response to that request is received. To do that without using deferred objects, we need to store specific `resolve` and `reject` functions to each invocation. This seems a little odd. In fact, Bluebird's migration quide code simulates the same phenomenon using Bluebird's API but not native support. So they even could not find a proper better way to replace what they call `discouraged`.
- Promise constructor pattern is noticeably slower compared to deferreds. See 4th paragraph of Esailija's answer here: http://programmers.stackexchange.com/questions/278778/why-are-native-es6-promises-slower-and-more-memory-intensive-than-bluebird. Esailjia is the author of Bluebird and he himself says Promise constructor pattern is slow by definition. Then Bluebird deprecates deferred pattern. This does not make sense. Keep in mind that Bluebird is claiming to be faster than native promises(they really are) and they explain why they are faster by commenting on inherently slow constructor of native API.
- Another Bluebird author says deferreds are likely to be never removed here: https://github.com/petkaantonov/bluebird/issues/555#issuecomment-88588920 . Also the reason he gives to justify the deprecation does not apply to our case in Invocations.
- Bluebird documentation says **Use deferreds when you have to**. See `When should deferred be used` section in https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns. So they admit deferreds need to be present in the library. So they are likely to be never removed.

On the other hand, I think we should not count on library and use only Promises/A+ specification. I will change deferreds to Promise constructor pattern but note that this will have some performance drawback.

PR #126, opened 2016-06-23
adds logging code sample

bounaykut:
Responses to different types of logging properties

### DIFF
--- a/code_samples/logging.js
+++ b/code_samples/logging.js
@@ -1,0 +1,67 @@
+var winston = require('winston');
+var Config = require('../.').Config;
+var HazelcastClient = require('../.').Client;
+
+if(process.argv.length != 3){
+    console.log('You must give exactly one argument to run this tool!!!');
+}else{
+
+
+    var cfg = new Config.ClientConfig();
+    cfg.properties['hazelcast.logging'] = process.argv[2];
+    
+    switch (cfg.properties['hazelcast.logging']){
+       
+        case 'winston':
+            var winstonAdapter = {
+                logger: new (winston.Logger)({
+                    transports: [
+                        new (winston.transports.Console)()
+                    ]
+                }),
+
+                levels: [
+                    'error',
+                    'warn',
+                    'info',
+                    'debug',
+                    'silly'
+                ],
+
+                log: function(level, className, message, furtherInfo) {
+                    this.logger.log(this.levels[level], className + ' ' + message);
+                }
+            };
+            
+            
+            //winston logging is active in case of winston object given as property
+            cfg.properties['hazelcast.logging'] = winstonAdapter;
+            break;
+            
+       
+        case 'default':
+            cfg.properties['hazelcast.logging'] = 'default';
+            break;
+       
+        case 'empty':
+            cfg.properties['hazelcast.logging'] = 'default';
+            break;
+       
+        case 'off':
+            cfg.properties['hazelcast.logging'] = 'off';
+            console.log('Client starting with off-logger');
+            break;
+        
+        default:
+            cfg.properties['hazelcast.logging'] = 'unknown';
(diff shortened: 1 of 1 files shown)


code_samples/logging.js
```diff
@@ -0,0 +1,81 @@
...
+    log: function(level, className, message, furtherInfo) {
+        this.logger.log(this.levels[level], className + ' ' + message);
+    }
+};
+
+var loggingHappened = false;
+
+
+winstonAdapter.logger.on('logging', function(transport, level, msg, meta) {     
+    loggingHappened = true;
```
mustafaiman: You can `console.log` some notification here. Assigning to a variable does not have much visibility from perspective of code sample user.

code_samples/logging.js
```diff
@@ -0,0 +1,81 @@
...
+    .then(function(){
+        //default logging is active in case of empty property
+        return HazelcastClient.newHazelcastClient(cfg);
+    })
+    .then(function(client){
+        return client.shutdown();
+    });
+
+//Error is thrown in case of unknown property
+cfg.properties['hazelcast.logging'] = 'asdas';
```
mustafaiman: Please use meaningful sentences/words. `nonexistent-logging` maybe?

code_samples/logging.js
```diff
@@ -0,0 +1,81 @@
...
+HazelcastClient.newHazelcastClient(cfg)
```
mustafaiman: It looks like you open two client instances simultaneously. This one and the one with "winstonAdapter". Their logs may mix.

mustafaiman:
It is good that you show different behaviour with different configurations. However, putting them all in on sample one after another is confusing. Maybe you can split them to different files so one can easily see what is the effect of different configurations. Even better idea is to write this sample with command line options. For example, a user may run the sample with `node logging.js winston` and see winston logging. `node logging.js none` and see no-logger version, etc.

test/LoggingTest.js
```diff
@@ -28,6 +28,14 @@ describe('Logging Test', function() {
         }
     };
 
+
```
mustafaiman: Why did you change `LoggingTest`?

code_samples/logging.js
```diff
@@ -0,0 +1,79 @@
...
+       
+        case 'offLogging.js':
+            HazelcastClient.newHazelcastClient().then(function(client){
+                console.log('There is no logging when configuration is off');
+                client.shutdown();
+            })
+            break;
+        
+        default:
+            console.log('Options for logging are '+'1-)winstonLogging.js, 2-)defaultLogging.js, 3-)emptyLogging.js, 4-)offLogging.js' +
```
mustafaiman: I think arguments does not need to end with `Logging.js`. `winston`, `default`, `empty`, and `off` are okay.

code_samples/logging.js
```diff
@@ -0,0 +1,79 @@
...
+                log: function(level, className, message, furtherInfo) {
+                    this.logger.log(this.levels[level], className + ' ' + message);
+                }
+            };
+
+            var loggingHappened = false;
+
+
+            winstonAdapter.logger.on('logging', function(transport, level, msg, meta) {
+                loggingHappened = true;
```
mustafaiman: Does this any use in this sample?

code_samples/logging.js
```diff
@@ -0,0 +1,79 @@
...
+            });
+
+            //winston logging is active in case of winston object given as property
+            var cfg = new Config.ClientConfig();
+            cfg.properties['hazelcast.logging'] = winstonAdapter;
```
mustafaiman: I think this sample is unneccessarily repetitive. What you can do is to put only `cfg.properties['hazelcast.logging'] = logType` in switch-case. The rest of the code is the same for all samples.

PR #128, opened 2016-06-29
MultiMap API

GhostInAMachine:


src/proxy/MultiMap.ts
```diff
@@ -0,0 +1,172 @@
...
+     * times, otherwise the lock will remain unavailable.
+     *
+     * If lease time is specified, then the lock will automatically become available
+     * after the specified time has passed.
+     * If lease time is not specified or is less than zero,
+     * then lock owner must call `unlock` to make the lock available.
+     * @param key key to be locked.
+     * @param leaseMillis lease time in milliseconds.
+     */
+    lock(key: K, leaseMillis: number): Promise<void>;
```
mustafaiman: I guess `leaseMillis` should be optional here.

src/proxy/MultiMap.ts
```diff
@@ -0,0 +1,172 @@
...
+    get(key: K): Promise<Array<V>>;
+
+    /**
+     * Removes an association of the specified value with the specified key. Calling this method does not affect
+     * other values associated with the same key.
+     * @param key key from which the value should be detached.
+     * @param value value to be removed.
+     * @return `true` if the value was detached from the specified key, `false` if it was not.
+     */
+    remove(key: K, value: V): Promise<boolean>;
```
mustafaiman: Why don't we make `value` optional here and get rid of `removeAll`?

src/proxy/MultiMap.ts
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * If lease time is specified, then the lock will automatically become available
+     * after the specified time has passed.
+     * If lease time is not specified or is less than zero,
+     * then lock owner must call `unlock` to make the lock available.
+     * @param key key to be locked
+     * @param timeoutMillis timeout for locking, in milliseconds
+     * @param leaseMillis lease time in milliseconds
+     */
+    tryLock(key: K, timeoutMillis: number, leaseMillis: number): Promise<boolean>;
```
mustafaiman: `timoutMillis` and `leaseMillis` should be optional.

src/proxy/MultiMap.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * @return an array of all keys in this multi-map.
+     */
+    keySet(): Promise<Array<K>>;
+
+    /**
+     * @return a flat array of all values stored in this multi-map.
+     */
+    values(): Promise<Array<V>>;
```
mustafaiman: As far as I know `Array` and `[]` does not have any difference in TS. You sometimes use one and sometimes the other in this interface. If they were consistent it would look better.

GhostInAMachine: I think `Array` looks more clear

src/proxy/MultiMap.ts
GhostInAMachine: Will fix

src/proxy/MultiMap.ts
GhostInAMachine: Remove with key and value and remove with just key have different return values.

src/proxy/MultiMap.ts
GhostInAMachine: Will fix

src/proxy/MultiMap.ts
mustafaiman: I see, we could merge these two methods as `remove(key: K, value?: V): Promise<boolean> | Promise<Array<K>>` anyway.
Separating them is fine too. 👍

src/proxy/MultiMap.ts
mustafaiman: Then, maybe you can change `[]` in `entrySet` with `Array` so they become consistent.

mustafaiman:
👍 I will merge once green.

PR #130, opened 2016-06-30
MultiMap Impl

GhostInAMachine:


src/proxy/MultiMapProxy.ts
```diff
@@ -0,0 +1,204 @@
...
+        var keyData = this.toData(key);
+        var valueData = this.toData(value);
+        return this.encodeInvokeOnKey<boolean>(MultiMapPutCodec, keyData, keyData, valueData, 1);
+    }
+
+    get(key: K): Promise<Array<V>> {
+        var keyData = this.toData(key);
+
+        var deserializeCallback = (items: Array<Data>) => {
+            return items.map((item) => {
```
mustafaiman: No need to create second function here. This can be replaced with simply `return items.map(this.toObject)`

src/proxy/MultiMapProxy.ts
```diff
@@ -0,0 +1,204 @@
...
+        var keyData = this.toData(key);
+        var valueData = this.toData(value);
+        return this.encodeInvokeOnKey<boolean>(MultiMapRemoveEntryCodec, keyData, keyData, valueData, 1);
+    }
+
+    removeAll(key: K): Promise<Array<V>> {
+        var keyData = this.toData(key);
+        return this.encodeInvokeOnKey<Array<Data>>(MultiMapRemoveCodec, keyData, keyData, 1)
+            .then<Array<V>>((items: Array<Data>) => {
+                return items.map((item) => {
```
mustafaiman: This can be simplified too. As above.

src/proxy/MultiMapProxy.ts
```diff
@@ -0,0 +1,204 @@
...
+                    return this.toObject(item);
+                });
+            });
+    }
+
+    keySet(): Promise<Array<K>> {
+        return this.encodeInvokeOnRandomTarget<Array<Data>>(MultiMapKeySetCodec)
+            .then<Array<K>>((items: Array<Data>) => {
```
mustafaiman: Actually I see this piece of code is used everywhere in this method. Why not just extract this as a private method and use it everywhere rather than creating another function everytime?

src/proxy/MultiMapProxy.ts
```diff
@@ -0,0 +1,204 @@
...
+                    return this.toObject(item);
+                });
+            });
+    }
+
+    entrySet(): Promise<Array<[K, V]>> {
+        return this.encodeInvokeOnRandomTarget<Array<[Data, Data]>>(MultiMapEntrySetCodec)
+            .then<Array<[K, V]>>((entrySet: [Data, Data][]) => {
+                return entrySet.map((entry: Array<any>) => {
+                    var tuple: [K, V] = [this.toObject(entry[0]), this.toObject(entry[1])];
```
mustafaiman: return `[this.toObject(entry[0]), this.toObject(entry[1])]`;

GhostInAMachine: TypeScript cannot infer the types unless you explicitly specify it

```
Error:(88, 34) TS2345: Argument of type '(entrySet: [Data, Data][]) => any[][]' is not assignable to parameter of type '(value: [Data, Data][]) => [K, V][] | Thenable<[K, V][]>'.
 Type 'any[][]' is not assignable to type '[K, V][] | Thenable<[K, V][]>'.
 Type 'any[][]' is not assignable to type 'Thenable<[K, V][]>'.
 Property 'then' is missing in type 'any[][]'.
```

GhostInAMachine:
I squashed a couple of intermediate commits

src/proxy/MultiMapProxy.ts
mustafaiman: you can still use `return <[K, V]>[this.toObject(entry[0]), this.toObject(entry[1])];`

mustafaiman:
👍

PR #133, opened 2016-07-20
Codec Fix

GhostInAMachine:


mustafaiman:
@GhostInAMachine This is a big PR with a lot of clutter. Can you name which classes you actually changed? Also can you reformat code using your IDE? If you use IntelliJ you need to just select codec folder, right click, select `Reformat code`. I think that will reduce clutter in the PR.

GhostInAMachine:
@mustafaiman I generated codecs from master of client protocol and then replaced all exiting codecs with newly generated ones

I had to fix some distributed object proxies to match these changes:

1) Map representation as array of `{key: "x", value: "y"}` objects
2) Addition of reference ID to locking operations
3) Newly generated codecs pass toObjectFunction to custom codecs, even though only EntryViewCodec is using them. I had to add dummy arguments to others in order to avoid compilation errors

mustafaiman:
@GhostInAMachine 
> 3) Newly generated codecs pass toObjectFunction to custom codecs, even though only EntryViewCodec is using them. I had to add dummy arguments to others in order to avoid compilation errors

Why did you not add toObject function as an optional argument to custom codecs instead of passing dummy arguments? For example in AddressCodec.ts constructor is:
```
static encode(clientMessage: ClientMessage, target: Address): void {
```
Isn't it possible to just assign target null or undefined value as default:
```
static encode(clientMessage: ClientMessage, target: Address = undefined): void {
```
So you will not have to pass dummy objects.

GhostInAMachine:
It is actually the other way around. 

Generated codecs pass the toObjectFunction and custom codecs do not have it

mustafaiman:
I see. I assumed that we(not only generated codecs but our hand written code) somehow call custom codecs and we need pass dummy objects to them. I see this is not the issue, so my comment was not relevant. 👍

PR #137, opened 2016-08-03
PartitionAware

mustafaiman:
PartitionAware support.
Users should put a `getPartitionKey` method to their objects in order to set the partition key explicitly.

### DIFF
--- a/src/core/PartitionAware.ts
+++ b/src/core/PartitionAware.ts
@@ -1,0 +1,24 @@
+/**
+ * Implementing PartitionAware allows one to override the default partitioning scheme.
+ * Instead of using the keys themselves to spread the data around the cluster the
+ * key returned by {@link getPartitionKey} is used.
+ * <p/>
+ * This provides the user with an ability to contain related keys within the same
+ * partition and, consequently, within the same node.
+ * <p/>
+ * In Hazelcast, disparate data structures will be stored on the same partition,
+ * based on the partition key. For example, if "Steve" was used, then the following would be on one partition.
+ * <ul>
+ *     <li>a customers IMap with an entry of key "Steve"</li>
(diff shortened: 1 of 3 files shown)


src/core/PartitionAware.ts
```diff
@@ -0,0 +1,22 @@
+/**
```
GhostInAMachine: I know this is a straight copy of the Javadoc, but I still think this is poorly worded. 

```PartitionAware means that data will be based in the same member based on the partition key
and implementing tasks will be executed on the {@link getPartitionKey}'s owner member.
```

I would phrase it like this:

Implementing PartitionAware allows one to override the default partitioning scheme. Instead of using the keys themselves to spread the data around the cluster the key returned by `getPartitionKey()` is used. 

This provides the user with an ability to contain related keys within the same partition and, consequently, within the same node. 


`Data and execution occurs on the same partition.`

What is "data occurring"?

I would omit any references to an executor service since it is not likely we will have it in Node client.

src/serialization/SerializationService.ts
```diff
@@ -47,10 +48,17 @@ export class SerializationServiceV1 implements SerializationService {
         this.registerGlobalSerializer(serializationConfig.globalSerializer);
     }
 
-    toData(object: any, partitioningStrategy: any = this.defaultPartitionStrategy): Data {
+    toData(object: any | PartitionAware<any>, partitioningStrategy: any = this.defaultPartitionStrategy): Data {
```
GhostInAMachine: I think `any` here already has a wider "scope" than `PartitionAware`, so it can be removed.

src/core/PartitionAware.ts
mustafaiman: 👍 thanks for the doc. I am changing it as you say.

PR #140, opened 2016-08-16
Ringbuffer

GhostInAMachine:


gAmUssA:
👍  looks good to me @GhostInAMachine.

src/proxy/IRingbuffer.ts
```diff
@@ -0,0 +1,124 @@
+import * as Promise from 'bluebird';
+import {DistributedObject} from '../DistributedObject';
+import {OverflowPolicy} from '../core/OverflowPolicy';
+
+export interface IRingbuffer<E> extends DistributedObject {
```
gAmUssA: why `IRingbuffer`? in java it's `Ringbuffer`. @asimarslan WDYT?

gAmUssA:
added small question regarding naming. `IRingbuffer` vs `Ringbuffer`

src/proxy/IRingbuffer.ts
asimarslan: I think a language convention. We choose to use I notation for interfaces based on lang.

gAmUssA: @asimarslan I'm fine with that.  I consider this as a sign off from you

src/proxy/RingbufferProxy.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    headSequence(): Promise<Long> {
+        return this.encodeInvoke<Long>(RingbufferHeadSequenceCodec);
+    }
+
+    remainingCapacity(): Promise<Long> {
+        return this.encodeInvoke<Long>(RingbufferRemainingCapacityCodec);
+    }
+
+    add(item: E, overflowPolicy: OverflowPolicy = OverflowPolicy.OVERWRITE): Promise<Long> {
```
mustafaiman: `overflowPolicy` is not specified as optional in `IRingBuffer`.
Also, Java RingBuffer's `addAsync` requires user to provide `overflowPolicy` everytime. I think there should not be a default value here.
Also, a null check is required for `overflowPolicy`

test/ringbuffer/RingbufferTest.js
```diff
@@ -0,0 +1,113 @@
...
+
+    it("correctly reports size", function () {
+        return rb.addAll([1, 2]).then(function () {
+            return rb.size().then(function (size) {
+                expect(size.toNumber()).to.equal(2);
+            });
+        })
+    });
+
+
```
mustafaiman: `overflowPolicy` is a **required** field for `addAll`. `addAll` calls are actually broken in this test.

GhostInAMachine: @mustafaiman there is a default value for overflowPolicy, I suppose including that every time would be too verbose

mustafaiman: @GhostInAMachine it is a required value in the interface. If this has a default value, interface should show it as `optional`.

Apart from that, `Ringbuffer#addAllAsync` does not provide a default `overflowPolicy` in Java. So node.js should not, either.

GhostInAMachine: @mustafaiman Could you please check my latest updates to see if you are alright with such design?

mustafaiman:
```
Apart from that, Ringbuffer#addAllAsync does not provide a default overflowPolicy in Java. So node.js should not, either.
```
I think Node.js API should conform to Java's approach about data structures unless there is really a good reason not to. Why do you think that `addAll` and `add` should have a default overflow policy? You have to justify this decision.

GhostInAMachine: @mustafaiman 

1. `add` method in Java API has a default overflow policy `OVERWRITE` (though `addAsync` does not) 
2. Having default parameters for methods is a very widespread practice in the JavaScript world. Just look at jQuery API

mustafaiman: @GhostInAMachine 
1. `add` method of Node.js is actually `addAsync` of Java. So Java does not have default overflow policy for this function.
2. I do not think this is a good reason not to follow our client conventions.
@asimarslan WDYT about the discussion here?

asimarslan: From Java RingBuffer you have two methods, add and addAsync. If we map these to js fncs.
`add(item)` --> `add(item)`
`addAsync(item, overflowPolicy)` --> `add(item, overflowPolicy)`

And the only way to merge the above mapping is assigning a default-value:
`add(item, overflowPolicy=OVERWRITE)`
So the resulting is add with default value.
If we offer to update Java interface with `addAsync(item)` there won't be a function problem(other than it make the api crowded)
So I'm totally OK with 
`add(item: E, overflowPolicy: OverflowPolicy = OverflowPolicy.OVERWRITE)`

GhostInAMachine:
Can someone review this again?

gAmUssA:
verify

gAmUssA:
tests are passed.
@mustafaiman if you don't have any objections, I go ahead and merge

PR #142, opened 2016-09-15
Implemented custom authentication

GhostInAMachine:


gAmUssA:
@GhostInAMachine kindly, add tests

GhostInAMachine:
Custom auth requires enterprise at the server side. Also, there's quite a lot of setup involved, custom auth object, this and that. So it cannot be automated.

gAmUssA:
@GhostInAMachine is it backward compatible with group+pass auth?

GhostInAMachine:
Regular auth can still be used without a problem. Is that what you mean?

jackspaniel:
Thanks a lot guys. I will try this out next week with our COSC team and let you know if it works.

gAmUssA:
@jackspaniel ping us in gitter then.

PR #143, opened 2016-10-04
nmishra@commercehub.com: Fix bug in ClusterService.ts.

nitish661401869:
Fix bug in ClusterService.ts. that throws an error and breaks the unit tests and code execution in latest versions of Node (v6.3.0).

The setTimeout function used in ClusterService.ts contains a function invocation as its first argument (where as the first argument of setTimeout should be a function definition that is invoked later). This bug is masked in lower versions of Node as no error is thrown.

### DIFF
--- a/src/invocation/ClusterService.ts
+++ b/src/invocation/ClusterService.ts
@@ -143,7 +143,8 @@ class ClusterService extends EventEmitter {
                     deferred.reject(error);
                     return;
                 } else {
-                    setTimeout(this.tryAddressIndex(0, attemptLimit, attemptPeriod, deferred), attemptPeriod);
+                    var self = this;
+                    setTimeout(function() {self.tryAddressIndex(0, attemptLimit, attemptPeriod, deferred)}, attemptPeriod);
                 }
             } else {
                 var currentAddress = this.knownAddresses[index];
--- a/src/invocation/ClusterService.ts
+++ b/src/invocation/ClusterService.ts
@@ -143,8 +143,10 @@ class ClusterService extends EventEmitter {
                     deferred.reject(error);
                     return;
                 } else {
-                    var self = this;
-                    setTimeout(function() {self.tryAddressIndex(0, attemptLimit, attemptPeriod, deferred)}, attemptPeriod);
+                    setTimeout(
+                        this.tryAddressIndex.bind(this, 0, attemptLimit, attemptPeriod, deferred),
+                        attemptPeriod
+                    );
                 }
             } else {
                 var currentAddress = this.knownAddresses[index];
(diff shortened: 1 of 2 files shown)


mearns:
It's worth noting that the code (prior to the fix) is broken in _all versions of node_, it just doesn't cause an exception in earlier versions. The function is intended to recursively call itself as a timeout, but calls itself synchronously only one additional time, and passes the result (`undefined`) into `setTimeout`. In previous versions of node, `setTimeout` accepted null and undefined as valid no-op first arguments, but in node 6.3, it throws an error.

This is a significant bug because the error is throw in asynchronously, so there's absolutely no way for client code to catch it, it just causes the entire node process to terminate in error.

src/invocation/ClusterService.ts
```diff
@@ -143,7 +143,8 @@ class ClusterService extends EventEmitter {
                     deferred.reject(error);
                     return;
                 } else {
-                    setTimeout(this.tryAddressIndex(0, attemptLimit, attemptPeriod, deferred), attemptPeriod);
+                    var self = this;
+                    setTimeout(function() {self.tryAddressIndex(0, attemptLimit, attemptPeriod, deferred)}, attemptPeriod);
```
mustafaiman: Can you change the function decleration to a function binding as `this.tryAddressIndex.bind(this, 0, attemptLimit, attemptPeriod, deferred)`. It is just to be consistent with the rest of the codebase. Then it's ok. Thanks for the fix.

mustafaiman:
verify

mearns:
Can someone with permission to the cloudbees project provide some information about what build /test failures there were, if any? The message seems to indicate that it actually built successfully, but there were no tests to be run; not sure if I'm interpreting that correctly.

mustafaiman:
@mearns it failed because of missing semicolon at the end of `self.tryAddressIndex(0, attemptLimit, attemptPeriod, deferred)`. Then no tests were run.

mearns:
👌 Thanks @mustafaiman. We'll try to get that fix and the requested changes in shortly.

hasancelik:
verify

nitish661401869:
@mustafaiman The changes you suggested are in. I have also edited the Readme.md to mention that the linting step also needs to be run. This was why I had missed the semicolon. 

I am also planning to add a unit test around the logic that is being fixed.

gAmUssA:
@nitish661401869 @mearns PR builder results now publicly available.

mearns:
@gAmUssA Thanks, unfortunately Cloudbees still says I'm unauthorized, using Github authentication.

mustafaiman:
verify

mustafaiman:
@nitish661401869 thanks for your effort. Additional test is also very welcome. Can I ask you to sign our contributor agreement before merging your PR: https://hazelcast.atlassian.net/wiki/display/COM/Hazelcast+Contributor+Agreement ?

nitish661401869:
I was wondering when this will be merged, @mustafaiman ?

gAmUssA:
looks good me @mustafaiman 👍

gAmUssA:
@GhostInAMachine review and let's merge it

mustafaiman:
I did not realize you have already signed the agreement. I am merging it now. Thanks for your effort.

PR #144, opened 2016-10-26
Reliable Topic Implementation

GhostInAMachine:


src/Address.ts
```diff
@@ -1,15 +1,42 @@
...
+
+    readData(input: DataInput): any {
+        this.port = input.readInt();
+        // IPv4 or IPv6 - not used here
+        input.readByte();
+        this.host = input.readUTF();
+    }
+
+    writeData(output: DataOutput): void {
+        output.writeInt(this.port)
```
gAmUssA: @GhostInAMachine I  believe problem is here :)

GhostInAMachine:
verify

mustafaiman:
hey @GhostInAMachine . I see ReliableTopic requires a few more changes than we are used to. Can you add some description about why it is different than other proxy implementations?

GhostInAMachine:
@mustafaiman 

- It has some client-side configuration, since it is a wrapper around a ringbuffer
- The topic message have references to address of the sender, so I had to make it serializable
- Topic message has some complex behavior, so its source code is split into separate files for clarity

gAmUssA:
@mustafaiman are you good with merging this?

mustafaiman:
verify

PR #145, opened 2016-10-27
migration to Typescript 2.0

mustafaiman:
- Typescript 2.0 introduces `@types` repository support. This eliminates dependency for `typings` project.
- New version of mocha test runner throws exception when a test both
  - returns a promise
  - takes `done` callback as an argument
    in earlier versions, two methods raced to succeed or fail the test.
  
  We had a couple of this kind of tests so I fixed them.
- Documentation generator plugin (typedoc) does not work with this version yet. I hope they fix it soon. If they do not fix it until our next release, I think we can create another branch with TS 1.8 and generate documentation from there.
- I also updated other dependencies where availale.


gAmUssA:
issue about typedoc support for TS 2.0 https://github.com/TypeStrong/typedoc/issues/234

PR #148, opened 2016-10-31
TLS/SSL connection support

asimarslan:


asimarslan:
The committed pem file is from oss hazelcast repo in test resources

scripts/download-rc.sh
```diff
@@ -3,13 +3,23 @@ HAZELCAST_VERSION="3.6.5"
...
 RELEASE_REPO="http://repo1.maven.apache.org/maven2"
-#ENTERPRISE_REPO="https://repository-hazelcast-l337.forge.cloudbees.com/release/"
+ENTERPRISE_REPO="https://repository-hazelcast-l337.forge.cloudbees.com/release/"
 
 
 mvn dependency:get -DrepoUrl=${SNAPSHOT_REPO} -Dartifact=com.hazelcast:hazelcast-remote-controller:${HAZELCAST_RC_VERSION} -Ddest=hazelcast-remote-controller-${HAZELCAST_RC_VERSION}.jar
 mvn dependency:get -DrepoUrl=${RELEASE_REPO} -Dartifact=com.hazelcast:hazelcast:${HAZELCAST_VERSION} -Ddest=hazelcast-${HAZELCAST_VERSION}.jar
 
-nohup java -cp hazelcast-remote-controller-${HAZELCAST_RC_VERSION}.jar:hazelcast-${HAZELCAST_VERSION}.jar:test/javaclasses  com.hazelcast.remotecontroller.Main>rc_stdout.log 2>rc_stderr.log &
+if [ -n "${HAZELCAST_ENTERPRISE_KEY}" ]; then
```
asimarslan: Server download script now checks for hazelcast key on environment vars.

scripts/dev-test-rc.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
mustafaiman: do we use this file or is this a left over? it looks like a copy of download-rc.sh

src/Config.ts
```diff
@@ -22,6 +22,44 @@ export class SocketOptions {
     //TO-DO
 }
 
+export interface SslOptions {
```
mustafaiman: I would prefer this as a class rather than interface. For JS use this does not matter but if somebody uses this through Typescript they would like to instentiate an object of this and assign relevant parameters. This is meant to be used as an object, afterall, right?

src/invocation/ClientConnection.ts
```diff
@@ -42,14 +48,23 @@ class ClientConnection {
...
             // Send the protocol version
             var buffer = new Buffer(3);
             buffer.write('CB2');
             this.socket.write(buffer);
             ready.resolve(this);
-        });
+        };
+
+        if (this.clientNetworkConfig.sslOptions) {
+            let sslSocket = tls.connect(this.address.port, this.address.host, this.clientNetworkConfig.sslOptions, conCallback);
```
mustafaiman: There is a catch about `let` keyword. Node.js 4.x does not support `let` so our Typescript setup is configured to translate `let` keywords to `var` anyway. Since `let` is block-scoped and `var` is not, use of `let` may fool developers. Also this may cause hard to spot anomalies in behavior of the client if we move to newer versions of Node.js in future. So I prefer not to use `let` at all.

scripts/dev-test-rc.sh
asimarslan: I created it for handy development purpose to run test from ide to see the log on console etc.

src/Config.ts
asimarslan: There is a similar interface in `tls` module called `ConnectionOptions`, can you look at the source and re-evaluate, I don't know why they did it interface?

src/invocation/ClientConnection.ts
asimarslan: Ok, I removed all `let` usage.

src/Config.ts
asimarslan: I didn't realize that class's can not have optional field. So we want optional fields.

mustafaiman: I reconsidered it and I think interface is ok. 

Just last and a very minor point: Can we make name of the class `SSLOptions` since SSL is an abreviation?

PR #160, opened 2016-12-29
Adding support for ReplicatedMap

zemd:


gAmUssA:
Verify

gAmUssA:
verify

src/proxy/BaseProxy.ts
```diff
@@ -74,6 +74,19 @@ export class BaseProxy {
     }
 
     /**
+     * Encodes a request from a codec and invokes it on owner node of random partition.
+     * @param codec
+     * @param codecArguments
+     * @returns
+     */
+    protected encodeInvokeOnRandomPartition<T>(codec: any, ...codecArguments: any[]): Promise<T> {
```
mustafaiman: I do not think we need this method here. General approach in Hazelcast clients is to assign a random partition id to a proxy instance at first call and then use `encodeInvokeOnPartition` with that partition number for the rest of the calls in that proxy instance. You can see how it is done in hazelcast-python-client https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/proxy/replicated_map.py#L236

We try to keep implementations consistent across different clients.

src/invocation/ClusterService.ts
```diff
@@ -173,6 +174,17 @@ class ClusterService extends EventEmitter {
         return this.ownerConnection;
     }
 
+    /**
+     * Returns random connection
+     * @returns {Address}
+     */
+    getRandomAddress(): Address {
```
mustafaiman: I think this method is out of this pr's scope. See my comment about `encodeInvokeOnRandomPartition` method to see why.

src/invocation/InvocationService.ts
```diff
@@ -133,13 +140,13 @@ export class InvocationService {
...
-        } else if (invocation.hasOwnProperty('partitionId')) {
-            var address = this.client.getPartitionService().getAddressForPartition(invocation.partitionId);
+        } else if (invocation.hasPartitionId()) {
+            const address = this.client.getPartitionService().getAddressForPartition(invocation.partitionId);
             return this.sendToAddress(invocation, address);
         } else if (invocation.hasOwnProperty('address')) {
             return this.sendToAddress(invocation, invocation.address);
         } else {
-            return this.send(invocation, this.client.getClusterService().getOwnerConnection());
+            return this.sendToAddress(invocation, this.client.getClusterService().getRandomAddress());
```
mustafaiman: We deliberately send the message to the owner connection if there are no other bounds. It should not be random.
Also, this refactoring is out of this pr's scope.

src/proxy/ReplicatedMapProxy.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    remove(key: K): Promise<V> {
+        assertNotNull(key);
+
+        const keyData = this.toData(key);
+        return this.encodeInvokeOnKey<V>(ReplicatedMapRemoveCodec, keyData, keyData);
+    }
+
+    putAll(pairs: [K, V][]): Promise<void> {
+        let pair: [K, V];
```
mustafaiman: Regarding use of `const` and `let` keywords, our typescript compilation setup converts `let` and `const` keywords to `var`s. Use of these keywords may result in hard to spot bugs while maintaining the codebase. Can you use `var` instead.

test/replicatedmap/ReplicatedMapProxyTest.js
```diff
@@ -0,0 +1,130 @@
+var expect = require('chai').expect;
+var HazelcastClient = require('../../lib/index.js').Client;
+var Controller = require('./../RC');
+var Util = require('./../Util');
+var fs = require('fs');
+
+var Promise = require('bluebird');
+
+describe('ReplicatedMap Proxy', function () {
```
mustafaiman: In tests it is better if you do not use nested functions but return promises from each function. For example in `puts one entry and gets one entry` test, instead of 
```javascript
    it('should contain the key', function () {
        return rm.put('key', 'value', ONE_HOUR)
            .then(function (val) {
                return rm.containsKey('key')
                    .then(function (res) {
                        expect(res).to.equal(true);
                    });
            });
    });
```
you can write
```javascript
    it('should contain the key', function () {
        return rm.put('key', 'value', ONE_HOUR).then(function (val) {
            return rm.containsKey('key');
        }).then(function (res) {
            return expect(res).to.equal(true);
        });
    });
```

test/replicatedmap/ReplicatedMapProxyTest.js
```diff
@@ -0,0 +1,130 @@
...
+        return rm.put('key', 'value', ONE_HOUR)
+            .then(function (val) {
+                return rm.containsKey('key')
+                    .then(function (res) {
+                        expect(res).to.equal(true);
+                    });
+            });
+    });
+
+    // TODO(zemd): need to be verified separately due to strange NullPointerException from the server
```
mustafaiman: I think this test should pass. What is the problem here exactly? We can help you spot the problem.

test/replicatedmap/ReplicatedMapProxyTest.js
```diff
@@ -0,0 +1,130 @@
...
+                        return rm.remove('key')
+                            .then(function () {
+                                return rm.containsKey('key')
+                                    .then(function (contains) {
+                                        expect(contains).to.equal(false);
+                                    });
+                            });
+                    });
+            });
+    });
```
mustafaiman: can you also add tests for entry listener functionality?

test/replicatedmap/ReplicatedMapProxyTest.js
```diff
@@ -0,0 +1,130 @@
...
+});
```
mustafaiman: `putAll`, `keySet`, `entrySet` and `values` tests are missing.

src/invocation/InvocationService.ts
zemd: yep, I see, but during I was finding out what causes an exception in tests I found that in python library client sends to random address, have a look https://github.com/hazelcast/hazelcast-python-client/blob/master/hazelcast/invocation.py#L154

Do I need to return this code back?

src/proxy/IReplicatedMap.ts
```diff
@@ -0,0 +1,156 @@
...
+     * Returns a view of the key contained in this map.
+     *
+     * @return keys of this map as an array.
+     */
+    keySet(): Promise<K[]>;
+
+    /**
+     * Returns an array of values contained in this map.
+     */
+    values(): Promise<V[]>;
```
zemd: I don't find any references on Comparator usage with not predicate methods + I don't find it in HazelcastOpenBinaryClientProtocol-1.2.0.pdf and python client. Could you point me how to use it? Because I think Comparator interface from the core does not suit in this case, it compares 2 arrays as it's arguments. But here we have array of V.

I can add something like 
```javascript
interface ArrayComparator<T> {
(a: T, b: T): number;
}

values(comparator: ArrayComparator<V>): Promise<V[]> {
//...
if (comparator) {
return results.sort(comparator);
}
}
```
Will it be ok?

src/proxy/ReplicatedMapProxy.ts
zemd: I've been using `const/let` more than a year in different environments and compiled to es5, and had no issues related to them. But in opposite, I highly recommend to use them to reduce bugs during compilation time, even more - most of IDEs can highlight you dangerous code like assigning new value to the constant.

just made several examples:

```javascript
for (let i = 0; i < 10; i += 1) {
//
}
console.log(i);
```
`npm run compile` => exits with 1 and message "test.ts(4,13): error TS2304: Cannot find name 'i'."

```javascript
let a = 1;
let b = 2;
if (true) {
    let a = 3;
}

console.log(a);
```

`npm run compile` => compiled successfully to

```javascript
var a = 1;
var b = 2;
if (true) {
    var a_1 = 3;
}
console.log(a);
```
as you see variable "a" resolved correctly. 

But in case of using `var` you **won't have any compiler notice** and potentially will have much more hard spotted bugs.

test/replicatedmap/ReplicatedMapProxyTest.js
zemd: will add them shortly

test/replicatedmap/ReplicatedMapProxyTest.js
zemd: The problem is exception from server: https://gist.github.com/zemd/e1476c618caf92563dc018597b208682

As I noticed in hazelcast java core there are no so many things for ContainsKeyOperation which can throw such exception, but I haven't figured out.

src/proxy/IReplicatedMap.ts
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Associates a given value to the specified key and replicates it to the
+     * cluster. If there is an old value, it will be replaced by the specified
+     * one and returned from the call.
+     *
+     * @param key      key with which the specified value is to be associated.
+     * @param value    value to be associated with the specified key.
+     * @param ttl      ttl to be associated with the specified key-value pair.
+     */
+    put(key: K, value: V, ttl: number): Promise<V>;
```
mustafaiman: Yes, using a default value of 0 in implementation is the way to go. My comment is only for interface specification. If you specify the interface as `put(key: K, value: V, ttl?: number)`, users immediately understand that they do not have to provide a ttl value by looking at the interface.

src/proxy/IReplicatedMap.ts
mustafaiman: Comparator class from `core` was meant to be used in predicates. I thought it would suit the usage in here but I see it does not.
We can use the definition of comparator function from es6 type definition. It goes like this `(a: T, b: T) => number`. Then it is easy to implement it. You just call `values()`, then use array.sort method of Javascript with the given comparator function.

src/proxy/ReplicatedMapProxy.ts
mustafaiman: I did not know typescript compiler detected the misuse of `let`, `const`. I tried it now. It is ok. Thanks :)

src/invocation/InvocationService.ts
zemd: ok

zemd:
@mustafaiman Hi, please review once again my updates

src/proxy/IReplicatedMap.ts
```diff
@@ -0,0 +1,171 @@
...
+     * cluster. If there is an old value, it will be replaced by the specified
+     * one and returned from the call.
+     *
+     * @param key key with which the specified value is to be associated.
+     * @param value value to be associated with the specified key.
+     * @param ttl ttl to be associated with the specified key-value pair.
+     * @param unit
+     * @return old value if there was any, `null` otherwise.
+     */
+    put(key: K, value: V, ttl: Long|number|string, unit: TimeUnit): Promise<V>;
```
mustafaiman: I meant adding timeunit(which was milliseconds) to documentation when I said it needs a Timeunit. Although this Timeunit option is great, we need to add this to every other proxy at once if we add it here. Can we remove TimeUnit and just add word `(milliseconds)` to documentation of `ttl`.

mustafaiman:
@zemd Can you sign Hazelcast Contributor Agreement so we can merge your PR: https://hazelcast.atlassian.net/wiki/display/COM/Hazelcast+Contributor+Agreement

mustafaiman:
@zemd I tried the failing test you commented out. I could not reproduce the problem here: https://github.com/hazelcast/hazelcast-nodejs-client/pull/162. Do you still see the test failing in your local? Do you have any more information on how to reproduce this problem?

mustafaiman:
I do not see any tests verifying the following:
- `put` with no ttl specified
- an entry put with ttl actually gets evicted after ttl.
- the various callback functions of entrylistener such as `REMOVED`, `UPDATED`, etc.

src/proxy/IReplicatedMap.ts
zemd: 😆 ok, I will add documentation note, I've seen this timeunit inside hazelcast java so thought that you mean exactly this TimeUnit class :)

`string` indeed is good option for users, because it let people who are using bigint libraries to pass the result of these libs(usually they output strings).

zemd:
> @zemd I tried the failing test you commented out. I could not reproduce the problem here: #162. Do you still see the test failing in your local? Do you have any more information on how to reproduce this problem?

yes, it is still exists. but I see it randomly, and sometimes it passes fine without any errors, what type of information I could give you?

src/proxy/IReplicatedMap.ts
```diff
@@ -0,0 +1,169 @@
...
+     * Associates a given value to the specified key and replicates it to the
+     * cluster. If there is an old value, it will be replaced by the specified
+     * one and returned from the call.
+     *
+     * @param key key with which the specified value is to be associated.
+     * @param value value to be associated with the specified key.
+     * @param ttl milliseconds to be associated with the specified key-value pair.
+     * @return old value if there was any, `null` otherwise.
+     */
+    put(key: K, value: V, ttl: Long|number|string): Promise<V>;
```
mustafaiman: I still argue against accepting string as a valid argument. If we do this, then it is our responsibility to convert this string to a valid number. Users can provide too big numbers or malformed inputs as strings. This conversion is out of this client's interest. Even if I accept that, I do not see the relevant code that does the verification and conversion in case of string `ttl` argument. If we put that code here, we need to maintain it. Also, accepting a string in `ttl` area does not seem semantically correct to me. `ttl` is semantically a number(actually a non-negative integer but JS allows us to restrict it only to the point of a number). I see this as more of a burden to maintainers than it is a convenience to users.

zemd: "long" library provides maintenance of accepting strings - Long.fromValue accepts both strings and numbers, and keep in mind that `ttl` by spec should be int64, but js supports int32 only, thus strings are the rescue to support big numbers.

zemd: @mustafaiman I will apply `toMillis` function to this place and others ones in next pull request, so casting string to long will be made early.

mustafaiman: JS numbers are doubles. So they support integers upto 2^53-1=9007199254740991. If you assume this number as number of milliseconds, you can represent approximately 285000 years

zemd: well... yes I wasn't accurate, int53, but my favorite "sanitizing" bitwise operation `|0` returns always int32, that's why I told about int32. removed `string`.

src/util/TimeUnit.ts
```diff
@@ -0,0 +1,35 @@
+import Long = require('long');
```
mustafaiman: TimeUnit is not used anymore, right? We can delete it.

zemd: done

mustafaiman:
verify

gAmUssA:
@zemd Thanks for your contribution!

PR #161, opened 2016-12-31
Adding semaphore support

zemd:
WIP, tests are required

ps this code is branched from master where ReplicatedMap implemented. After it will be merged this pull request will show correct diff.


gAmUssA:
@zemd could you rebase with master since ReplicatedMap branch was merged?

mustafaiman:
@gAmUssA he already did that. I will review this soon.

src/proxy/ISemaphore.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export interface ISemaphore extends DistributedObject {
+    /**
+     * Try to initialize this ISemaphore instance with the given permit count.
+     * @param permits
+     */
+    init(permits: number): Promise<boolean>;
+
+    /**
+     * Acquires the given number of permits if they are available, and returns
```
mustafaiman: I see this is copied from Java client's documentation. Unfortunately, there are major differences between Java and Node.js with respect to this proxy. Java client blocks the calling thread until semaphore acquire succeeds. Node.js client should fulfill the returned promise when acquire request succeeds, otherwise promise is simply not fulfilled. This client already behaves this way as far as I see but documentation is wrong.

src/proxy/ISemaphore.ts
```diff
@@ -0,0 +1,66 @@
...
+     * some other thread the current thread. If the current thread has its interrupted
+     * status set on entry to this method, or is while waiting for a permit, then is
+     * thrown and the current thread's interrupted status is cleared.
+     * @param permits
+     */
+    acquire(permits: number): Promise<void>;
+
+    /**
+     * Returns the current number of permits currently available in this semaphore.
+     * This method is typically used for debugging and testing purposes.
```
mustafaiman: We do not have to state that.

src/proxy/ISemaphore.ts
```diff
@@ -0,0 +1,66 @@
...
+    availablePermits(): Promise<number>;
+
+    /**
+     * Acquires and returns all permits that are immediately available.
+     */
+    drainPermits(): Promise<number>;
+
+    /**
+     * Shrinks the number of available permits by the indicated reduction.
+     * This method differs from acquire in that it does not block waiting
```
mustafaiman: Again, neither acquire nor this thread should block the thread.

src/proxy/ISemaphore.ts
```diff
@@ -0,0 +1,66 @@
...
+    reducePermits(reduction: number): Promise<void>;
+
+    /**
+     * Releases the given number of permits, increasing the number of available
+     * permits by that amount. There is no requirement that a thread that releases a
+     * permit must have acquired that permit by calling one of the acquire()acquire
+     * methods. Correct usage of a semaphore is established by programming convention
+     * in the application
+     */
+    release(permits: number): Promise<void>;
```
mustafaiman: We can make `permits` an optional parameter. Default value should be 1 in this case. What I mean is put `release(permits?: number)` in the interface instead of this method and in the implementation method signature should be `release(permits: number = 1)`

src/proxy/ISemaphore.ts
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Acquires the given number of permits, if they are available, and returns
+     * immediately, with the value true, reducing the number of available permits
+     * by the given amount. If insufficient permits are available then this
+     * method will return immediately with the value false and the number of
+     * available permits is unchanged.
+     * @param permits
+     * @param timeout
+     */
+    tryAcquire(permits: number, timeout: Long|number|string): Promise<boolean>;
```
mustafaiman: the same comment about optional `permits` parameter as in `release` method. Plus, we do not accept `string` values for timeout.

src/proxy/ISemaphore.ts
```diff
@@ -0,0 +1,66 @@
...
+     * disabled for thread scheduling purposes and lies dormant until one of three
+     * things happens: some other thread invokes one of the methods for this semaphore,
+     * the current thread is next to be assigned permits and the number of available
+     * permits satisfies this request, this ISemaphore instance is destroyed, or
+     * some other thread the current thread. If the current thread has its interrupted
+     * status set on entry to this method, or is while waiting for a permit, then is
+     * thrown and the current thread's interrupted status is cleared.
+     * @param permits
+     */
+    acquire(permits: number): Promise<void>;
```
mustafaiman: possible optional `permits` parameter. See the comment on `release` method.

test/semaphore/SemaphoreProxyTest.js
```diff
@@ -0,0 +1,126 @@
...
+                return semaphore.tryAcquire(10);
+            })
+            .then(function (res) {
+                expect(res).to.equal(true);
+                return semaphore.availablePermits();
+            })
+            .then(function (res) {
+                expect(res).to.equal(0);
+            });
+    });
```
mustafaiman: can you add tests for negative scenarios, too? For example we want make sure that one of the 2 acquire calls does not succeed if the semaphore has only 1 permit. Another scenario I want to test goes like this: Start 2 clients. Initiate a sempahore with 1 permit. Try to acquire the semaphore and make sure that only one of them proceeds. Also, in another test I would like to see the other client is able to proceed after the first client releases a permit on a semaphore both clients race to acquire. There needs to be a handful of these tests. The actual use case of this proxy is this.

.gitignore
```diff
@@ -24,7 +24,7 @@ npm-debug.log
 
 # TypeScript
 *.js.map
-lib/*
+lib/
```
mustafaiman: Why do you need to change this?

mustafaiman:
@zemd I squashed your commits and did the fixes I requested myself. I preserved your commit in the new branch and added a new commit to do the fixes. Thank you for your effort. You can track the status of semaphore here: https://github.com/hazelcast/hazelcast-nodejs-client/pull/165

zemd:
@mustafaiman good job 👍 sorry, had no time to fix your review comments.

Issue #157, opened 2016-12-17
read emoji string has error


cycerto:
when try to read the following string
```
let str1 =  "1⚐中💦2😭‍🙆😔5"
let str2 = "﹏ JOJO糾ั͡✾✎"
```

error emit:

```
fetch caster-store data (key: (instance1) key-1) has error:  { AssertionError: false == true
    at ObjectDataInput.assertAvailable (/Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/serialization/ObjectData.js:261:9)
    at ObjectDataInput.read (/Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/serialization/ObjectData.js:274:14)
    at ObjectDataInput.readByte (/Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/serialization/ObjectData.js:289:21)
    at ObjectDataInput.readUTF (/Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/serialization/ObjectData.js:426:36)
    at StringSerializer.read (/Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/serialization/DefaultSerializer.js:16:22)
    at SerializationServiceV1.toObject (/Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/serialization/SerializationService.js:45:27)
    at MapProxy.BaseProxy.toObject (/Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/proxy/BaseProxy.js:54:54)
    at Function.MapGetCodec.decodeResponse (/Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/codec/MapGetCodec.js:32:38)
    at /Users/cycerto/ProjectWorkspace/webstorm/caster-client-sdk-nodejs/node_modules/hazelcast-client/lib/proxy/BaseProxy.js:12:33
[...]

PR #163, opened 2017-01-09
fix utf8 serialization/deserialization algorithm

mustafaiman:
fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/157

JS strings use surrogate pairs to represent characters spanning more than 2 bytes. However `String.toBuffer` function of JS converts a surrogate pair into unicode code point first and then encodes it as a single character. Hazelcast serialization encodes surrogate pairs as two individual characters. This PR replaces default JS serialization with custom utf8 serialization.

Adds new test cases for international utf8 strings


mustafaiman:
verify

Issue #175, opened 2017-05-06
In ReliableTopic, a message is delivered to the message listener even though the listener has been removed.


kishoreannapureddy:
In ReliableTopic, a message is delivered to the message listener even though the listener has been removed. The listener gets removed after that.

This is because of not checking for 'cancelled' in the ring buffer's readMany's then() function in the ReliableTopicListenerRunner's next() method

gAmUssA:
@kishoreannapureddy could you provide a reproducer?

PR #183, opened 2017-10-04
reliable topic listener fix

mustafaiman:
Reliable topic may continue to push the events to listeners even after they are removed.
Fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/175

### DIFF
--- a/src/proxy/topic/ReliableTopicListenerRunner.ts
+++ b/src/proxy/topic/ReliableTopicListenerRunner.ts
@@ -36,16 +36,18 @@ export class ReliableTopicListenerRunner<E> {
         }
 
         this.ringbuffer.readMany(this.sequenceNumber, 1, this.batchSize).then((result: Array<RawTopicMessage>) => {
-            result.forEach((raw: RawTopicMessage) => {
-                var msg = new TopicMessage<E>();
-                msg.messageObject = this.serializationService.toObject(raw.payload);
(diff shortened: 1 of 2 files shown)


src/proxy/topic/ReliableTopicListenerRunner.ts
```diff
@@ -36,16 +36,18 @@ export class ReliableTopicListenerRunner<E> {
...
         this.ringbuffer.readMany(this.sequenceNumber, 1, this.batchSize).then((result: Array<RawTopicMessage>) => {
-            result.forEach((raw: RawTopicMessage) => {
-                var msg = new TopicMessage<E>();
-                msg.messageObject = this.serializationService.toObject(raw.payload);
-                msg.publisher = raw.publisherAddress;
-                msg.publishingTime = raw.publishTime;
-                setImmediate(this.listener, msg);
-                this.sequenceNumber++;
-            });
+            if (!this.cancelled) {
```
asimarslan: This cancelled check is in the for loop on java implementation, can you check it?

mustafaiman: For Node.js it does not matter. If `this.cancelled` is false before going into `forEach` loop below, it will stay false until all items in `result` are processed because the loop will not be interrupted (single thread execution of node.js)

PR #191, opened 2017-10-17
Fix listener service

mustafaiman:
Previously, listener service did not register listeners to all members. It used only one member. Also, when that member died, it did not re-register the listeners to another member. Therefore, listeners are lost after reconnection.

This PR changes ListenerService so that

- all listeners are registered to all members and they receive local-only events.
- when a new connection is established, all existing listeners are registered to the new member
- client periodically verifies connection status to all known members
- listener service internal api is changed
- all proxies implement new internal api

- PR also includes a small fix in invocation service that caused listener service to brake. When ownerconnection is null, invocationservice retries the operation.

~~One test does not pass because of a bug in partition service. This PR will be green after merging https://github.com/hazelcast/hazelcast-nodejs-client/pull/194 and rebasing.~~

Fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/167


mustafaiman:
verify

src/invocation/InvocationService.ts
```diff
@@ -322,43 +351,221 @@ export class InvocationService {
 /**
  * Handles registration and de-registration of cluster-wide events listeners.
  */
-export class ListenerService {
+export class ListenerService implements ConnectionHeartbeatListener {
```
asimarslan: Can we move this class to its own file ?

src/invocation/InvocationService.ts
```diff
@@ -322,43 +351,221 @@ export class InvocationService {
...
+        this.activeRegistrations.forEach((registrationsOnUserKey: Map<ClientConnection, ClientEventRegistration>,
+                                          userKey: string) => {
+            var eventRegistration: ClientEventRegistration = registrationsOnUserKey.get(connection);
+            if (eventRegistration !== undefined) {
+                this.client.getInvocationService().removeEventHandler(eventRegistration.correlationId.toNumber());
+            }
+        });
+    }
+
+    invokeRegistrationFromRecord(userRegistrationKey: string, connection: ClientConnection): Promise<ClientEventRegistration> {
```
asimarslan: is this method share any code with `registerListenerInternal` ?  A shared invoke method will increase readability, if possible.

src/proxy/MapProxy.ts
```diff
@@ -552,32 +552,42 @@ export class MapProxy<K, V> extends BaseProxy implements IMap<K, V> {
...
         var request: ClientMessage;
+        var encodeFunc: Function;
         var handler: Function;
         var responser: Function;
         if (key && predicate) {
             var keyData = this.toData(key);
             var predicateData = this.toData(predicate);
-            request = MapAddEntryListenerToKeyWithPredicateCodec.encodeRequest(this.name, keyData,
-                predicateData, includeValue, flags, false);
+            encodeFunc = (localOnly: boolean) => {
```
asimarslan: Doesn't the encodeFunc always create the same message? so why not just use that encoded clientMessage instead and save some cpu cycles ?

src/invocation/InvocationService.ts
mustafaiman: This method takes the record of a previously registered listener and invokes the registration again when a connection is restored whereas `registerListenerInternal` registers a register for the first time. `registerListenerInternal` creates records of listeners while this method uses those records. Invocation part is only `this.client.getInvocationService().invoke(invocation)` part. The actions taken at the result of invocation are also different. It is not worth extracting another method and sharing it.

src/invocation/RegistrationKey.ts
```diff
@@ -0,0 +1,42 @@
+export class RegistrationKey {
+    private readonly userRegistrationId: string;
+    private handler: Function;
+    private encoder: Function;
```
asimarslan: Can we rename all encoder and decoder naming to give the meaning of

* `registration Encoder Function` ( something like `registerEncodeFunc`) 
* `registration Decoder Function`
* `deregistration Encoder Function`
* `deregistration Decoder Function`

I can see encode decode keywords everywhere, it's hard to follow.

mustafaiman: did that.

src/proxy/MapProxy.ts
mustafaiman: I think it can be done that way. We actually needed to pass encode function to listenerservice for only de-registration because listener-id that user sees and actual listener-id are different. So I made both registration and de-registration apis similar.

mustafaiman:
@asimarslan I changed `registerListener` api so it takes `ClientMessage` now instead of register listener encode function.

mustafaiman:
verify

src/ListenerService.ts
```diff
@@ -0,0 +1,239 @@
...
+            }).then(() => {
+                this.activeRegistrations.delete(userRegistrationKey);
+                this.userRegistrationKeyInformation.delete(userRegistrationKey);
+            });
+        });
+
+        return deferred.promise;
+    }
+
+    private trySyncConnectToAllConnections(): Promise<void> {
```
asimarslan: This method should timeout after invocation Timeout period.

PR #197, opened 2017-10-24
Codec update

mustafaiman:
Updates codecs to client-protocol version 1.5.0-9.

Client.getPartitionsCodec requires client to report its version now. Therefore we needed a build info file that contains version of the client at runtime. For this purpose, I added a precompile script that puts build info in output directory.

Hand edited files:
- create-build-info.js
- package.json
- ConnectionAuthenticator.ts
- BuildInfoLoader.ts
- MultiMapProxy.ts
- LockProxy.ts
- HazelcastClient.ts
- LockReferenceIdGenerator.ts
- MultiMapProxyListenersTests.js

The rest are generated.


mustafaiman:
verify

Issue #158, opened 2016-12-22
Serialization of nested objects which have different version numbers


mustafaiman:
https://github.com/hazelcast/hazelcast/pull/9415

There is no such issue in this client. However it is not tested. It needs a test.

PR #198, opened 2017-10-24
test for nested different versions portable predicate

mustafaiman:
fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/158
Looks like node.js client already works with this scenario.


asimarslan:
Do we need compiled class files ?

mustafaiman:
@asimarslan I precompiled them against Hazelcast to use in tests. Otherwise, we need to download Hazelcast java library and recompile the java files every time. I think that is unneccessary overhead for testing.

PR #200, opened 2017-10-26
ListenerService refactor

mustafaiman:
ListenerService does not check the result of RemoveListener request coming from the server.
I also refactored the code to follow Java client. So addListener method of each proxy creates a ListenerMessageCodec which has methods like encodeAddRequest. We pass this object around. It makes the interface much clearer.

One difference is that this implementation creates AddListener client-message only once, when proxy's method is invoked. It re-uses the same message for sending requests to all members.


mustafaiman:
verify

PR #206, opened 2017-11-02
Fix MembershipListener test

mustafaiman:
in `sees member removed event` test, sometimes assertions are called before listener callback runs. This makes test fragile. Now assertions wait for callback to be called first.

Also changes `assert` interface to `expect` interface for uniformity of tests.


mustafaiman:
verify

mustafaiman:
verify

PR #216, opened 2017-11-14
Ssl mutual authentication

mustafaiman:
SSL support in Node.js depends on  user providing an `SSLOptions` object in  configuration which includes certificates, ca, etc. This options are provided to Node.js runtime when establishing a new connection and the runtime handles all of it. Therefore, there is no change needed in the codebase  for authentication. This PR only adds tests.

Includes changes in https://github.com/hazelcast/hazelcast-nodejs-client/pull/215


mustafaiman:
verify

PR #224, opened 2017-12-01
Eventually consistent near cache (single invalidation)

mustafaiman:
Working Stale Read Detector and Anti Entropy Task implementation. Left following two issues for another PR as this one is already very large.
- Batch invalidation
- Read your own writes consistency

Also `InvalidationMetadataDistortionTest` is skipped because it needs `DistortInvalidationMetadataEntryProcessor` to be merged into Hazelcast repo first.

InvalidationMetadataDistortionTests requires https://github.com/hazelcast/hazelcast/pull/12052 to pass


scripts/download-rc.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-HZ_VERSION="3.9"
+HZ_VERSION="3.10-SNAPSHOT"
```
asimarslan: We should use the latest maintenance release instead of a snapshot unless we have to.

src/nearcache/MetadataFetcher.ts
```diff
@@ -0,0 +1,84 @@
...
+        partitionIdSequenceList.forEach((partitionIdSeq: [string, [number, Long][]]) => {
+            let pairs = partitionIdSeq[1];
+            pairs.forEach((pair: [number, Long]) => {
+                handler.checkOrRepairSequence(pair[0], pair[1]);
+            });
+        });
+    }
+
+    protected scanMembers(objectNames: string[]): Promise<ClientMessage>[] {
+        let members = this.client.getClusterService().getMembers();
```
ahmetmircik: Are these only data `members` as in [java client](https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/invalidation/ClientMapMetaDataFetcher.java#L64)?

mustafaiman:
@ahmetmircik i addressed your concern. Do you have further comments?

ahmetmircik:
@mustafaiman no further comment, thanks

mustafaiman:
verify

mustafaiman:
verify

PR #225, opened 2017-12-06
adds copyright

mustafaiman:
This adds copyright notice to all files.


asimarslan:
Can we update the date range to include 2018. I think we can wait one more week to merge this.

PR #234, opened 2017-12-28
Prevent delayed promise resolve from overwriting entries in nearcache

mustafaiman:
Execution order of resolved promises after a remote `get` or `remove` operation returned is not guaranteed. Therefore a get operation might put a stale value back into near cache after remove operation deleted it. This PR prevents such a scenario by not putting the result of get operation into near cache in case there was an intervening remove operation. The mechanism is identical to stale read prevention in Java client.


mustafaiman:
verify

PR #237, opened 2018-01-02
Refactor Hearbeat Service

mustafaiman:
Fixed comparing `Address`s incorrectly by `==`. They are now compared using `Address.equals()`

### DIFF
--- a/src/invocation/ClusterService.ts
+++ b/src/invocation/ClusterService.ts
@@ -131,7 +131,7 @@ export class ClusterService extends EventEmitter {
 
     private onConnectionClosed(connection: ClientConnection) {
         this.logger.warn('ClusterService', 'Connection closed to ' + connection.address.toString());
-        if (connection.address === this.getOwnerConnection().address) {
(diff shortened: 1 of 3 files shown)


src/invocation/ClusterService.ts
```diff
@@ -131,15 +131,15 @@ export class ClusterService extends EventEmitter {
 
     private onConnectionClosed(connection: ClientConnection) {
         this.logger.warn('ClusterService', 'Connection closed to ' + connection.address.toString());
-        if (connection.address === this.getOwnerConnection().address) {
+        if (connection.address.equals(this.getOwnerConnection().address)) {
```
asimarslan: can connection or address be null at here?

mustafaiman: No. `connectionClosed` event is emitted here: https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/invocation/ClientConnectionManager.ts#L112 . It checks there exists such a connection before emitting the event.

PR #240, opened 2018-01-03
Invocation service fails retrying invocations when client is not active

mustafaiman:
* InvocationService retries the invocations until they are expired even when client is not active anymore.

Realized this issue when HeartbeatTest failed due to previous test's invocations affecting the next test.


mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

PR #241, opened 2018-01-09
Server version check in tests

mustafaiman:
adds a utility method to skip tests if expected server version is not satisfied

Without the check newly added eventually consistent near cache tests fail in client compatibility test suite against older server versions.


mustafaiman:
Jenkins refuses to run

mustafaiman:
verify

mustafaiman:
verify

PR #243, opened 2018-01-10
PagingPredicate fix

mustafaiman:
`CustomComparator` is javascript counterpart of server side `CustomComparator` test class. JS version's sort method is very short and **wrong**. CustomComparator on Java side compares entries after converting key and value to strings. JS version should do the same. Line
```
o1[0] - o2[0]
```
works for strings as Java's `String.compareTo` to some extent. Except when the object is a string that represents a number such as `"39"`. Javascript type coercion converts this string to a number when it sees the subtract operation. Therefore the operation in `sort` was actually a subtract operation between two numbers instead of string comparison in this test. So the assumption of this test was wrong. Therefore this PR fixes the test.
Thankfully, there is nothing wrong with the implementation. Only the test was broken.

This PR also puts `IterationType` object in `index` so that the users can import it and use it when they implement custom comparators. That was not possible before.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,6 +22,7 @@ import * as Predicates from './core/Predicate';
 import Address = require('./Address');
 import TopicOverloadPolicy = require('./proxy/topic/TopicOverloadPolicy');
 import * as HazelcastErrors from './HazelcastError';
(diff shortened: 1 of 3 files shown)


mustafaiman:
verify

PR #244, opened 2018-01-10
increases timeout for ClusterServiceTests.js. This test failed with 3…

mustafaiman:
….6 server because that server was slower to shutdown compared to newer servers


mustafaiman:
verify

PR #245, opened 2018-01-11
updates tsdoc

mustafaiman:
Previous typedoc had problems generating the docs. We had to move out some dependencies by hand to have typedoc compile the files. New version does not have this problem. Finally, it takes only a single command to generate the docs.


asimarslan:
verify

PR #250, opened 2018-01-26
aggregations

mustafaiman:
Adds fast aggregations + tests + API Doc
Code sample will follow up.

PR #254, opened 2018-02-09
simplify code samples

mustafaiman:
Before code samples loaded client module using `require('../.')` because they actually refer to the same module that they are present in. I think this causes confusion for our users. I created a dummy `hazelcast-client` module inside code_sample folder. This dummy client actually loads the actual module in this repo. So when we do `require('hazelcast-client')` inside `code_samples` the samples refer to the actual implementation. I think this looks much tidier and easy to use for our users.

Also, I simplified samples for map, set, queue and list. They had too much unnecessary indirection for simple code samples.

Update: Adds code sample for newly implemented Aggregators. https://github.com/hazelcast/hazelcast-nodejs-client/pull/250

### DIFF
--- a/code_samples/aggregation.js
+++ b/code_samples/aggregation.js
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+ *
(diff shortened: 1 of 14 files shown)


mustafaiman:
verify

mustafaiman:
verify

PR #258, opened 2018-02-14
refactor cluster connection

mustafaiman:
Takes out connection establish functionality from `ClientConnection` and puts it into `ClientConnectionManager`.

### DIFF
--- a/src/invocation/ConnectionAuthenticator.ts
+++ b/src/invocation/ConnectionAuthenticator.ts
@@ -43,7 +43,7 @@ export class ConnectionAuthenticator {
             .then((msg: ClientMessage) => {
                 var authResponse = ClientAuthenticationCodec.decodeResponse(msg);
                 if (authResponse.status === 0) {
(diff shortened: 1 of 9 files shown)


mustafaiman:
verify

src/invocation/ConnectionAuthenticator.ts
```diff
@@ -43,7 +43,6 @@ export class ConnectionAuthenticator {
             .then((msg: ClientMessage) => {
                 var authResponse = ClientAuthenticationCodec.decodeResponse(msg);
                 if (authResponse.status === 0) {
-                    this.connection.address = authResponse.address;
```
sancar: The address coming from server needs to be set to connection and connection needs to be stored to `establishedConnections` with this address. Because this address is guaranteed to be consistent with member list. Users configured address may not.

mustafaiman: I put back old behaviour.

PR #263, opened 2018-02-20
adds default export support to import configs

mustafaiman:
When `exportedName` is not specified in Json config, dynamic module loader loads default export.


mustafaiman:
verify

PR #272, opened 2018-03-05
Differentiate between remote errors

mustafaiman:
Adds error factories
All remote exceptions were wrapped by RemoteException. Client should be able to differentiate between server exceptions so it can determine when to retry.


mustafaiman:
verify

PR #274, opened 2018-03-07
Full review of the Config.md

Serdaro:


mustafaiman:
verify

CONFIG.md
```diff
@@ -145,40 +159,50 @@ established connection between the client and that member goes down.
...
 ```
+
 Default value is 3000.
+
 ### Enabling Client TLS/SSL
+
 You can use TLS/SSL to secure the connection between the client and members. If you want TLS/SSL enabled
 for the client-cluster connection, you should set an SSL configuration. Once set, the connection (socket) is 
-established out of an options object supplied by the user.
+established out of an `options`" object supplied by the user.
```
mustafaiman: leftover `"`

PR #277, opened 2018-03-12
adds filter argument to RingBuffer.ReadMany

mustafaiman:
adds filter argument to RingBuffer.ReadMany
fixes return value of RingBuffer.ReadMany

Depends on https://github.com/hazelcast/hazelcast/pull/12575


mustafaiman:
verify

PR #279, opened 2018-03-13
Fix/configpattern

mustafaiman:
Adds config pattern matching.

Also, it fixes a reliable topic proxy that I found while implementing this. I realized that reliable topic proxy was not created by proxy manager. This may cause creating a new proxy for the same reliable topic every time `client.getReliableTopic` is called.


mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

PR #280, opened 2018-03-14
Fixes FlakeIdGenerator proxy test

mustafaiman:
- We add small wait to the test:`ids are from new batch after prefetched ones are exhausted`. Even when the client correctly requests and receives a new batch, the server may send a new batch within the same millisecond range. So the new batch immediately follows the old one. In that case the test has no way of verifying if the batch is new. When we add a few milliseconds of wait time, server is guaranteed to send a new batch with a later millisecond range.

- sort function in `unique long` test implicitly did a string conversion and string comparison for flake id array. It now does a proper long comparison.


mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

PR #301, opened 2018-06-08
fix SSL tests

### LINKED ISSUE #300: SSL Client Authentication Test fails
`ma:required, server does not know client, client knows server should fail` test is expected to rejected with `IllegalStateError` but it returns client.

furkansenharputlu:
In this pr, the bug in the `SSL Client Authentication Test` is fixed. `maRequiredXML` and `maOptionalXML` weren't being used. I actived them.

The other problem was that client and server handshake wasn't provided due to lack of keys. I generated `keys` by using `client1.keystore` and `client2.keystore` on here:

https://github.com/hazelcast/hazelcast/tree/master/hazelcast/src/test/resources/com/hazelcast/nio/ssl-mutual-auth 

`keystore`s contain private keys and certificates. I extracted them as `client1-key.pem` and `client1-cert.pem` by getting help from [here](https://serverfault.com/questions/715827/how-to-generate-key-and-crt-file-from-jks-file-for-httpd-apache-server). Then, I used these files to connect to server and it worked.

However, I am not comfortable with using private keys to connect to server. It works but I'm not sure it is the correct way. What do you think ? @mustafaiman @sancar 

fixes #300


test/ssl/ClientSSLAuthenticationTest.js
```diff
@@ -88,62 +90,61 @@ describe('SSL Client Authentication Test', function () {
...
             it('ma:optional, server knows client, client does not know server should fail', function () {
-                return createMemberWithXML(maRequiredXML).then(function () {
-                    return expect(Client.newHazelcastClient(createClientConfigWithSSLOpts('./client1.pem', './server2.pem')))
+                return createMemberWithXML(maOptionalXML).then(function () {
+                    return expect(Client.newHazelcastClient(createClientConfigWithSSLOpts('./client1-key.pem', './client1-cert.pem', './server2-cert.pem')))
                         .to.be.rejectedWith(HzErrors.IllegalStateError);
                 });
             });
 
             it('ma:optional, server does not know client, client knows server should connect', function () {
```
mustafaiman: the name of the test should change too

PR #320, opened 2018-07-25
add HazelcastCloudDiscovery code sample

### LINKED ISSUE #318: Hazelcast Cloud Discovery Configurations
The implementation will be out with 3.10.3 and 3.11
But we want to document this feature when hazelcast.cloud is out. So adding this to referance manual will wait for the hazelcast.cloud release. 

Note that the configurations are different in 3.10.3 and 3.11.

### LINKED ISSUE #319: Hazelcast Cloud Discovery Sample
We need to add a code sample for cloud discovery when hazelcast.cloud is out.

furkansenharputlu:
This pr adds HazelcastCloudDiscovery feature to `README.md` and its configuration to `CONFIG.md`. It also adds a sample usage of HazelcastCloudDiscovery.

 fixes #318 
 fixes #319

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -20,7 +20,7 @@ Refer to [Hazelcast Node.js Client API Docs](http://hazelcast.github.io/hazelcas
 # Declarative Configuration (JSON)
 For declarative configuration, the Hazelcast client looks at the following places for the client configuration file.
 
-1. Environment variable: The client first looks for the environment variable `HAZELCAST_CLIENT_CONFIG`. If it exists, 
+1. Environment variable: The client first looks for the environment variable `HAZELCAST_CLIENT_CONFIG`. If it exists,
 the client looks for the configuration file in the specified location.
 2. Current working directory: If there is no environment variable set, the client tries to load `hazelcast-client.json`
 from the current working directory.
@@ -147,7 +147,7 @@ Default value is 5000 milliseconds.
 ### Setting Connection Attempt Limit
 
 While the client is trying to connect initially to one of the members in the  address list, that member
-might not be available at that moment. Instead of giving up, throwing an error and stopping the client, 
+might not be available at that moment. Instead of giving up, throwing an error and stopping the client,
 the client will retry as many as connection attempt limit times. This is also the case when the previously
 established connection between the client and that member goes down.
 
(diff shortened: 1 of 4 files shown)


CONFIG.md
```diff
@@ -291,6 +291,30 @@ the properties section in the JSON configuration file. Lastly, the client calls
 
 For information about the path resolution, please refer to the [Path Resolution](#path-resolution-and-object-loading) section.
 
+### Enabling Hazelcast Cloud Discovery
+The purpose of Hazelcast Cloud Discovery is to provide clients to use ip addresses provided by `hazelcast orchestrator`. To enable Hazelcast Cloud Discovery, you should add a discovery token to the `discoveryToken` field and make true the `enabled` field.
```
Serdaro: The purpose of Hazelcast Cloud Discovery is to provide clients to use IP addresses provided by `hazelcast orchestrator`. To enable Hazelcast Cloud Discovery, specify a token for the `discoveryToken` field and set the `enabled` field to "true".

CONFIG.md
```diff
@@ -291,6 +291,30 @@ the properties section in the JSON configuration file. Lastly, the client calls
...
+        "hazelcastCloud": {
+            "discoveryToken": "EXAMPLE_TOKEN",
+            "enabled": true
+        }
+    }
+}
+
+```
+
+To be able to connect to the provided ip addresses, you should use secure TLS/SSL connection between the client and members. Therefore, you should set an SSL configuration as described in the previous section.
```
Serdaro: ....provided IP addresses, you should....

CONFIG.md
```diff
@@ -291,6 +291,30 @@ the properties section in the JSON configuration file. Lastly, the client calls
 
 For information about the path resolution, please refer to the [Path Resolution](#path-resolution-and-object-loading) section.
 
+### Enabling Hazelcast Cloud Discovery
+The purpose of Hazelcast Cloud Discovery is to provide clients to use IP addresses provided by `hazelcast orchestrator`. To enable Hazelcast Cloud Discovery, specify a token for the `discoveryToken` field and set the `enabled` field to "true".
```
mustafaiman: I think what `a token` is is not clear here. Maybe you should express where this token comes from (It is provided by hazelcast orchestrator I assume)

furkansenharputlu: For now, I don't know how user gets the token. When Hazelcast Cloud is out, I think it becomes clear.

PR #327, opened 2018-08-06
implement Client Statistics and NearCacheManager

### LINKED ISSUE #275: The hazelcast.client.statistics.enabled is not supported
Setting the `hazelcast.client.statistics.enabled` property on Hazelcast client config does not seem to work, because of the client connection is not listed in the Hazelcast management center.

```typescript
import { Config } from "hazelcast-client";

const hazelcastConfig = new Config.ClientConfig();

hazelcastConfig.properties["hazelcast.client.statistics.enabled"] = true;
```

I would expected setting the property as in the code snippet above, would make the client appear in the clients list in the Hazelcast management center.

### LINKED ISSUE #322: implement NearCacheManager
`NearCache` works in a different way currently. `NearCacheManager` should be implemented to be consistent with other clients. Also, `Client Statistics` feature needs `NearCacheManager`. Therefore, its priority is high.

furkansenharputlu:
fixes #275 
fixes #322

### DIFF
--- a/src/BuildMetadata.ts
+++ b/src/BuildMetadata.ts
@@ -21,7 +21,7 @@ export class BuildMetadata {
     private static readonly MINOR_VERSION_MULTIPLIER = 100;
     private static readonly PATTERN = /^([\d]+)\.([\d]+)(?:\.([\d]+))?(-[\w]+)?(-SNAPSHOT)?(-BETA-.)?$/;
 
-    public static calculateVersion(versionString: string): number {
+    public static calculateVersionFromString(versionString: string): number {
         if (versionString == null) {
             return BuildMetadata.UNKNOWN_VERSION_ID;
         }
@@ -37,6 +37,10 @@ export class BuildMetadata {
         } else {
             patch = Number.parseInt(info[3]);
         }
+        return this.calculateVersion(major, minor, patch);
+    }
+
+    public static calculateVersion(major: number, minor: number, patch: number): number {
         return BuildMetadata.MAJOR_VERSION_MULTIPLIER * major + BuildMetadata.MINOR_VERSION_MULTIPLIER * minor + patch;
     }
 
--- a/src/Util.ts
+++ b/src/Util.ts
@@ -267,3 +267,27 @@ function setAnchor(list: any[], predicate: PagingPredicate, nearestPage: number)
         predicate.setAnchor(nearestPage, anchor);
     }
 }
+
+export class Task {
+    intervalId: NodeJS.Timer;
+    timeoutId: NodeJS.Timer;
+}
+
+export function scheduleWithRepetition(callback: (...args: any[]) => void, initialDelay: number,
+                                       periodMillis: number): Task {
+    const task = new Task();
+    task.timeoutId = setTimeout(function (): void {
+        callback();
+        task.intervalId = setInterval(callback, periodMillis);
+    }, initialDelay);
+
+    return task;
+}
+
+export function cancelRepetitionTask(task: Task): void {
+    if (task.intervalId != null) {
+        clearInterval(task.intervalId);
+    } else if (task.timeoutId != null) {
+        clearTimeout(task.timeoutId);
+    }
+}
--- a/src/invocation/ClusterService.ts
+++ b/src/invocation/ClusterService.ts
@@ -26,6 +26,7 @@ import {IllegalStateError} from '../HazelcastError';
 import * as assert from 'assert';
 import {MemberSelector} from '../core/MemberSelector';
 import {createAddressFromString} from '../Util';
+import {BuildMetadata} from '../BuildMetadata';
 import Address = require('../Address');
 import ClientMessage = require('../ClientMessage');
 
@@ -205,7 +206,7 @@ export class ClusterService extends EventEmitter {
     private tryConnectingToAddresses(index: number, remainingAttemptLimit: number,
                                      attemptPeriod: number, cause?: Error): Promise<void> {
         this.logger.debug('ClusterService', 'Trying to connect to addresses, remaining attempt limit: ' + remainingAttemptLimit
-            + 'attempt period: ' + attemptPeriod);
+            + ', attempt period: ' + attemptPeriod);
         if (this.knownAddresses.length <= index) {
             remainingAttemptLimit = remainingAttemptLimit - 1;
             if (remainingAttemptLimit === 0) {
--- a/src/nearcache/NearCacheManager.ts
+++ b/src/nearcache/NearCacheManager.ts
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {NearCache, NearCacheImpl} from './NearCache';
+import {SerializationService} from '../serialization/SerializationService';
+import HazelcastClient from '../HazelcastClient';
+
+export class NearCacheManager {
+
+    protected readonly serializationService: SerializationService;
+    private readonly caches: Map<string, NearCache> = new Map();
+    private readonly client: HazelcastClient;
+
+    constructor(client: HazelcastClient) {
+        this.client = client;
+    }
+
+    public getOrCreateNearCache(name: string): NearCache {
+        let nearCache = this.caches.get(name);
+        if (nearCache == null) {
+            nearCache = new NearCacheImpl(this.client.getConfig().getNearCacheConfig(name),
+                this.client.getSerializationService());
+
+            this.caches.set(name, nearCache);
+        }
+        return nearCache;
+    }
+
+    public destroyNearCache(name: string): void {
+        const nearCache = this.caches.get(name);
+        if (nearCache != null) {
+            this.caches.delete(name);
+            nearCache.clear();
+        }
+    }
+
+    public destroyAllNearCaches(): void {
+        for (const key of Array.from(this.caches.keys())) {
+            this.destroyNearCache(key);
+        }
+    }
+
+    public listAllNearCaches(): NearCache[] {
+        return Array.from(this.caches.values());
+    }
+
+}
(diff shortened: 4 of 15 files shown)


src/BuildMetadata.ts
```diff
@@ -21,6 +21,9 @@ export class BuildMetadata {
     private static readonly MINOR_VERSION_MULTIPLIER = 100;
     private static readonly PATTERN = /^([\d]+)\.([\d]+)(?:\.([\d]+))?(-[\w]+)?(-SNAPSHOT)?(-BETA-.)?$/;
 
+    /* tslint:disable:member-ordering */
```
mustafaiman: Why do we need to disable `member-ordering`? This line can go just below UNKNOWN_VERSION_ID.

furkansenharputlu: No, it cannot because `MAJOR_VERSION_MULTIPLIER ` isn't declared there.

src/statistics/Statistics.ts
```diff
@@ -0,0 +1,256 @@
...
+        this.registerGauge('os.freePhysicalMemorySize', () => os.freemem());
+        this.registerGauge('os.freeSwapSpaceSize', () => Statistics.EMPTY_STAT_VALUE);
+        this.registerGauge('os.maxFileDescriptorCount', () => Statistics.EMPTY_STAT_VALUE);
+        this.registerGauge('os.openFileDescriptorCount', () => Statistics.EMPTY_STAT_VALUE);
+        this.registerGauge('os.processCpuTime', () => Statistics.EMPTY_STAT_VALUE);
+        this.registerGauge('os.systemLoadAverage', () => '[' + os.loadavg().toString() + ']');
+        this.registerGauge('os.totalPhysicalMemorySize', () => os.totalmem());
+        this.registerGauge('os.totalSwapSpaceSize', () => Statistics.EMPTY_STAT_VALUE);
+        this.registerGauge('runtime.availableProcessors', () => os.cpus().length);
+        this.registerGauge('runtime.freeMemory', () => Statistics.EMPTY_STAT_VALUE);
```
mustafaiman: we may be able to retrieve some of these information from V8 module: https://nodejs.org/api/v8.html . I am just guessing. Have you checked there?

src/statistics/Statistics.ts
```diff
@@ -0,0 +1,256 @@
...
+            this.sendStats(stats.join(''), ownerConnection);
+        }, 0, periodSeconds * 1000);
+    }
+
+    sendStats(newStats: string, ownerConnection: ClientConnection) {
+        const request = ClientStatisticsCodec.encodeRequest(newStats);
+
+        this.client.getInvocationService().invokeOnTarget(request, ownerConnection.getAddress()).catch((err) => {
+            // suppress exception, do not print too many messages
+            this.logger.trace('Statistics', 'Could not send stats ', err);
```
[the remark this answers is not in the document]
furkansenharputlu: Java side:
```
 // suppress exception, do not print too many messages
            if (logger.isFinestEnabled()) {
                logger.finest("Could not send stats ", e);
            }
````
I thought `finest` equals `trace` and if trace is not enabled it won't log.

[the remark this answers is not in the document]
Comment (mustafaiman) [reply src/statistics/Statistics.ts]: when you say `do not print too many messages` i expect some of the messages to get suppressed but every once in a while a message is printed. Having `trace` logs mean having `trace` logs. I do not see any value in above comment.

File: src/BuildMetadata.ts [hunk 4]
```diff
@@ -21,6 +21,9 @@ export class BuildMetadata {
...
+    public static readonly VERSION_3_8 = BuildMetadata.MAJOR_VERSION_MULTIPLIER * 3 + BuildMetadata.MINOR_VERSION_MULTIPLIER * 8;
```
ihsandemir: why is name of variable bound to a specific version?

src/ClientInfo.ts
```diff
@@ -28,5 +28,5 @@ export class ClientInfo {
     /**
      * Type of this client. It is always NodeJS.
      */
-    type: string = 'NodeJS';
+    type: string = 'NODEJS';
```
ihsandemir: Conflicts with the above doc. is this communicated on the wire?

src/invocation/ClusterService.ts
```diff
@@ -161,6 +162,14 @@ export class ClusterService extends EventEmitter {
         return this.ownerConnection;
     }
 
+    getConnectedServerVersion(): number {
```
ihsandemir: each connected server may have a different version, are you returning the connection for the owner member? Or some other member? Why is there a for loop here?

src/nearcache/NearCacheManager.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import {NearCache, NearCacheImpl} from './NearCache';
+import {SerializationService} from '../serialization/SerializationService';
+import HazelcastClient from '../HazelcastClient';
+
+export class NearCacheManager {
+
+    protected readonly serializationService: SerializationService;
+    private readonly caches: Map<string, NearCache> = new Map();
```
ihsandemir: is this concurrent map?

src/statistics/Statistics.ts
```diff
@@ -0,0 +1,264 @@
...
+            }
+            // cache the last connected server address for decreasing the log prints
+            this.ownerAddress = ownerConnectionAddress;
+            return null;
+        }
+        return connection;
+    }
+
+    private registerMetrics() {
+        this.registerGauge('os.committedVirtualMemorySize', () => Statistics.EMPTY_STAT_VALUE);
```
ihsandemir: Too many empty state gauges. registering them will mean unnecessary parsing. do we really need to register unsupplied stats?

test/HazelcastClientTest.js
```diff
@@ -108,7 +108,7 @@ configParams.forEach(function (cfg) {
             expect(info.localAddress.host).to.equal(client.clusterService.getOwnerConnection().localAddress.host);
             expect(info.localAddress.port).to.equal(client.clusterService.getOwnerConnection().localAddress.port);
             expect(info.uuid).to.equal(client.clusterService.uuid);
-            expect(info.type).to.equal('NodeJS');
+            expect(info.type).to.equal('NODEJS');
```
ihsandemir: breaking change?

test/statistics/StatisticsTest.js
```diff
@@ -0,0 +1,222 @@
...
+        }).then(function (stats) {
+            expect(stats).to.not.equal('');
+        });
+    });
+
+    it('should contain statistics content', function () {
+        return Util.promiseWaitMilliseconds(1000).then(function () {
+            return getClientStatisticsFromServer(cluster);
+        }).then(function (stats) {
+            expect(stats).to.not.be.null;
```
ihsandemir: what about if exact number of stats exist or if there are more or less, can you test for it?

src/nearcache/NearCacheManager.ts
furkansenharputlu: No, it isn't. `ConcurrentMap` isn't necessary for Node.js because there is just main thread.

src/statistics/Statistics.ts
furkansenharputlu: I don't know exact reason. I referenced the C# client. See: https://github.com/hazelcast/hazelcast-csharp-client/blob/89e7faa67ac5d25497d9f81bc43bcc7fcdf74442/Hazelcast.Net/Hazelcast.Client/Statistics.cs#L170

test/HazelcastClientTest.js
furkansenharputlu: I changed this because it is mentioned as `NODEJS` in server side.

test/statistics/StatisticsTest.js
furkansenharputlu: It can't be less because `contains` tests are for all stats. However I can write for more stats case.

src/statistics/Statistics.ts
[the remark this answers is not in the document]
furkansenharputlu: What is the equivalent of `logger.isFinestEnabled()` in Node.js side ?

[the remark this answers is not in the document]
mustafaiman: `trace` is fine. I just find the above comment useless. We can remove the comment.

test/statistics/StatisticsTest.js
furkansenharputlu: I updated the contains method. It also checks whether a stat exists once or not.

src/BuildMetadata.ts
furkansenharputlu: It is needed on near cache implementation. Near cache behaves differently after 'VERSION_3_8'.

mustafaiman: @furkansenharputlu I think we should not define specific versions in `BuildMetadata` as this class contains information about current specific release. We can introduce `Versions` class as in Java if this is neccessary.

Additionally, I think we should introduce a new method that takes major, minor and patch as parameters and creates the specific number as the current `calculateVersion` does. Current approach to create a specific version number is error prone.

mustafaiman: Sorry I mixed `BuildMetadata` and `BuildInfo` classes. You can dismiss my first paragraph but the second still applies

src/invocation/ClusterService.ts
mustafaiman: I think the original method is flawed. This should look for the owner connection's version.

test/BuildMetadataTest.js
```diff
@@ -19,21 +19,21 @@ var assert = require('chai').assert;
...
 describe('BuildMetadata', function () {
     it('version calculation test', function () {
-        assert.equal(-1, BuildMetadata.calculateVersion(null));
-        assert.equal(-1, BuildMetadata.calculateVersion(""));
-        assert.equal(-1, BuildMetadata.calculateVersion("a.3.7.5"));
-        assert.equal(-1, BuildMetadata.calculateVersion("3.a.5"));
-        assert.equal(-1, BuildMetadata.calculateVersion("3,7.5"));
-        assert.equal(-1, BuildMetadata.calculateVersion("3.7,5"));
-        assert.equal(-1, BuildMetadata.calculateVersion("10.99.RC1"));
+        assert.equal(-1, BuildMetadata.calculateVersionFromString(null));
```
mustafaiman: can you also add tests for newly added `calculateVersion`

src/Util.ts
```diff
@@ -267,3 +267,27 @@ function setAnchor(list: any[], predicate: PagingPredicate, nearestPage: number)
...
+    const task = new Task();
+    task.timeoutId = setTimeout(function (): void {
+        callback();
+        task.intervalId = setInterval(callback, periodMillis);
+    }, initialDelay);
+
+    return task;
+}
+
+export function cancelRepetitionTask(task: Task): void {
```
mustafaiman: what happens if I cancel the same task twice?

furkansenharputlu: I am not sure but why do we cancel same task twice ?

mustafaiman: i can imagine a scenario that there are two different code paths leading to cancel a task. In that case, the second one to reach cancellation code, should not cause a runtime error.

furkansenharputlu: I added a test for that scenario. There is no error.

PR #331, opened 2018-08-13
prevent client hanging due to incompatible server

### LINKED ISSUE #132: Client does not fail fast with incompatible server
If server does not support open binary protocol, it seems client hangs forever without giving any error. It should fail fast with a reasonable error message.

furkansenharputlu:
This PR fixes client hanging defect that occurs when there is a server that does not support open binary protocol. It returns an `IllegalStateError` with message `Connection timed-out`.

fixes #132

### DIFF
--- a/test/ConnectionManagerTest.js
+++ b/test/ConnectionManagerTest.js
@@ -14,13 +14,17 @@
  * limitations under the License.
  */
 
-var Controller = require('./RC');
 var chai = require('chai');
 chai.should();
 chai.use(require('chai-as-promised'));
+var expect = chai.expect;
+
+var net = require('net');
+
 var Config = require('../.').Config;
 var Hazelcast = require('../.').Client;
-var net = require('net');
+var Controller = require('./RC');
+var Errors = require('../').HazelcastErrors;
 
 describe('ConnectionManager', function () {
 
@@ -36,7 +40,7 @@ describe('ConnectionManager', function () {
             return Controller.startMember(cluster.id);
         }).then(function (m) {
             member = m;
-        })
+        });
     });
 
     beforeEach(function () {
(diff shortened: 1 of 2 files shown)


test/ConnectionManagerTest.js
```diff
@@ -100,5 +108,17 @@ describe('ConnectionManager', function () {
                 done(new Error('Client should be retrying!\n' + e));
             }
         });
-    })
+    });
+
+    it('should throw IllegalStateError if there is an incompatible server', function () {
```
mustafaiman: Isn't this the same test as `gives up connecting after timeout`?

furkansenharputlu: No, that test just checks first connection attempt and was passing. Client was hanging at the second attempt. This test checks whether client hangs or not after first attempt.

mustafaiman: Nevermind, I see the difference. However, I did not like this test as it assumes a certain order of running tests. The test shutdowns the cluster that the others depend on. You can isolate the unresponsive server by starting it at a different port and having client configuration list it.

furkansenharputlu: However, client will try `5701` port when it cannot connect. Therefore, I need to run the server on `5701`.

mustafaiman: Why should it? If I give only a single port in config, will it try 5701 anyway? It may try but I doubt it should.

furkansenharputlu: It adds here: https://github.com/hazelcast/hazelcast-nodejs-client/blob/22ecdbf996a35761c63aee0eccd72ef919160d7b/src/connection/DefaultAddressProvider.ts#L39

furkansenharputlu: Actually, you are right when there is no address passed on config, it automatically tries to connect to `5071`. When I pass `127.0.0.1:9999`, it won't try to connect to `5071`.

Issue #332, opened 2018-08-14
Node.js client tests failures on windows


olukas:
After tests from https://github.com/hazelcast/hazelcast-nodejs-client/issues/307 were fixed we hit another bulk of test failures on Windows machine for node.js client. All of them seems to be intermittent failures:

1) Listeners on reconnect restart member, listener still receives map.put event [smart=true]:
      Error: Timeout of 120000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

see: https://hazelcast-l337.ci.cloudbees.com/job/NodeJS-4-windows/19/console
 
2) Lost connection M2 starts, M1 goes down, client sets M2 as owner:
  TypeError: Cannot read property 'address' of null
   at Timeout._onTimeout (test\LostConnectionTest.js:57:74) 

see: https://hazelcast-l337.ci.cloudbees.com/job/NodeJS-8-windows/16/console

3) SSL Client Authentication Test via programmatic configuration "after each" hook for "ma:optional, neither knows the other should fail":
      Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

see: https://hazelcast-l337.ci.cloudbees.com/job/NodeJS-10-windows/17/console

puzpuzpuz:
Closing this one as it seems to be obsolete and there are up-to-date issues for flaky tests.

PR #337, opened 2018-08-30
fix Listeners on reconnect tests

furkansenharputlu:
This PR fixes `Listeners on reconnect` tests failing frequently on windows builds. Here is my playground for windows builds: https://hazelcast-l337.ci.cloudbees.com/job/NodeJS-windows-playground/

#332

### DIFF
--- a/test/ListenersOnReconnectTest.js
+++ b/test/ListenersOnReconnectTest.js
@@ -36,53 +36,54 @@ describe('Listeners on reconnect', function () {
     });
 
     afterEach(function () {
-        map.destroy();
-        client.shutdown();
-        return Controller.shutdownCluster(cluster.id);
+        return map.destroy().then(function () {
+            client.shutdown();
+            return Controller.shutdownCluster(cluster.id);
+        });
     });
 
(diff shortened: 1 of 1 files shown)


test/ListenersOnReconnectTest.js
```diff
@@ -144,11 +147,11 @@ describe('Listeners on reconnect', function () {
                 return Controller.terminateMember(cluster.id, member.uuid);
             }).then(function () {
                 return Controller.startMember(cluster.id);
+            }).then(function () {
+                return Util.promiseWaitMilliseconds(5000);
```
mustafaiman: Doesn't this wait contradict with the purpose of the test? The client should not lose neither `put` request nor the listener event. Losing the connection to the cluster must not disturb normal operation as long as the cluster comes back alive within connection timeout period. We do not have to give the client a break to get itself together before sending a put.

furkansenharputlu:
That waiting time is about listener to be active. When it doesn’t wait, event doesn’t fire and this causes the test to timeout.

Issue #340, opened 2018-08-31
Node.js Client Readme Should be Rewritten


Serdaro:
As depicted in the PRD https://hazelcast.atlassian.net/wiki/spaces/PM/pages/868155612/Guide+to+using+the+Node.js+Client+with+Hazelcast , we expect the Readme document to include the following chapters and sections.

Please refer to the aforementioned PRD for the content description of each section.

- [x] Make sure that the issue and the PRD outlines are compatible at the end.
- [x] Indexing should be made when all chapters are added.
- [x] Check for redundant parts and merge good content from redundant parts at the end.
- [x] Check for broken links.

## Introduction

| Content                                 | In Progress | Done  | Assigned to
|----------------------------|-------------|-------|------------
| TOC                                       |                      |    X     | Serdar
| Introduction Paragraph       |                      |   X       | Serdar
| Version Compatibility          |                      |    X      | Serdar
| Resources                            |                      |     X     | Serdar
| Release Notes                      |                      |     X     | Serdar

## Chapter 1. Getting Started

| Section                                                 | In Progress | Done  | Assigned to
|-------------------------------------|-------------|-------|-------------
| 1.1 Requirements                                 |                     |        X    | Mustafa
| 1.2 Working with Hazelcast Clusters |                     |        X    | Mustafa
| 1.2.1.1 Running Standalone Jars        |                     |     X       | Mustafa
| 1.2.1.2 Adding User Library to CLASSPATH |                     |     X       | Furkan
| 1.2.1.3 Using hazelcast-member Tool |                     |     X       | Mustafa
| 1.3 Downloading and Installing           |                     |       X     | Mustafa
| 1.4 Basic Configuration                       |                     |       X     | Mustafa
| 1.4.1 IMDG Configuration                      |                     |       X     | Mustafa
| 1.4.2 Hazelcast Client Configuration   |                     |       X     | Mustafa
| 1.5 Basic Usage                                   |                     |      X      | Mustafa
| 1.6 Code Samples                               |                     |    X        | Mustafa

## Chapter 2. Features 

| Content                                 | In Progress | Done  | Assigned to
|----------------------------|-------------|-------|------------
| List of features                    |                     |     X       | Mustafa

## Chapter 3. Configuration Overview

| Section                                                         | In Progress | Done  | Assigned to
|------------------------------------------|-------------|-------|--------------
| 3.1 Configuration Options                           |                      |      X    | Furkan
[...]

Issue #248, opened 2018-01-24
Clarify the Serialization Consideration chapter


jerrinot:
For example it says: "Register the serializers for nested objects". It should show how to do it. 
Also this part is not clear to me: "Regarding arrays in a serializable object, you can use methods like writeIntArray if the array is of a primitive type." What's so special about arrays? Is this related to customer serializers?

mustafaiman:
@jerrinot can you take a look at https://github.com/hazelcast/hazelcast-nodejs-client/pull/278 and see if it looks better that way?

PR #341, opened 2018-08-31
add documentation for serialization content and IdentifiedDataSeriali…

furkansenharputlu:
…zable

#248 
#340

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,6 @@
 
 * [Hazelcast Node.js Client](#hazelcast-nodejs-client)
 * [Features](#features)
-* [Installing the Client](#installing-the-client)
-* [Using the Client](#using-the-client)
 * [Configuration Overview](#configuration-overview)
   * [1. Configuration Options](#1-configuration-options)
     * [1.1. Programmatic Configuration](#11-programmatic-configuration)
@@ -11,7 +9,7 @@
   * [2. Importing Multiple Configurations](#2-importing-multiple-configurations)
   * [3. Loading Objects and Path Resolution](#3-loading-objects-and-path-resolution)
 * [Code Samples](#code-samples)
-* [Serialization Considerations](#serialization-considerations)
+* [Serialization](#serialization)
 * [Setting Up Client Network](#setting-up-client-network)
   * [1. Providing the Member Addresses](#1-providing-the-member-addresses)
   * [2. Setting Smart Routing](#2-setting-smart-routing)
@@ -590,42 +588,95 @@ Please see Hazelcast Node.js [code samples](https://github.com/hazelcast/hazelca
 
 You can also refer to Hazelcast Node.js [API Documentation](http://hazelcast.github.io/hazelcast-nodejs-client/api/current/docs/).
 
-# Serialization Considerations
+# Serialization
 
-Hazelcast needs to serialize objects in order to be able to keep them in the server memory. For primitive types, it uses Hazelcast native serialization. For other complex types (e.g. JS objects), it uses JSON serialization.
+Serialization is the process of converting an object into a stream of bytes to store the object in memory, a file or database, or transmit it through network. Its main purpose is to save the state of an object in order to be able to recreate it when needed. The reverse process is called deserialization. Hazelcast offers you its own native serialization methods. You will see these methods throughout the chapter.
 
-For example, when you try to query your data using predicates, this querying is handled on the server side so Hazelcast does not have to bring all data to the client but only the relevant entries. Otherwise, there would be a lot of unneccessary data traffic between the client and the server and the performance would severely drop.
-Because predicates run on the server side, the server should be able to reason about your objects. That is why you need to implement serialization on the server side.
+Hazelcast serializes all your objects before sending them to the server. The `boolean`, `number`,`string` and `Long` types are serialized natively and you cannot override this behavior. The following table is the conversion of types for Java server side.
 
-The same applies to MapStore. The server should be able to deserialize your objects in order to store them in MapStore.
+| Node.js | Java                                |
+|---------|-------------------------------------|
+| boolean | Boolean                             |
+| number  | Byte, Short, Integer, Float, Double |
+| string  | String                              |
+| Long    | Long                                |
 
-Regarding arrays in a serializable object, you can use methods like `writeIntArray` if the array is of a primitive type.
+> Note: A `number`type is serialized as `Double` by default. You can configure this behaviour from `SerializationConfig.defaultNumberType`.
 
-If you have nested objects, these nested objects also need to be serializable. Register the serializers for nested objects and the method `writeObject` will not have any problem with finding a suitable serializer for and writing/reading the nested object.
+Arrays of the above types can be serialized as `boolean[]`, `byte[]`, `short[]`, `int[]`, `float[]`, `double[]`, `long[]` and `string[]` for Java server side respectively. 
 
-If you have arrays of custom objects, you can serialize/deserialize them like the following:
+Note that if the object is not one of the above-mentioned types, the Node.js client uses `JSON Serialization` by default.
+
+However, `JSON Serialization` is not the best way of serialization in terms of performance and interoperability between the clients in different languages. If you want the serialization to work faster or you use the clients in different languages, Hazelcast offers its own native serialization types, such as [IdentifiedDataSerializable Serialization](#1-identifieddataserializable-serialization) and [Portable Serialization](#2-portable-serialization).
+
+On top of all, if you want to use your own serialization type, you can use a [Custom Serialization](#3-custom-serialization).
+
+> **NOTE: Hazelcast Node.js Client is a TypeScript-based project but JavaScript does not have interfaces. Therefore, 
+ some interfaces are given to user by using the TypeScript files that have `.ts` extension. In the documentation, implementing an interface means an object to have the necessary functions that are listed in the interface inside the `.ts` file. Also, this object is mentioned as `an instance of the interface`. You can search the [API Documentation](http://hazelcast.github.io/hazelcast-nodejs-client/api/current/docs/) or Github repository for a required interface.**
+
+## 1. IdentifiedDataSerializable Serialization
+For a faster serialization of objects, Hazelcast recommends to implement IdentifiedDataSerializable interface.
+
+Here is an example of an object implementing IdentifiedDataSerializable interface:
 
 ```javascript
-writeData(dataOutput) {
-    ...
-    dataOutput.writeInt(this.arrayOfCustomObjects);
-    this.arrayOfCustomObjects.forEach(function(element) {
-        dataOutput.writeObject(element);
-    });
-    ...
+function Address(street, zipCode, city, state) {
+    this.street = street;
+    this.zipCode = zipCode;
+    this.city = city;
+    this.state = state;
 }
 
-readData(dataInput) {
-    ...
-    var arrayOfCustomObjects = [];
-    var lenOfArray = dataInput.readInt();
-    for (i=0;i<lenOfArray;i++) {
-        arrayOfCustomObjects.push(dataInput.readObject());
-    }
-    this.arrayOfCustomObjects = arrayOfCustomObjects;
-    ...
-}
+Address.prototype.getClassId = function () {
+    return 1;
+};
+
+Address.prototype.getFactoryId = function () {
+    return 1;
+};
+
+Address.prototype.writeData = function (objectDataOutput) {
+    objectDataOutput.writeUTF(this.street);
+    objectDataOutput.writeInt(this.zipCode);
+    objectDataOutput.writeUTF(this.city);
+    objectDataOutput.writeUTF(this.state);
+};
+
+Address.prototype.readData = function (objectDataInput) {
+    this.street = objectDataInput.readUTF();
+    this.zipCode = objectDataInput.readInt();
+    this.city = objectDataInput.readUTF();
+    this.state = objectDataInput.readUTF();
+};
 ```
+
+IdentifiedDataSerializable uses `getClassId()` and `getFactoryId()` to reconstitute the object. To complete the implementation `IdentifiedDataSerializableFactory` should also be implemented and registered into `SerializationConfig` which can be accessed from `Config.serializationConfig`. The factory's responsibility is to return an instance of the right `IdentifiedDataSerializable` object, given the class id. 
+
+A sample `IdentifiedDataSerializableFactory` could be implemented as following:
+
+```javascript
+var myIdentifiedFactory = {
+    create: function (type) {
+        if (type === 1) {
+            return new Address();
+        }
+    }
+};
+```
+
+The last step is to register the `IdentifiedDataSerializableFactory` to the `SerializationConfig`.
+
+```javascript
+var config = new Config.ClientConfig();
+config.serializationConfig.dataSerializableFactories[1] = myIdentifiedFactory;
+```
+
+## 2. Portable Serialization
+
+## 3. Custom Serialization
+
+## 4. Global Serialization
+
 
 # Setting Up Client Network
 


README.md
```diff
@@ -108,43 +108,92 @@ See [CONFIG.md](CONFIG.md) for details.
...
 
-Hazelcast needs to serialize objects in order to be able to keep them in the server memory. For primitive types, it uses Hazelcast native serialization. For other complex types (e.g. JS objects), it uses JSON serialization.
 
-For example, when you try to query your data using predicates, this querying is handled on the server side so Hazelcast does not have to bring all data to the client but only the relevant entries. Otherwise, there would be a lot of unneccessary data traffic between the client and the server and the performance would severely drop.
-Because predicates run on the server side, the server should be able to reason about your objects. That is why you need to implement serialization on the server side.
+Hazelcast serializes all your objects into an instance of `Data`. `Data` is the binary representation of an object. When Hazelcast serializes an object into `Data`, it first checks whether the object is an instance of `IdentifiedDataSerializable`, if not it checks if it is an instance of `Portable` and serializes it accordingly. For the following types Hazelcast optimizes the serialization a user can not override this behaviour.
```
mustafaiman: `Data` is an implementation detail. There is no way user sees a `Data` instance in normal operation afaik. It should not be part of the guide.

furkansenharputlu: I referenced Java reference manual. See: http://docs.hazelcast.org/docs/3.0/manual/html/ch03.html. I am trying to be as consistent with it as possible.

README.md
```diff
@@ -108,43 +108,92 @@ See [CONFIG.md](CONFIG.md) for details.
 See [Code Samples](https://github.com/hazelcast/hazelcast-nodejs-client/tree/master/code_samples)
 
 
-# Serialization Considerations
+# Serialization
+**Table of Contents**
```
Serdaro: For the sake of simplicity, I would rather not include the in-section TOC here. And maybe an introductory paragraph talking about serialization itself and why it is needed would be nice.

furkansenharputlu:
@mustafaiman It is explained in detail at the beginning. `Note that if the object is not instance of any explicit type, Node.js client uses JSON Serialization. The default behaviour can be changed by using a Custom Serialization.` This note can be useful.

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
 See [Code Samples](https://github.com/hazelcast/hazelcast-nodejs-client/tree/master/code_samples)
 
 
-# Serialization Considerations
+# Serialization
```
Serdaro: Agreeing with Mustafa: We need to have an intro here explaining what is serialization and why do we need it. After those explanations, we also need to give a paragraph about custom serialization including some use cases so that the readers understand when they need a custom serialization.

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
 
-Hazelcast needs to serialize objects in order to be able to keep them in the server memory. For primitive types, it uses Hazelcast native serialization. For other complex types (e.g. JS objects), it uses JSON serialization.
+Hazelcast serializes all your objects into an instance of `Data`. `Data` is the binary representation of an object. When Hazelcast serializes an object into `Data`, it first checks whether the object is an instance of `IdentifiedDataSerializable`, if not it checks if it is an instance of `Portable` and serializes it accordingly. For the following types Hazelcast optimizes the serialization a user can not override this behaviour.
```
Serdaro: ....Hazelcast optimizes the serialization and you cannot override this behavior

(notice the usage of "cannot" and "behavior").

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
+- A `boolean` type which is a primitive on Node.js can be serialized as a `Boolean` for Java server-side.
```
Serdaro: I am guessing there is a new line before the first bulleted item.

...a primitive on the Node.js client can be serialized as a `Boolean` for the Java server side.

(need to apply to below "server-side" since it is not used as an adjective here)

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
+- A character which is a string with length one can be serialized as a `Character` for Java server-side.
```
Serdaro: For the sake of correctly structured and easily understandable sentence, we need to be explicit on the "string with length one" phrase. Giving a unit always helps (string with 8 bit length, etc.). Or using it like "single-length string" can be another alternative.

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
+- A `number` type which is a primitive on Node.js can be serialized as `Byte`, `Short`, `Integer`, `Float`, `Double` for Java server-side.
```
Serdaro: ...`Float` and `Double` for Java server side.

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
+- A `Long` type which represents a 64 bit two's-complement integer value can be serialized as a `Long` for Java server-side.
```
Serdaro: No need a dash: two's complement integer...

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
+- A `string` type which is a primitive on Node.js can be serialized as a `String` for Java server-side.
+- Arrays of the above types can be serialized as `boolean[]`, `char[]`, `byte[]`, `short[]`, `int[]`, `float[]`, `double[]`, `long[]`, `string[]` for Java server-side respectively.
```
Serdaro: [1] ...`long[]` and `string[]` for Java server side...

[2] To which order does "respectively" refer here?

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
+- A `Long` type which represents a 64 bit two's-complement integer value can be serialized as a `Long` for Java server-side.
+- A `string` type which is a primitive on Node.js can be serialized as a `String` for Java server-side.
+- Arrays of the above types can be serialized as `boolean[]`, `char[]`, `byte[]`, `short[]`, `int[]`, `float[]`, `double[]`, `long[]`, `string[]` for Java server-side respectively. 
 
-For example, when you try to query your data using predicates, this querying is handled on the server side so Hazelcast does not have to bring all data to the client but only the relevant entries. Otherwise, there would be a lot of unneccessary data traffic between the client and the server and the performance would severely drop.
-Because predicates run on the server side, the server should be able to reason about your objects. That is why you need to implement serialization on the server side.
+Note that if the object is not instance of any explicit type, Node.js client uses `JSON Serialization`. The default behaviour can be changed by using a `Custom Serialization`.
```
Serdaro: Note that if the object is not an instance of any explicit type, Node.js client uses `JSON Serialization` by default. This default behavior can be changed by using a `Custom Serialization`.

(Please give a link to Custom Serialization section here by making it a cross-reference in the above sentence).

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
+
+Address.prototype.readData = function (objectDataInput) {
+    this.street = objectDataInput.readUTF();
+    this.zipCode = objectDataInput.readInt();
+    this.city = objectDataInput.readUTF();
+    this.state = objectDataInput.readUTF();
+};
 ```
 
+IdentifiedDataSerializable uses `getClassId()` and `getFactoryId()` to reconstitue the object. To complete the implementation `IdentifiedDataSerializableFactory` should also be implemented and registered into `SerializationConfig` which can be accessed from `Config.serializationConfig`. The Factories responsibility is to return an instance of the right `IdentifiedDataSerializable` object, given the class id.
```
Serdaro: ..reconstitute the object.


...The factory's responsibility is....
(or) 
...`IdentifiedDataSerializableFactory`s responsibility is...

README.md
```diff
@@ -108,43 +108,86 @@ See [CONFIG.md](CONFIG.md) for details.
...
+```
+
+The last step is to register the `IdentifiedDataSerializableFactory` to the `SerializationConfig`.
+
+```javascript
+var config = new Config.ClientConfig();
+config.serializationConfig.dataSerializableFactories[1] = myIdentifiedFactory;
+```
+
+So far this is the most efficient way of Serialization that Hazelcast supports of the shelf.
```
Serdaro: No need to capitalize serialization within the body context.

README.md
mustafaiman: @furkansenharputlu please do not use `custom serialization` as a term describing any serialization that is not Hazelcast native serialization. `custom serialization` has a specific meaning. Custom serializer is one of the serialization methods that the user can configure (along with IdentifiedDataSerializable, Portable and Global Serializer).

What I understand from this section is everything except primitives are converted to JSON by default(which is correct) and the only way to use something else is to configure a custom serializer (which is wrong and not a popular choice at all)

README.md
mustafaiman: Is `explicit type` a defined thing? If I have to define the types that Hazelcast serialization takes care of, I would say primitive types. Yet primitive types also do not cover it because we also take care of serializing `Date` objects which are not primitive.

README.md
furkansenharputlu: All types are mentioned respectively while reading the header.

README.md
furkansenharputlu: I gave a link to `Custom Serialization`.

README.md
```diff
@@ -108,43 +108,89 @@ See [CONFIG.md](CONFIG.md) for details.
 See [Code Samples](https://github.com/hazelcast/hazelcast-nodejs-client/tree/master/code_samples)
 
 
-# Serialization Considerations
+# Serialization
 
-Hazelcast needs to serialize objects in order to be able to keep them in the server memory. For primitive types, it uses Hazelcast native serialization. For other complex types (e.g. JS objects), it uses JSON serialization.
+Serialization is the process of converting an object into a stream of bytes to store the object or transmit it to memory, a database, or a file. Its main purpose is to save the state of an object in order to be able to recreate it when needed. The reverse process is called deserialization.
```
mustafaiman: `to store the object in memory, a file or database, or transmit it through network`

README.md
mustafaiman: This does not justify putting an implementation detail in the guide. Furthermore, being consistent with the reference manual is an anti-goal. This is a guide targeting people that does not know anything about Hazelcast. Reference manual targets a different audience. The reference manual is Java centric which is again just the opposite of what we are trying to do here.

To me, as a new user trying to understand how I can use Hazelcast and Hazelcast Node.js Client, objects being converted to some internal representation does not mean anything.

README.md
```diff
@@ -108,43 +108,89 @@ See [CONFIG.md](CONFIG.md) for details.
...
 
-For example, when you try to query your data using predicates, this querying is handled on the server side so Hazelcast does not have to bring all data to the client but only the relevant entries. Otherwise, there would be a lot of unneccessary data traffic between the client and the server and the performance would severely drop.
-Because predicates run on the server side, the server should be able to reason about your objects. That is why you need to implement serialization on the server side.
+Hazelcast serializes all your objects into an instance of `Data`. `Data` is the binary representation of an object. When Hazelcast serializes an object into `Data`, it first checks whether the object is an instance of `IdentifiedDataSerializable`, if not it checks if it is an instance of `Portable` and serializes it accordingly. For the following types Hazelcast optimizes the serialization and you cannot override this behavior.
```
mustafaiman: "whether the object is an instance of `IdentifiedDataSerializable`" is not correct. IdentifiedDataSerializable object are not instances of IdentifiedDataSerializable. IdentifiedDataSerializable does not even exist in compiled code. What we check is if an object has the necessary methods to be serialized using a registered identified data serializer.

The same for portable part too.

README.md
mustafaiman: This comment shows collapsed because the line has changed but my concern is not resolved.

README.md
mustafaiman: @furkansenharputlu I am not sure how giving a link to custom serialization addresses my concern. The documentation still uses custom serialization as the only alternative to default json behaviour. see `the default behavior can be changed by using a [Custom Serialization](#3.-custom-serialization).`

README.md
furkansenharputlu: Doesn't `whether the object is an instance of IdentifiedDataSerializable` mean `whether the object implements necessary methods of IdentifiedDataSerializable` interface ?

mustafaiman: It is correct for typescript but we need to write this guide for Javascript users. Javascript does not have interfaces or implementing some stuff of another class. There is no class at all. I think it is important to be correct when using terms. I am not very sure how to explain this either. Better thing to do is to take a look at maybe other documentations or how js users generally explain this kind of relationship on the internet.

README.md
furkansenharputlu: I changed that sentence.

README.md
```diff
@@ -108,43 +108,94 @@ See [CONFIG.md](CONFIG.md) for details.
...
-Because predicates run on the server side, the server should be able to reason about your objects. That is why you need to implement serialization on the server side.
+> **NOTE: Hazelcast Node.js Client is a TypeScript-based project but JavaScript does not have interfaces. Therefore, 
+ some interfaces are given to user by using the TypeScript files that have `.d.ts` extension. In the documentation, implementing an interface
+ means an object to have the necessary functions that are listed in the interface inside the `.d.ts` file. Also, this object is mentioned as 
+ `an instance of the interface`.**
 
-The same applies to MapStore. The server should be able to deserialize your objects in order to store them in MapStore.
+Hazelcast serializes all your objects into an instance of `Data`. `Data` is the binary representation of an object. When Hazelcast serializes an object into `Data`, it first checks whether the object is an instance of `IdentifiedDataSerializable`, if not it checks if it is an instance of `Portable` and serializes it accordingly. For the following types Hazelcast optimizes the serialization and you cannot override this behavior.
```
mustafaiman: I wrote this before but looks like the comment is lost. `Data` is an implementation detail. There is no need for the user to know about it.

What the user needs to know in the beginning is (preferably in this order):
1. Hazelcast has a serialization method for some well known types
    - types and their Java counterparts as explained below is very nice.
2. The client JSON serializes anything that is not Hazelcast native serializable by default.
    - It is important to emphasize that this behaviour can be overridden by using one of the following methods.
3. Hazelcast has some ways for users to plug-in their own serialization code
     1. IdentifiedDataSerializable
    2. Portable
    3. Custom Serializer
    4. Global Serializer

First you told the user that he may want to know about something called serialization above. This is good. Now tell the user there is a default way of handling things in Hazelcast Node.js Client. Tell them why they might not like the default behavior sometimes (language interoperability, entry processors). Tell them they have some ways of overriding the default behavior. Give them an outline of how many ways they have to override the behavior and their names. Then explain all of them one by one.

Do not try to tell many different things in one sentence (even in one paragraph). Above paragraph is trying to tell 3 different things in 2 sentences. Let me tell you what it tries to tell
- There is a default serialization
- There are ways to override default serialization
- The order of non-default serialization preference

Each of these should be in their own simple parahgraphs.

README.md
```diff
@@ -108,43 +108,94 @@ See [CONFIG.md](CONFIG.md) for details.
...
 
-For example, when you try to query your data using predicates, this querying is handled on the server side so Hazelcast does not have to bring all data to the client but only the relevant entries. Otherwise, there would be a lot of unneccessary data traffic between the client and the server and the performance would severely drop.
-Because predicates run on the server side, the server should be able to reason about your objects. That is why you need to implement serialization on the server side.
+Hazelcast serializes all your objects before sending them to the server. For the following types Hazelcast optimizes the serialization and you cannot override this behavior.
 
-The same applies to MapStore. The server should be able to deserialize your objects in order to store them in MapStore.
+- A `boolean` type which is a primitive on the Node.js client can be serialized as a `Boolean` for Java server side.
```
mustafaiman: I think Java counterparts are not necessary at this point. We can just give a list of the types here. Then, we can add a table below to show which type corresponds to which data type in Java.

README.md
```diff
@@ -108,43 +108,94 @@ See [CONFIG.md](CONFIG.md) for details.
...
+- A character which is a single-length string one can be serialized as a `Character` for Java server side.
```
mustafaiman: I dont think this is true. We still use string serializer for one character strings on Node.js side.

README.md
```diff
@@ -108,43 +108,94 @@ See [CONFIG.md](CONFIG.md) for details.
...
 ```
 
+The last step is to register the `IdentifiedDataSerializableFactory` to the `SerializationConfig`.
+
+```javascript
+var config = new Config.ClientConfig();
+config.serializationConfig.dataSerializableFactories[1] = myIdentifiedFactory;
+```
+
+So far this is the most efficient way of serialization that Hazelcast supports of the shelf.
```
mustafaiman: It is not right to say Identified is the most efficient. Portable is more efficient in terms of querying. They have some strengths over the other. That is why we maintain both of them.

README.md
```diff
@@ -108,43 +108,94 @@ See [CONFIG.md](CONFIG.md) for details.
...
 
-Regarding arrays in a serializable object, you can use methods like `writeIntArray` if the array is of a primitive type.
+Note that if the object is not one of the above-mentioned types, the Node.js client uses `JSON Serialization` by default. You can create your complex object and it is serialized by using `JSON Serialization`. After serialization, the byte array format of your object is sent to the server through network.
 
-If you have nested objects, these nested objects also need to be serializable. Register the serializers for nested objects and the method `writeObject` will not have any problem with finding a suitable serializer for and writing/reading the nested object.
+However, `JSON Serialization` is not the best way of serialization in terms of performance. If you want the serialization to work faster, Hazelcast offers its own native serialization types, such as [IdentifiedDataSerializable Serialization](#1-identifieddataserializable-serialization) and [Portable Serialization](#2-portable-serialization).
```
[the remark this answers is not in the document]
mustafaiman: Also, I would prefer to change `However, JSON Serialization is not the best way of serialization in terms of performance.` to `JSON serialization is not the only way of serialization`. Performance is not the only reason to use a different serialization method. Interoperability between different languages is an equally important motivation to use a different serialization schemes. You can mention that, too.

README.md
```diff
@@ -108,43 +108,94 @@ See [CONFIG.md](CONFIG.md) for details.
...
+- A character which is a single-length string one can be serialized as a `Character` for Java server side.
+- A `number` type which is a primitive on the Node.js client can be serialized as `Byte`, `Short`, `Integer`, `Float`, `Double` for Java server side.
+- A `Long` type which represents a 64 bit two's complement integer value can be serialized as a `Long` for Java server side.
+- A `string` type which is a primitive on the Node.js client can be serialized as a `String` for Java server side.
+- Arrays of the above types can be serialized as `boolean[]`, `char[]`, `byte[]`, `short[]`, `int[]`, `float[]`, `double[]`, `long[]` and `string[]` for Java server side respectively. 
 
-Regarding arrays in a serializable object, you can use methods like `writeIntArray` if the array is of a primitive type.
+Note that if the object is not one of the above-mentioned types, the Node.js client uses `JSON Serialization` by default. You can create your complex object and it is serialized by using `JSON Serialization`. After serialization, the byte array format of your object is sent to the server through network.
```
mustafaiman: I think this part does not provide any valuable information.
```
You can create your complex object and it is serialized by using `JSON Serialization`. After serialization, the byte array format of your object is sent to the server through network.
```
We can remove that part keeping only the first sentence and merge the next paragraph to this one.

README.md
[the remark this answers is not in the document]
furkansenharputlu: I think saying that JSON Serialization is not the best way is important. However, I can mention about interoperability between different languages .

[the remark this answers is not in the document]
furkansenharputlu: I made it like this: `However, `JSON Serialization` is not the best way of serialization in terms of performance and interoperability between the clients in different languages. If you want the serialization to work faster or you use the clients in different languages, Hazelcast offers its own native serialization types, such as [IdentifiedDataSerializable Serialization](#1-identifieddataserializable-serialization) and [Portable Serialization](#2-portable-serialization).
`

README.md
furkansenharputlu: I mean `uint16` type.

README.md
furkansenharputlu: I will remove it.

README.md
mustafaiman: still, it says `a character which is single-length string`. It means `'a'` without a doubt to me.

PR #344, opened 2018-09-05
add documentation for Securing Client Connection chapter

furkansenharputlu:
This PR adds documentation for Securing Client Connection chapter in the issue #340.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,13 @@
   * [4. Setting Connection Timeout](#4-setting-connection-timeout)
   * [5. Setting Connection Attempt Limit](#5-setting-connection-attempt-limit)
   * [6. Setting Connection Attempt Period](#6-setting-connection-attempt-period)
-  * [7. Enabling Hazelcast Cloud Discovery](#7-enabling-hazelcast-cloud-discovery)
+  * [7. Enabling Client TLS/SSL](#7-enabling-client-tlsssl)
+  * [8. Enabling Hazelcast Cloud Discovery](#8-enabling-hazelcast-cloud-discovery)
+* [Securing Client Connection](#securing-client-connection)
+  * [1. TLS/SSL](#1-tlsssl)
+    * [1.1. TLS/SSL for Hazelcast Members](#11-tlsssl-for-hazelcast-members)
+    * [1.2. TLS/SSL for Hazelcast Node.js Clients](#12-tlsssl-for-hazelcast-nodejs-clients)
+    * [1.3. Mutual Authentication](#13-mutual-authentication)
 * [Using Node.js Client with Hazelcast IMDG](#using-nodejs-client-with-hazelcast-imdg)
   * [1. Node.js Client API Overview](#1-nodejs-client-api-overview)
   * [2. Using Distributed Data Structures](#2-using-distributed-data-structures)
@@ -808,7 +814,14 @@ clientConfig.networkConfig.connectionAttemptPeriod = 5000;
 
 Its default value is `3000` milliseconds.
 
-## 7. Enabling Hazelcast Cloud Discovery
+## 7. Enabling Client TLS/SSL
+
+You can use TLS/SSL to secure the connection between the clients and members. If you want TLS/SSL enabled
+for the client-cluster connection, you should set an SSL configuration. Please see [TLS/SSL section](#1-tlsssl).
+
+As explained in the [TLS/SSL section](#1-tlsssl), Hazelcast members have key stores used to identify themselves (to other members) and Hazelcast Node.js clients have certificate authorities used to define which members they can trust. Hazelcast has the mutual authentication feature which allows the Node.js clients also to have their private keys and public certificates and members to have their certificate authorities so that the members can know which clients they can trust. Please see the [Mutual Authentication section](#13-mutual-authentication).
+
+## 8. Enabling Hazelcast Cloud Discovery
 
 The purpose of Hazelcast Cloud Discovery is to provide clients to use IP addresses provided by `hazelcast orchestrator`. To enable Hazelcast Cloud Discovery, specify a token for the `discoveryToken` field and set the `enabled` field to `true`.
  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
 }
 ```
 
+# Setting Up Client Network
+
+### 7. Enabling Client TLS/SSL
+
+You can use TLS/SSL to secure the connection between the client and members. If you want TLS/SSL enabled
```
Serdaro: If there is a possibility to have more than one client, let's use "clients" here.

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
 }
 ```
 
+# Setting Up Client Network
+
+### 7. Enabling Client TLS/SSL
+
```
Serdaro: We are discouraging Hazelcast users to use SSL (the Java ones). I don't know if it also applies to the connections between Node.js and clusters. Just noticing. @mustafaiman ?

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+You can use TLS/SSL to secure the connection between the client and members. If you want TLS/SSL enabled
+for the client-cluster connection, you should set an SSL configuration. Once set, the connection (socket) is
+established out of an `options` object supplied by the user. Please see [TLS/SSL section](#1-tlsssl).
+
+As explained in the [TLS/SSL section](#1-tlsssl), Hazelcast members have key stores used to identify themselves (to other members) and Hazelcast Node.js clients have certificate authorities used to define which members they can trust. Starting with Hazelcast 3.8.1 release, mutual authentication is introduced. This allows the Node.js clients also to have their private keys and public certificates and members to have their certificate authorities so that the members can know which clients they can trust. Please see the [Mutual Authentication section](#13-mutual-authentication).
```
Serdaro: Let's remove the release specific information (remove the sentence starting with "Starting with...". Instead:

"Hazelcast has the mutual authentication feature which allows the Node.js clients also....."

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+
+
+# Securing Client Connection
```
Serdaro: As I see, the only security protocol Node.js client supports is TLS/SSL. 
So, let's remove the "Supported Security Protocols" heading below.

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+
+## 1. Supported Security Protocols
+
+This chapter describes the security features of Hazelcast Node.js Client. These features allow you to perform security activities, such as TLS/SSL and its sub-section Mutual Authentication. All of the Security features explained in this chapter are the features of **Hazelcast IMDG Enterprise** edition.
```
Serdaro: This chapter describes the security features of Hazelcast Node.js Client. These include using TLS/SSL for connections between clients and members (*between clients too?*) and mutual authentication. These security features are the features of **Hazelcast IMDG Enterprise** edition.

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+
+- A Java `trustStore` is a file that includes a list of certificates trusted by your application which is named certificate authority. The equivalent of a trust store is a `ca.pem` file with pem extension at the Node.js client side.
+
+#### 1.1. TLS/SSL for Hazelcast Members
+
+Hazelcast allows you to encrypt socket level communication between Hazelcast members and between Hazelcast clients and members, for end to end encryption. To use it, see [TLS/SSL for Hazelcast Members section](http://docs.hazelcast.org/docs/3.10.4/manual/html-single/index.html#tls-ssl-for-hazelcast-members).
+
+#### 1.2. TLS/SSL for Hazelcast Node.js Clients
+
+Hazelcast Node.js clients which support TLS/SSL should have the following a user supplied SSL `options` object to pass to
```
Serdaro: ...following user supplied SSL `options` object, to pass to...

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+    ca: [fs.readFileSync(__dirname + '/server-cert.pem')],
+    servername: 'foo.bar.com'
+};
+```
+
+#### 1.3. Mutual Authentication
+
+As explained above, Hazelcast members have key stores used to identify themselves (to other members) and Hazelcast clients have trust stores used to define which members they can trust.
+
+Starting with Hazelcast 3.8.1, mutual authentication is introduced. This allows the clients also to have their key stores and members to have their trust stores so that the members can know which clients they can trust.
```
Serdaro: Instead of the "Starting with..." sentence:

"Using mutual authentication, the clients also have their key stores and members have their trust stores so that..."

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+
+To enable mutual authentication, you need to set the following property at the Java member side:
```
Serdaro: Do we need to say "Java member" here?

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+Hazelcast Node.js Client includes a utility factory class that creates the necessary `options` object out of the supplied
+properties. All you need to do is specifying your factory as `BasicSSLOptionsFactory` and provide the following options:
+
+    caPath
+    keyPath
+    certPath
+    servername
+    rejectUnauthorized
+    ciphers
+
```
Serdaro: Should be bulleted. Otherwise, it will look like a squeezed text I guess.

README.md
mustafaiman: I was not aware of such discouragement. I will check with the team.

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
 }
 ```
 
+# Setting Up Client Network
+
+### 7. Enabling Client TLS/SSL
+
+You can use TLS/SSL to secure the connection between the client and members. If you want TLS/SSL enabled
+for the client-cluster connection, you should set an SSL configuration. Once set, the connection (socket) is
+established out of an `options` object supplied by the user. Please see [TLS/SSL section](#1-tlsssl).
```
mustafaiman: If this is only an introduction of the feature to lead the users to real chapter, then `Once set, the connection (socket) is established out of an 'options' object supplied by the user.` is not neccessary. It lacks context and is confusing I think.

README.md
mustafaiman: I do not know if this is an issue but this section comes before `TLS/SSL` section. Therefore `As explained in the [TLS/SSL section]` seems odd to me. @Serdaro wdyt?

README.md
mustafaiman: I would prefer `requires Hazelcast IMDG Enterprise` instead of `are features of Hazelcast IMDG Enterprise`.

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+
+## 1. Supported Security Protocols
+
+This chapter describes the security features of Hazelcast Node.js Client. These features allow you to perform security activities, such as TLS/SSL and its sub-section Mutual Authentication. All of the Security features explained in this chapter are the features of **Hazelcast IMDG Enterprise** edition.
+
+### 1. TLS/SSL
+
+One of the offers of Hazelcast is the TLS/SSL protocol which you can use to establish an encrypted communication across your cluster with key stores and trust stores.
+
+- A Java `keyStore` is a file that includes a private key and a public certificate. The equivalent of a key store is the combination of `key.pem` and `cert.pem` files with pem extension at the Node.js client side.
```
mustafaiman: Do we really care about the extensions in the implementation? I guess it would work with any extension.

mustafaiman: I think we lack a little bit of context here. `keyStore` has come out of nowhere. I would prefer an easier introduction to the topic. You can first basically explain that clients authorize servers by looking at their certificates. Clients use their list of trusted certificates to decide if they should connect to the server. The same goes other way around too. Then you can tell that in Java, you store your certificates in a file named `keyStore` and list of trusted certificates in `trustStore`.

Trying to squeeze meaning in a few sentences is not the way to go in this guide. Too long is better than too short I think.

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+
+- A Java `trustStore` is a file that includes a list of certificates trusted by your application which is named certificate authority. The equivalent of a trust store is a `ca.pem` file with pem extension at the Node.js client side.
+
+#### 1.1. TLS/SSL for Hazelcast Members
+
+Hazelcast allows you to encrypt socket level communication between Hazelcast members and between Hazelcast clients and members, for end to end encryption. To use it, see [TLS/SSL for Hazelcast Members section](http://docs.hazelcast.org/docs/3.10.4/manual/html-single/index.html#tls-ssl-for-hazelcast-members).
```
mustafaiman: The first sentence looks like it would be a better fit for the introduction part (`###TLS/SSL`). I think we do not need "TLS/SSL for Hazelcast Members" and "TLS/SSL for Hazelcast Node.js Clients" headings. We do not care about the encryption between members. We care about the encryption between member and client. You already started talking about truststores and stuff before. Why don't you put here a paragraph or two saying "in order to use ssl, you need to configure your keystore on the server side. this is how you do it. then you need to configure your list of trusted certificates on the client side. this is how you do it. by the way you can configure any option stated in node.js ssl documentation by simply adding it to our 'options' object."

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+#### 1.3. Mutual Authentication
+
+As explained above, Hazelcast members have key stores used to identify themselves (to other members) and Hazelcast clients have trust stores used to define which members they can trust.
+
+Starting with Hazelcast 3.8.1, mutual authentication is introduced. This allows the clients also to have their key stores and members to have their trust stores so that the members can know which clients they can trust.
+
+To enable mutual authentication, you need to set the following property at the Java member side:
+
+```java
+props.setProperty("javax.net.ssl.mutualAuthentication", "REQUIRED");
```
mustafaiman: keep in mind that the audience should not be writing java code. They are using standalone servers and config file. Give xml configuration instead.

README.md
mustafaiman: ok, i read the rest and this part is explained later ` then you need to configure your list of trusted certificates on the client side. this is how you do it. by the way you can configure any option stated in node.js ssl documentation by simply adding it to our 'options' object.`. I am fine with that but we still need a parahraph to show how to set keystore and truststore for the member

README.md
```diff
@@ -145,6 +156,174 @@ readData(dataInput) {
...
+
+1. Using the built-in `BasicSSLOptionsFactory` bundled with the client.
+2. Writing an SSLOptionsFactory.
+
+Below subsections describe each way.
+
+**Using Built-in BasicSSLOptionsFactory**
+
+Hazelcast Node.js Client includes a utility factory class that creates the necessary `options` object out of the supplied
+properties. All you need to do is specifying your factory as `BasicSSLOptionsFactory` and provide the following options:
```
mustafaiman: how do they specify their factory as `BasicSSLOptionsFactory`?

README.md
furkansenharputlu: I will merge these two.

README.md
furkansenharputlu: `To use it, see [TLS/SSL for Hazelcast Members section](http://docs.hazelcast.org/docs/latest/manual/html-single/index.html#tls-ssl-for-hazelcast-members).`
 This guides the user to the Reference Manual. He can see how to set `keyStore` and `trustStore` there. 

Also, I will add `You should set keyStore and trustStore before starting the members. See the next section how to set keyStore and trustStore on the server side.` before `TLS/SSL for Hazelcast Members` heading.

README.md
furkansenharputlu: Do you mean configuring `hazelcast-client.json` ?

mustafaiman: yes, json sample

PR #346, opened 2018-09-05
add documentation for Node.js Client API Overview and some data struc…

furkansenharputlu:
…tures

This PR adds documentation for Node.js Client API Overview and some data structures in the issue #340.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,13 @@
 * [Using the Client](#using-the-client)
 * [Code Samples](#code-samples)
 * [Serialization Considerations](#serialization-considerations)
+* [Using Node.js Client with Hazelcast IMDG](#using-nodejs-client-with-hazelcast-imdg)
+  * [1. Node.js Client API Overview](#1-nodejs-client-api-overview)
+  * [2. Using Distributed Data Structures](#2-using-distributed-data-structures)
+    * [2.1. Using Map](#21-using-map)
+    * [2.2. Using MultiMap](#22-using-multimap)
+    * [2.3. Using ReplicatedMap](#23-using-replicatedmap)
+    * [2.4. Using Queue](#24-using-queue)
 * [Development](#development)
   * [Building And Installing from Sources](#building-and-installing-from-sources)
   * [Using Locally Installed Package](#using-locally-installed-package)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,6 +152,117 @@ readData(dataInput) {
 }
 ```
 
+# Using Node.js Client with Hazelcast IMDG
+
+## 1. Node.js Client API Overview
+
+The first step is configuration. You can configure the Node.js client declaratively or programmatically. We will use the programmatic approach throughout this chapter. Please refer to the [Node.js Client Declarative Configuration section](#declarative-configuration) for details.
+
```
mustafaiman: I think this chapter is a good place to mention that most of the API is promise based and they can use these in async/await fashion too if they like. I saw some questions on gitter that the user calls a lock and they just do not wait for lock call to be resolved. Then they complain that the locks do not work. I think if we can convey the general usage of the api to the user. It will be easier for them to understand everything.

README.md
```diff
@@ -145,6 +152,117 @@ readData(dataInput) {
...
+```javascript
+var clientConfig = new Config.ClientConfig();
+clientConfig.groupConfig.name = 'dev';
+clientConfig.networkConfig.addresses.push('10.90.0.1', '10.90.0.2:5702');
```
mustafaiman: i think one address would be enough in here. One being with port and the other being without port is also confusing.

README.md
```diff
@@ -145,6 +152,117 @@ readData(dataInput) {
...
+}).then(function () {
+    return mapCustomers.put('3', new Customer("Muhammet", "Ali"));
+});
+```
+
+As a final step, if you are done with your client, you can shut it down as shown below. This will release all the used resources and will close connections to the cluster.
+
+```javascript
+...
+.then(function () {
```
mustafaiman: When I see this snippet, I would assume that I can just add this as a last step to the above chain, it would work. However, it will not work because `client` is defined in scope of only the first closure above. `client` here would be undefined and error

README.md
```diff
@@ -145,6 +152,117 @@ readData(dataInput) {
...
+    client.shutdown();
+});
+```
+
+## 2. Using Distributed Data Structures
+
+Most of the Distributed Data Structures are supported by the Node.js client. When you use clients in other languages, you should check for the exceptions.
```
mustafaiman: I do not understand what you mean by `check for the exceptions` here

README.md
```diff
@@ -145,6 +152,117 @@ readData(dataInput) {
...
+
+As a general rule, you configure these data structures on the server side and access them through a proxy on the client side.
```
mustafaiman: This implies that you have to do configuration on the server side in order to use these data structures. I am trying to look from the point of view of a developer with zero Hazelcast experience. I would say how am I gonna configure these on the server?

README.md
```diff
@@ -145,6 +152,117 @@ readData(dataInput) {
...
+
+## 2.1. Using Map
```
mustafaiman: These are just samples but not individual subsections, right? I do not think they need `#` title and title number.

README.md
furkansenharputlu: I want to be consistent with Java documentatiton. See http://docs.hazelcast.org/docs/3.10.4/manual/html-single/index.html#client-api

README.md
furkansenharputlu: It says maybe some data structures is not supported by other languages.

README.md
furkansenharputlu: Again, it was like that in Java Documentation. See http://docs.hazelcast.org/docs/3.10.4/manual/html-single/index.html#using-supported-distributed-data-structures

README.md
furkansenharputlu: I will add them to the `Table of Contents` at the end so users can access them easily.

README.md
Serdaro: Why do we need to mention about the clients in other languages here?

README.md
Serdaro: Java documentation mentioned here is a part of a huge Hazelcast IMDG reference manual. But, Node.js client ReadMe document is not. So, we need to be specific here. Giving a link to the IMDG reference manual's Distributed Data Structures chapter would work.

furkansenharputlu: @mustafaiman I agree that this sentence implies that the user has to do something on the server. I will remove it.

README.md
furkansenharputlu: I will remove it.

README.md
furkansenharputlu: I will add a `var client;` in the first snippet.

README.md
```diff
@@ -145,6 +152,119 @@ readData(dataInput) {
 }
 ```
 
+# Using Node.js Client with Hazelcast IMDG
+
+## 1. Node.js Client API Overview
+
+In the API, most of the functions return `Promise`. You need to be able to use promises to use the client. Promises provide a better way of working with callbacks. You can chain asynchronous functions by `then()` function of promise. Also, you can use `async/await`, if you use Node v8 and higher. If you are not familiar with promises, please first learn how to use them. If you are ready to go, let's start to use Hazelcast Node.js Client!
```
mustafaiman: `If you are not familiar with promises, please first learn how to use them.` sounds kind of aggressive to me. I am not objecting it, just making a note. I leave the language to @Serdaro

Serdaro: My suggestion for the whole paragraph:

Most of the functions in the API return `Promise`. Therefore, you need to be familiar with the concept of promises to use the Node.js client. If not, you can learn about them using various online resources. 

Promises provide a better way of working with callbacks. You can chain asynchronous functions by `then()` function of promise. Also, you can use `async/await`, if you use Node.js 8 and higher versions. 

If you are ready to go, let's start to use Hazelcast Node.js client!

mustafaiman:
@Serdaro @furkansenharputlu is there any work left here?

PR #352, opened 2018-09-10
add default and sample configuration files

furkansenharputlu:


hazelcast-client-sample.json
```diff
@@ -0,0 +1,130 @@
+{
```
mustafaiman: I think a sample should be short. Its purpose is to convey what the configuration would look like. This is full configuration and it already exists in the repo as `hazelcast-client-full.json`

mustafaiman:
sample file addition, merging with 1 ok

PR #356, opened 2018-09-13
update the version in package-lock.json

### LINKED ISSUE #338: package-lock.json stuck with version 0.8
Apparently package-lock.json did not get updated during the release process. Client version is still 0.8 there although package.json is at 0.9.0. We need to

- [x] Check if there is any adverse effect of this discrepancy for our users
- [ ] Update release process to keep these in sync

furkansenharputlu:
fixes #338

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "hazelcast-client",
-    "version": "0.8.0",
+    "version": "0.9.0",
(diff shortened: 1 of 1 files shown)


mustafaiman:
how do we make sure that they will not go out of sync again?

furkansenharputlu:
I tried in my local. In the current release process, there isn’t a reason for inequality.

PR #360, opened 2018-09-19
add MemberAttributeEvent and refactor membership listener

furkansenharputlu:


src/invocation/ClusterService.ts
```diff
@@ -262,7 +271,9 @@ export class ClusterService extends EventEmitter {
     }
 
     private handleMemberAttributeChange(uuid: string, key: string, operationType: number, value: string): void {
-        this.emit(EMIT_ATTRIBUTE_CHANGE, uuid, key, ATTRIBUTE_CHANGE[operationType], value);
+        const member = this.getMember(uuid);
+        const memberAttributeEvent = new MemberAttributeEvent(member, key, ATTRIBUTE_CHANGE[operationType], value);
```
mustafaiman: If we change operationType in MemberAttributeEvent to an enum, we can get rid of ugly `ATTRIBUTE_CHANGE` object, too.

src/core/MemberAttributeEvent.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * the key for this MemberAttributeEvent.
+     */
+    key: string;
+
+    /**
+     * the type of member attribute change for this MemberAttributeEvent.
+     */
+    operationType: string;
```
mustafaiman: We can make this an enum as in Java client.

PR #362, opened 2018-09-19
add DistributedObjectEvent and refactor DistibutedObjectListener

furkansenharputlu:
This PR adds `DistibutedObjectEvent` and refactors `DistibutedObjectListener`.


src/core/DistributedObjectListener.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * DistributedObjectEvent is fired when a DistributedObject
+ * is created or destroyed cluster-wide.
+ */
+export class DistributedObjectEvent {
+
+    /**
+     * The type of this event; one of 'created' or 'destroyed'.
+     */
+    eventType: string;
```
mustafaiman: should we make this an enum instead of string?

furkansenharputlu: I think no because `eventType` comes from server as a string.

PR #364, opened 2018-09-25
Add redoOperation implementation

### LINKED ISSUE #345: add missing redoOperation key to network in ConfigBuilder
`redoOperation` key is missing in `handleNetwork` method in ConfigBuilder.

ubeydk:
Implementation and related config properties have been added. In addition, unit tests are passed succesfully in my local computer, but currently I didn't add any unit test for **redoOperation**.

Fixes #345


CONFIG.md
```diff
@@ -132,6 +134,20 @@ Smart routing defines whether the client mode is smart or unisocket. The followi
 
 Default is smart routing mode.
 
+### Enabling Redo Operation
+
+Specifies whether the redo operations are enabled or not.
```
mustafaiman: Actually we can use Java Client's description of the switch here. It is better
```
     * If true, client will redo the operations that were executing on the server and client lost the connection.
     * This can be because of network, or simply because the member died. However it is not clear whether the
     * application is performed or not. For idempotent operations this is harmless, but for non idempotent ones
     * retrying can cause to undesirable effects. Note that the redo can perform on any member.
```

ubeydk:
@mustafaiman, Thanks for your suggestions. I will try to add tests you suggested.

Holmistr:
verify

ubeydk:
I will refine this PR in few weeks. 

Currently, I do not have much spare time, but I can try to hurry if it needs to be done quickly. @mustafaiman please ping me if needed.

mdumandag:
Hi @ubeydk , there is no need to rush. 

If you have further questions you can also contact me

ubeydk:
I have enabled redoOperation config for the tests in [ClientReconnectTest.js](https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/test/ClientReconnectTest.js) and [ListenersOnReconnectTest.js](https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/test/ListenersOnReconnectTest.js) because reconnection tests are made by invoking map.put operation but client shouldn't retry these operations because it is not idempotent as I mentioned in the previous comment.

What do you think about it? @mdumandag

puzpuzpuz:
Closing this one as the option was implemented in #590

PR #368, opened 2018-10-01
add AddressHelper object and refactor the connection behavior

furkansenharputlu:
This PR changes the connection behaviour. Currently, the client tries just `localhost:5701` address while connecting to cluster. It should try `5071`, `5072`, `5073` when port number is not defined.


furkansenharputlu:
verify

Issue #370, opened 2018-10-01
getOrCreateProxy function should return Promise


furkansenharputlu:
`getOrCreateProxy` function does not return Promise. It should return Promise. This affects all distributed data structure get methods(getMap, getList, getSet..). After the change, the followings should be made:
- [x] Refactor tests
- [x] Refactor code samples on Node.js repo
- [x] Refactor code samples on https://github.com/hazelcast/hazelcast-code-samples/tree/master/clients/nodejs
- [x] Refactor documentation
- [x] Update the code samples on org website at release time

degerhz:
A ticket for website update request has been created. Closing this last remaining issue in this milestone.

PR #372, opened 2018-10-03
fix Client API get functions

furkansenharputlu:
This PR fixes the Client API get functions and tests. #370

### DIFF
--- a/test/ListenerServiceTest.js
+++ b/test/ListenerServiceTest.js
@@ -18,6 +18,7 @@ var RC = require('./RC');
 var HazelcastClient = require('../.').Client;
 var Config = require('../.').Config;
 var expect = require('chai').expect;
+var Util = require('./Util');
 
 
 [true, false].forEach(function (isSmartService) {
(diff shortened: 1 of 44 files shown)


test/ListenerServiceTest.js
```diff
@@ -18,68 +18,73 @@ var RC = require('./RC');
...
 var Config = require('../.').Config;
 var expect = require('chai').expect;
+var Util = require('./Util');
 
 
-[true, false].forEach(function (isSmartService) {
-    describe('ListenerService[smart=' + isSmartService + ']', function () {
-        var cluster;
-        var client;
+var isSmartService = true;
```
sancar: It looks like we are not testing `isSmart =false` anymore.

test/list/ListProxyTest.js
```diff
@@ -38,7 +38,9 @@ describe("List Proxy", function () {
     });
 
     beforeEach(function () {
-        listInstance = client.getList('test')
+        return client.getList('test').then(function (list) {
```
sancar: Does this guarantee that listInstance is set or should we call some sort of wait on promise before returning.

furkansenharputlu: Yes.

PR #375, opened 2018-10-10
make ConfigBuilder accessible by user

### LINKED ISSUE #366: ConfigBuilder is not exposed
ConfigBuilder should be exposed in index.ts, so users can use it to extend configuration. Currently it is private.

furkansenharputlu:
This PR makes `ConfigBuilder` accessible. It fixes #366


mustafaiman:
could you also change ConfigBuilder tests to use the public api?

furkansenharputlu:
verify

PR #376, opened 2018-10-12
add documentation for Portable Serialization

furkansenharputlu:
This PR adds documentation for Portable Serialization. #248  #340


mustafaiman:
Also we can give a link to documentation or samples for java side and stress that Java side has to have the same implementation too.

furkansenharputlu:
I will explain the `Java side has to have the same implementation too.` in another section in detail.

PR #377, opened 2018-10-15
fix portable factory declarative configuration

furkansenharputlu:
This PR fixes the bug in declarative configuration of portable factories. It also adds tests for declarative configuration of serialization.


furkansenharputlu:
verify

PR #381, opened 2018-10-19
Serialization doc

furkansenharputlu:
This PR adds
- test for Custom Serialization declarative configuration
- Custom Serialization documentation #340


README.md
```diff
@@ -772,8 +776,128 @@ Note that the id that is passed to the `SerializationConfig` is same as the `fac
 
 ## 3. Custom Serialization
 
+Hazelcast lets you to plug a custom serializer to be used for serialization of objects.
```
mustafaiman: ...lets **you plug** a...

README.md
```diff
@@ -772,8 +776,128 @@ Note that the id that is passed to the `SerializationConfig` is same as the `fac
...
+
+Let's say you have an object `Musician` and you would like to customize the serialization. The reason could be you are not happy with the default serialization.
```
mustafaiman: I think the most common reason for using a custom serializer is that the user wants to use an external serializer for only one object but let Hazelcast deal with the others. 

`The reason could be you are not happy with the default serialization.` I think this sentence does not convey any information.

README.md
```diff
@@ -772,8 +776,128 @@ Note that the id that is passed to the `SerializationConfig` is same as the `fac
...
+
+```javascript
+function Musician(name) {
+    this.name = name;
+}
```
mustafaiman: This object must have a `hzGetCustomId` method that will return the custom serialization id(10 in this example) AFAIK.

test/serialization/config/CustomSerializerTest.js
```diff
@@ -0,0 +1,53 @@
...
+            return Controller.startMember(cluster.id);
+        });
+    });
+
+    after(function () {
+        Controller.shutdownCluster(cluster.id);
+        process.chdir('../../../');
+    })
+
+    it('should be configured declaratively', function () {
```
mustafaiman: This test does not verify custom serializer is used. The system can fallback to Json serializer and the test would still pass. Moreover I think this is what happens right now. Since `Musician` does not have `hzGetCustomId` method, it cannot match the custom serializer to musicion object.

README.md
```diff
@@ -772,8 +776,128 @@ Note that the id that is passed to the `SerializationConfig` is same as the `fac
...
+    var len = objectDataInput.readInt();
+    var name = '';
+    for (var i = 0; i < len; i++) {
+        name = name + String.fromCharCode(objectDataInput.readInt());
+    }
+    return new Musician(name);
+}
+```
+
+Note that the serializer id must be unique as Hazelcast will use it to lookup the `MusicianSerializer` while it deserializes the object. Now the last required step is to register the `MusicianSerializer` to the configuration.
```
Serdaro: ...serializer `id`...

README.md
furkansenharputlu: I will change it to `The reason may be you want to use an external serializer for only one object.`

PR #396, opened 2018-11-05
refactor item listeners

furkansenharputlu:
This PR refactor item listeners for `IList`, `ISet` and `IQueue` .

### DIFF
--- a/src/core/ItemListener.ts
+++ b/src/core/ItemListener.ts
@@ -16,12 +16,51 @@
 
 import {Member} from './Member';
 
+/**
+ * Item listener for IQueue, ISet, IList.
+ */
 export interface ItemListener<E> {
     itemAdded?: ItemEventListener<E>;
     itemRemoved?: ItemEventListener<E>;
 }
 
-export type ItemEventListener<E> = (item: E, member: Member, eventType: ItemEventType) => void;
+/**
+ * A type which is used for item events.
+ */
+export type ItemEventListener<E> = (itemEvent: ItemEvent<E>) => void;
+
+/**
+ * IQueue, ISet, IList item event.
+ */
+export class ItemEvent<E> {
+
+    /**
+     * The name of the data structure for this event.
+     */
+    name: string;
+
+    /**
+     * The value of the item event.
+     */
+    item: E;
+    /**
+     * The event type.
+     */
+    eventType: ItemEventType;
+
+    /**
+     * The member that fired this event.
+     */
+    member: Member;
+
+    constructor(name: string, itemEventType: ItemEventType, item: E, member: Member) {
+        this.name = name;
+        this.eventType = itemEventType;
+        this.item = item;
+        this.member = member;
+    }
+
+}
 
 export enum ItemEventType {
     ADDED = 1,
--- a/src/proxy/ListProxy.ts
+++ b/src/proxy/ListProxy.ts
@@ -36,7 +36,7 @@ import {ListRemoveWithIndexCodec} from '../codec/ListRemoveWithIndexCodec';
 import {ListSetCodec} from '../codec/ListSetCodec';
 import {ListSizeCodec} from '../codec/ListSizeCodec';
 import {ListSubCodec} from '../codec/ListSubCodec';
-import {ItemListener} from '../core/ItemListener';
+import {ItemEvent, ItemEventType, ItemListener} from '../core/ItemListener';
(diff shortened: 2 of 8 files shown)


src/core/ItemListener.ts
```diff
@@ -16,12 +16,51 @@
...
-export type ItemEventListener<E> = (item: E, member: Member, eventType: ItemEventType) => void;
+/**
+ * A type which is used for item events.
+ */
+export type ItemEventListener<E> = (itemEvent: ItemEvent<E>) => void;
+
+/**
+ * IQueue, ISet, IList item event.
+ */
+export class ItemEvent<E> {
```
mustafaiman: This must be exported in `index.ts`

src/proxy/ListProxy.ts
```diff
@@ -132,15 +132,15 @@ export class ListProxy<E> extends PartitionSpecificProxy implements IList<E> {
...
-                let listenerFunction: Function;
-                if (eventType === 1) {
-                    listenerFunction = listener.itemAdded;
-                } else if (eventType === 2) {
-                    listenerFunction = listener.itemRemoved;
-                }
 
-                if (listenerFunction) {
-                    listenerFunction.call(responseObject, responseObject);
+                const member = this.client.getClusterService().getMember(uuid);
```
mustafaiman: Minor comment: Looks like this part is repeated. Is it possible to re-use this part in different proxies?

test/list/ListProxyTest.js
```diff
@@ -259,15 +260,41 @@ describe("List Proxy", function () {
...
+                expect(itemEvent.item).to.be.equal(2);
+                expect(itemEvent.eventType).to.be.equal(ItemEventType.ADDED);
+                expect(itemEvent.member).to.not.be.equal(null);
+            },
+            itemRemoved: function (itemEvent) {
+                expect(itemEvent.name).to.be.equal('test');
+                expect(itemEvent.item).to.be.equal(2);
+                expect(itemEvent.eventType).to.be.equal(ItemEventType.REMOVED);
+                expect(itemEvent.member).to.not.be.equal(null);
+                done();
```
mustafaiman: if one of the above assertions fail, the test will never reach any `done()` or `done(err)` calls, right? Then the test will fail with a time-out rather than the real reason of assertion failure. How about this:
```
         try {
                expect(itemEvent.name).to.be.equal('test');
                expect(itemEvent.item).to.be.equal(2);
                expect(itemEvent.eventType).to.be.equal(ItemEventType.REMOVED);
                expect(itemEvent.member).to.not.be.equal(null);
                done();
          } catch(e) {
                done(e);
          }
```

test/list/ListProxyTest.js
```diff
@@ -259,15 +260,41 @@ describe("List Proxy", function () {
         })
     });
 
+    it("listens for added and removed entry", function (done) {
+        this.timeout(5000);
+        var listener = {
+            itemAdded: function (itemEvent) {
```
mustafaiman: even if this method is never called, the test would pass.

test/list/ListProxyTest.js
mustafaiman: Above comment applies to all of the similar tests in this pr.

test/queue/QueueProxyTest.js
```diff
@@ -14,16 +14,18 @@
...
+var chaiAsPromised = require('chai-as-promised');
 chai.use(chaiAsPromised);
-var HazelcastClient = require("../../lib/index.js").Client;
 var Controller = require('./../RC');
 var Util = require('./../Util');
 var Promise = require('bluebird');
 var fs = require('fs');
 
+var HazelcastClient = require('../../').Client;
+var ItemEventType = require('../../lib/core/ItemListener').ItemEventType;
```
mustafaiman: `ItemEventType` should be retrieved via `require('../../').ItemEventType`

test/list/ListProxyTest.js
furkansenharputlu: It fails with the real reason.
Let say `expect(itemEvent.item).to.be.equal(234);`. The test fails with assertion rather than time-out.

test/list/ListProxyTest.js
furkansenharputlu: I will add a boolean check for it.

src/proxy/ListProxy.ts
furkansenharputlu: I think it is not a redundancy. It accesses the member with just 2 hops.

PR #397, opened 2018-11-05
add listeners documentation for data structures

furkansenharputlu:
This PR adds documentation for listening for data structures. #340

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,10 @@
       * [7.5.1.2. Listening for Distributed Object Events](#7512-listening-for-distributed-object-events)
       * [7.5.1.3. Listening for Lifecycle Events](#7513-listening-for-lifecycle-events)
     * [7.5.2. Distributed Data Structure Events](#752-distributed-data-structure-events)
-      * [7.5.2.1. Listening for Map Events](#7521-listening-for-map-events)
+      * [7.5.2.1. Map Listener](#7521-map-listener)
+      * [7.5.2.2. Entry Listener](#7522-entry-listener)
+      * [7.5.2.3. Item Listener](#7523-item-listener)           
+      * [7.5.2.4. Message Listener](#7524-message-listener)      
   * [7.6. Distributed Computing](#76-distributed-computing)
     * [7.6.1. Using EntryProcessor](#761-using-entryprocessor)
   * [7.7. Distributed Query](#77-distributed-query)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1920,6 +1926,174 @@ map.addEntryListener(mapEventListener).then(function () {
...
+
+See the following example.
+
+```javascript
+var entryEventListener = {
+    added: function (entryEvent) {
+        console.log('Entry Added:', entryEvent.key, entryEvent.value); // Entry Added: 1 Furkan
+    }
+};
+return mmp.addEntryListener(entryEventListener, undefined, true).then(function () {
```
mustafaiman: I think we should explain what the second and third arguments do since they are in the example.

README.md
```diff
@@ -1920,6 +1926,174 @@ map.addEntryListener(mapEventListener).then(function () {
...
+```javascript
+var itemListener = {
+    itemAdded: function (itemEvent) {
+        console.log('Item Added:', itemEvent.item); // Item Added: Furkan
+    },
+    itemRemoved: function (itemEvent) {
+        console.log('Item Removed:', itemEvent.item); // Item Removed: Furkan
+    }
+};
+return list.addItemListener(itemListener, true).then(function () {
```
mustafaiman: Should explain what the second argument does and why it matters. If they do not include value in the event, the message is shorter. Also the server does not have to serialize the value, the client does not have to deserialie the value. Therefore, whole conversation is faster.

furkansenharputlu:
I think explaining for all data structures is not redundancy because a user wanting to listen `Queue` events should just go to the `Listening for Queue Events` section. @Serdaro Wdyt ?

Serdaro:
@furkansenharputlu, @mustafaiman. If we are going to change the look'n'feel of this section, we should also apply it to the "Cluster Events" section for the sake of consistency. So, something similar to the below layout:

...
7.5.1 Listening for Cluster Events
  7.5.1.1 Membership Listener
  7.5.1.2 DistributedObject Listener
  7.5.1.3 Lifecycle Listener
7.5.2 Listening for Distributed Data Structure Events
  7.5.2.1 Map Listener
     Content on how to listen map and map entry events.
  7.5.2.2 Entry Listener
     Content on how to listen events for multimap, replicated map and the entries within them.
  7.5.2.3 Item Listener
      Content on how to listen events for queue, set, list and the items within them.
  7.5.2.4 Message Listener
       So on and so forth

I think the above structure looks simpler and cleaner. In order not to repeat/reuse a content in the other sections, maybe we can follow the above approach and re-arrange the content within these sections to make the structure types (map, multimap, set, etc.) more visible.

Serdaro:
@furkansenharputlu , Does this need further improvements?

PR #400, opened 2018-11-07
add Partition Aware documentation

furkansenharputlu:
This PR adds documentation for Partition Aware. #340


README.md
```diff
@@ -2398,8 +2400,85 @@ hazelcastClient.getMap('brothersMap').then(function (mp) {
 });
 ```
 
-## 7.9. Monitoring and Logging
+## 7.8. Performance
+
```
Serdaro: I think we need to mention the "performance" here, since when I read the following, I don't see an explicit explanation on how PartitionAware affects the performance. A content explaining this would be useful, either here or within the section 7.8.1 below.

README.md
```diff
@@ -2398,8 +2400,87 @@ hazelcastClient.getMap('brothersMap').then(function (mp) {
...
+    // lock operation will still execute on the same member
+    // of the cluster since the key ("key1") is same
+    return hazelcastClient.getLock('key1');
+}).then(function (l) {
+    lock = l;
+    return lock.lock();
+});
+```
+
+When the keys are the same, entries are stored on the same member. However, we sometimes want to have related entries stored on the same member, such as a customer and his/her order entries. We would have a customers map with `customerId` as the key and an orders map with `orderId` as the key. Since `customerId` and `orderId` are different keys, a customer and his/her orders may fall into different members in your cluster. So how can we have them stored on the same member? We create an affinity between customer and orders. If we make them part of the same partition then these entries will be co-located. We achieve this by making `orderId`s `PartitionAware`.
```
mustafaiman: "by making `OrderKey`s `PartitionAware`" here

README.md
```diff
@@ -2398,8 +2400,87 @@ hazelcastClient.getMap('brothersMap').then(function (mp) {
...
+}).then(function () {
+    // now create the orders for this customer
+    return mapOrders.putAll([
+        [new OrderKey(21, 1), order],
+        [new OrderKey(22, 1), order],
+        [new OrderKey(23, 1), order]
+    ]);
+});
+```  
+
```
mustafaiman: we can add a link to the reference manual

README.md
furkansenharputlu: It should also be changed in Hazelcast IMDG Reference Manual.

README.md
```diff
@@ -2480,6 +2480,8 @@ Client.newHazelcastClient().then(function (client) {
 });
 ```  
 
+For more details, see [PartitionAware section](https://docs.hazelcast.org/docs/latest/manual/html-single/#partitionaware) in the Hazelcast IMDG Reference Manual.
```
Serdaro: ... details, see the [PartitionAware ...

PR #402, opened 2018-11-09
refactor interface names as Java

furkansenharputlu:


mustafaiman:
Please check that we did not use these old names IReplicatedMap, IRingBuffer in the guide.

PR #404, opened 2018-11-12
refactor ssl behavior and add tests

### LINKED ISSUE #403: Client should connect to server when SSL is enabled by default
- [x] The implementation should be changed.
- [x] New tests should be added.
- [x] The documentation should be changed.

furkansenharputlu:
This PR changes the behavior of SSL. When user just enables the SSL, the client will be able to connect by default. fixes #403


test/declarative_config/ConfigBuilderSSLTest.js
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ */
+
+var chai = require('chai');
+chai.use(require('chai-as-promised'));
+var expect = chai.expect;
+var path = require('path');
+
+var ConfigBuilder = require('../../').ConfigBuilder;
+var Config = require('../../lib/index').Config;
```
mustafaiman: should use public API: `require('../../').Config`

test/declarative_config/configurations/invalid-ssl-factory.json
```diff
@@ -3,7 +3,7 @@
         "ssl": {
             "enabled": true,
             "factory": {
-                "exportedName": "BsicSSLOponsFary",
+                "exportedName": "MySSLOptionsFactory",
```
mustafaiman: as a minor improvement, we could name this as something like `NonExistingOptionsFactory` to communicate that this factory name is intended to be wrong.

test/ssl/ClientSSLTest.js
```diff
@@ -14,77 +14,66 @@
...
+var Promise = require('bluebird');
+var path = require('path');
 
-    describe("SSL rejectUnauthorized:" + cfg.networkConfig.sslOptions.rejectUnauthorized, function () {
+var markEnterprise = require('../Util').markEnterprise;
+var Controller = require('./../RC');
 
-        var cluster;
-        var client;
+var HazelcastClient = require("../../lib/index.js").Client;
```
mustafaiman: should use public api

furkansenharputlu:
Verify

furkansenharputlu:
Verify

mustafaiman:
verify

PR #422, opened 2018-11-26
Refactor logging service

furkansenharputlu:
Logging Service design wasn't good. This PR refactors it. User will be able to specify the log level thanks to this PR. Also, it makes `OFF` a log level.


src/logging/LoggingService.ts
```diff
@@ -14,77 +14,38 @@
...
-    }
-
-    info(className: string, message: string, furtherInfo: any = null): void {
-        this.log(LogLevel.INFO, className, message, furtherInfo);
-    }
-
-    debug(className: string, message: string, furtherInfo: any = null): void {
-        this.log(LogLevel.DEBUG, className, message, furtherInfo);
+    isLogger(loggingProperty: Property): loggingProperty is ILogger {
+        return (loggingProperty as ILogger).log !== undefined;
```
sancar: Is checking one method correct here ? We will be calling `warn` `err` etc on this. They will gave error when used, which will be too late.

src/aggregation/AggregatorFactory.ts
```diff
@@ -56,7 +56,7 @@ export class AggregatorFactory implements IdentifiedDataSerializableFactory {
     static readonly MAX_BY = 17; // needs object to implement Java's Comparable interface
     static readonly MIN_BY = 18; // needs object to implement Java's Comparable interface
 
-    private logger = LoggingService.getLoggingService();
+    private logger: ILogger;
```
mustafaiman: Where is this `logger` initialized? Is it used at all?

src/config/Properties.ts
```diff
@@ -14,8 +14,10 @@
...
  */
 
-import {ILogger} from '../logging/LoggingService';
+import {ILogger} from '../logging/ILogger';
 
 export interface Properties {
     [prop: string]: string | number | boolean | ILogger;
 }
+
+export type Property = string | number | boolean | ILogger;
```
mustafaiman: can we use this `Property` above, too?
```typescript
export interface Properties {
    [prop: string]: Property
}
```

src/logging/ILogger.ts
```diff
@@ -0,0 +1,72 @@
...
+import {LogLevel} from './LoggingService';
+
+/**
+ * The Hazelcast logging interface.
+ */
+export interface ILogger {
+
+    /**
+     * Logs a message with an associated objectName and furtherInfo at the given level.
+     * @param {LogLevel} the log level
```
mustafaiman: Is this the correct format for TSDoc? I think the first argument after `@param` should be the name of the argument to be explained. Can you check and fix the TSDoc format in other methods too?

src/logging/LoggingService.ts
mustafaiman: 👍 to sancar

src/logging/LoggingService.ts
```diff
@@ -14,77 +14,38 @@
  * limitations under the License.
  */
 
+import {Property} from '../config/Properties';
 import {DefaultLogger} from './DefaultLogger';
-import {NoLogger} from './NoLogger';
+import {ILogger} from './ILogger';
 
 export enum LogLevel {
+    OFF = -1,
```
mustafaiman: `OFF` is not a LogLevel. It is rather a state of logging.

src/logging/LoggingService.ts
```diff
@@ -14,77 +14,38 @@
...
-    static initialize(loggerModule: string | ILogger = null): void {
-        if (typeof loggerModule === 'string') {
-            if (loggerModule === 'off') {
-                LoggingService.loggingService = new LoggingService(new NoLogger());
-            } else if (loggerModule === 'default') {
-                LoggingService.loggingService = new LoggingService();
-            } else {
-                throw new RangeError('Logging type unknown: ' + loggerModule);
-            }
+    constructor(loggingProperty: Property, logLevel: number) {
```
mustafaiman: I see the option of completely removing the logger is done via `-1` logging level now. With this method, everytime a log call is made, it's log level will be compared to `-1` and the log function will not do anything because the comparison is false. However, we do an unneccessary comparison at every log call. The purpose of `NoLogger` was to suppress this unneccessary comparison. I think we should not remove `NoLogger`. We can check the configured log level once the LoggingService is first initialized and we can start a NoLogger in case of -1.

src/nearcache/MetadataFetcher.ts
```diff
@@ -20,16 +20,16 @@ import {MemberSelectors} from '../core/MemberSelectors';
...
 import {RepairingHandler} from './RepairingHandler';
 import ClientMessage = require('../ClientMessage');
+import {ILogger} from '../logging/ILogger';
 
 export class MetadataFetcher {
 
     private client: HazelcastClient;
     private partitionService: PartitionService;
-    private logger: LoggingService = LoggingService.getLoggingService();
+    private logger: ILogger;
```
mustafaiman: I dont see logger being initialized anywhere.

src/proxy/BaseProxy.ts
```diff
@@ -26,7 +26,7 @@ import ClientMessage = require('../ClientMessage');
  */
 export class BaseProxy {
 
-    protected client: HazelcastClient;
+    public client: HazelcastClient;
```
mustafaiman: Why do we need to access client out of a proxy object publicly?

src/logging/LoggingService.ts
furkansenharputlu: @sancar wanted me to make this change. Also, it is used as log level in other log libraries. See: https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm

src/logging/LoggingService.ts
furkansenharputlu: You are right about that it makes extra comparison but I think it is not a huge thing. Also, it provides consistency and simplicity. Therefore, I am in the favor of removing `NoLogger`.

src/proxy/BaseProxy.ts
furkansenharputlu: To access in `ReliableTopicListenerRunner`:
` this.logger = this.proxy.client.getLoggingService().getLogger();`

src/logging/LoggingService.ts
mustafaiman: It is an optimization. It is this PR making the change. Therefore, the PR needs justification not the current state of the code.

furkansenharputlu: In the first implementation, it wasn't a log level. It was a logger like default. However, now, it is a log level and there is no difference from others. I think it can justify why it is removed.

mustafaiman:
verify

PR #429, opened 2018-11-30
Use mocha-junit-reporter

olukas:
Any test failures in our NodeJS Jenkins jobs runs currently cause that that run is finished as failure (red ball). In correct behavior it should rather finish as unstable (yellow ball). 
It is possible by using `mocha-junit-reporter` reporter instead of `spec` reporter in `coverage` script in `package.json` and generate some report file (`report.xml` in this case). Then `Publish JUnit test result report` can be enabled in those Jenkins jobs.


mustafaiman:
@olukas can we use the same for PR builder too? PR builder uses `npm run test` not `coverage`. Can you also intentionally break one of the tests for the first run so we could see how it looks? (I want to make sure the reporter preserves the logs correctly in case of failure.

olukas:
@mustafaiman I broke 1 test intentionally when I implemented it. See https://hazelcast-l337.ci.cloudbees.com/job/NodeJS-10-temp/ - run 17 shows how it looks when there is no failure (green ball), run 16 show how it looks if there is 1 test failure (yellow ball), run 14 shows non-test error (red ball).

Similar jobs for windows is https://hazelcast-l337.ci.cloudbees.com/job/NodeJS-10-windows-temp/ (there is interesting run 23 which shows that red ball is reported when no junit report file is found)

I believe it can be implemented for `npm run test` as well, should I implement it as part of this PR?

mustafaiman:
@olukas Thanks for the links to demos. I think we can merge this and do the same for PR builder later in another PR 👍

olukas:
I added support for `npm run test` as part of this PR. Once this will be merged I will update all related Jenkins jobs. In execute shell:
* `npm run coverage` must be changed to `npm run coverage || true`
* `npm run test` must be changed to `npm run test || true`

And Post-build Action `Publish JUnit test result report` must be added with Test report XMLs `report.xml`.

Issue #435, opened 2018-12-12
NodeJS-8 Statistics test


mustafaiman:
https://hazelcast-l337.ci.cloudbees.com/view/Clients/job/NodeJS-8/587/console
```
19:00:05   1) Statistics with default period should contain statistics content:
19:00:05      AssertionError: expected null not to be null
19:00:05       at test/statistics/StatisticsTest.js:74:36
19:00:05       at <anonymous>
19:00:05 
19:00:05   2) Statistics with default period should contain near cache statistics content:
19:00:05      TypeError: Cannot read property 'indexOf' of null
19:00:05       at contains (test/statistics/StatisticsTest.js:122:30)
19:00:05       at test/statistics/StatisticsTest.js:114:20
19:00:05       at tryCatcher (node_modules/bluebird/js/release/util.js:16:23)
19:00:05       at Promise._settlePromiseFromHandler (node_modules/bluebird/js/release/promise.js:512:31)
19:00:05       at Promise._settlePromise (node_modules/bluebird/js/release/promise.js:569:18)
19:00:05       at Promise._settlePromise0 (node_modules/bluebird/js/release/promise.js:614:10)
19:00:05       at Promise._settlePromises (node_modules/bluebird/js/release/promise.js:693:18)
19:00:05       at Async._drainQueue (node_modules/bluebird/js/release/async.js:133:16)
19:00:05       at Async._drainQueues (node_modules/bluebird/js/release/async.js:143:10)
19:00:05       at Immediate.Async.drainQueues (node_modules/bluebird/js/release/async.js:17:14)
```

PR #438, opened 2018-12-12
Fix for statistics test failure

mustafaiman:
StatisticsTest gets client statistics of the first client from cluster. The assumption that the server must have only 1 client should be true in isolation but when there are hanging clients from other tests, these may authenticate to the cluster as well. In that case, server may send the statistics of a wrong client which is null mostly.

Adds logging to partition table request failure
Fixes hanging clients in LoggingTest, FactoriesTest, SemaphoreProxyTest, CustomSerializerTest
Various test refactors

Also, StatisticsTests is refactored to ask for a specific client's statistics.

fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/435


mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

mustafaiman:
verify

src/invocation/ConnectionAuthenticator.ts
```diff
@@ -53,7 +53,6 @@ export class ConnectionAuthenticator {
                 const authResponse = ClientAuthenticationCodec.decodeResponse(msg);
                 switch (authResponse.status) {
                     case AuthenticationStatus.AUTHENTICATED:
-                        this.connection.setAddress(authResponse.address);
```
sancar: as far as I understand this line is the fix. Actually, there are deeper problems related to mixed usages, like (localhost <-> 127.0.0.1)
We have analyzed the problem in more detail here. 
https://hazelcast.atlassian.net/wiki/spaces/EN/pages/1199407524/Connection+Strategies+-+Blue+Green+by+way+of+CNAME+Configuration+Design#ConnectionStrategies-Blue/GreenbywayofCNAMEConfigurationDesign-Membersandclientswithdifferentinterfacesforsameaddress
I am preparing an extensive fix https://github.com/hazelcast/hazelcast/pull/14104

But currently, we are not supporting mixed usages. It is users responsibility to configure members and clients consistently. I think instead of making this fix, we need to fix the test for now. 
We will work on mixed scenarios as a team and decide on behavior together. Before that, clients should behave the same.

mustafaiman:
@sancar can I get another review? Turns out the issue was hanging clients from other tests. Kudos to @SaitTalhaNisanci for finding the actual scenario.

test/statistics/StatisticsTest.js
```diff
@@ -190,36 +190,41 @@ describe('Statistics with negative period', function () {
...
 });
 
-function getClientStatisticsFromServer(cluster) {
-    var deferred = DeferredPromise();
-    var script = 'client0=instance_0.getClientService().getConnectedClients().' +
-        'toArray()[0]\nresult=client0.getClientStatistics();';
-    RC.executeOnController(cluster.id, script, 1).then(function (response) {
+function getClientStatisticsFromServer(cluster, client) {
+    var clientUuid = client.getClusterService().uuid;
+    console.log(clientUuid);
```
SaitTalhaNisanci: Do you want to keep this here?

mustafaiman: leftover, good catch

mustafaiman: This PR needs to be merged assap because all builds constantly fail. I'll merge it this way and remove the leftover in another PR.

PR #443, opened 2018-12-14
marks some tests with minimum required server versions

mustafaiman:
also increases timeout for some server start heavy tests. Compatibility job fails because of these short timeouts. That job probably runs on a weaker computer.

### DIFF
--- a/test/LostConnectionTest.js
+++ b/test/LostConnectionTest.js
@@ -47,7 +47,7 @@ describe('Lost connection', function () {
     });
 
     it('M2 starts, M1 goes down, client sets M2 as owner', function (done) {
-        this.timeout(12000);
+        this.timeout(32000);
         var newMember;
         var membershipListener = {
             memberAdded: function (membershipEvent) {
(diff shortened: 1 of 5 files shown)


test/LostConnectionTest.js
```diff
@@ -47,7 +47,7 @@ describe('Lost connection', function () {
     });
 
     it('M2 starts, M1 goes down, client sets M2 as owner', function (done) {
-        this.timeout(12000);
+        this.timeout(32000);
```
SaitTalhaNisanci: why dont we have a global timeout ? Like 1 minute maybe and use it everywhere? These timeouts are hard to justify in my opinion. They seem random values.

Most of the time if a test fails in x seconds it will also fail in 2x seconds too.

mustafaiman: At first I was doing this because default mocha timeout was too short for only a handful of tests we had and majority of tests passing fine with the default timeout. Also, this way we would now if a supposedly short term test took a lot of time (e.g. simple map.put taking a few seconds). However, I agree specifying timeout for individual tests is causing more pain than it solves. So I guess we can put a global large timeout. Can you create an issue for that?

SaitTalhaNisanci: I created the issue #452.

mustafaiman:
thanks for the reviews!

PR #444, opened 2018-12-17
Client stats code sample

mustafaiman:


code_samples/client_statistics.js
```diff
@@ -0,0 +1,29 @@
...
+
+function createConfig() {
+    var cfg = new ClientConfig();
+    cfg.properties['hazelcast.client.statistics.enabled'] = true;
+    cfg.properties['hazelcast.client.statistics.period.seconds'] = 2;
+    return cfg;
+}
+
+Client.newHazelcastClient(createConfig()).then(function (client) {
+    client.shutdown();
```
ihsandemir: you may want to let the client sleep or wait here until the statistics is populated and tell the user to go an check at management center for the stats. (Check example https://github.com/hazelcast/hazelcast-cpp-client/blob/master/examples/client-statistics/ClientStatistics.cpp)

mustafaiman: Thanks for the good sample. I updated node.js sample to do similar actions. Can you take another look @ihsandemir ?

degerhz:
#370

Issue #347, opened 2018-09-06
Failure "Heartbeat Heartbeat stopped fired when second member stops heartbeating" on Windows


olukas:
```
Heartbeat Heartbeat stopped fired when second member stops heartbeating:
Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

This error has been seen on testsuite run with node.js 8 executed on Windows machine, see:
https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/NodeJS-8-windows/22/consoleFull

mustafaiman:
This was fixed by https://github.com/hazelcast/hazelcast-nodejs-client/pull/450

PR #450, opened 2018-12-31
Fixes client not sending pings when server continuosly pushes messages

### LINKED ISSUE #336: Client Connection cannot detect idle connection to send ping and server shutdown connection
Client connection should detect that a connection is reading but not sending any data to server to detect heartbeat idle case.
Current implementation checks only last read time which is not enough for reading only connections like just simple listener use case.

Issue from java client:  https://github.com/hazelcast/hazelcast/issues/13576
Fix for java client       :  https://github.com/hazelcast/hazelcast/pull/13577

mustafaiman:
Fixes #336

### DIFF
--- a/test/heartbeat/HeartbeatFromClientTest.js
+++ b/test/heartbeat/HeartbeatFromClientTest.js
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var RC = require('../RC');
+var HazelcastClient = require('../../').Client;
+var expect = require('chai').expect;
+var Config = require('../../').Config;
+var Util = require('../Util');
+var fs = require('fs');
+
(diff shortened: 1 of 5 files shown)


test/heartbeat/HeartbeatFromServerTest.js
```diff
@@ -63,7 +65,9 @@ describe('Heartbeat', function () {
         }).then(function () {
             return RC.startMember(cluster.id);
         }).then(function (member2) {
-            simulateHeartbeatLost(client, member2.host + ':' + member2.port, 2000);
+            return memberAddedPromise.promise.then(function () {
```
mustafaiman: fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/347

test/heartbeat/HeartbeatFromClientTest.js
```diff
@@ -0,0 +1,92 @@
...
+        return RC.createCluster(null, serverConfig).then(function (resp) {
+            cluster = resp;
+        });
+    });
+
+    afterEach(function () {
+        return RC.shutdownCluster(cluster.id);
+    });
+
+    it('client sends heartbeat periodically even when server continuously pushing messages', function () {
```
SaitTalhaNisanci: `continuously pushes messages`

test/heartbeat/HeartbeatFromClientTest.js
```diff
@@ -0,0 +1,92 @@
...
+            client1.getConnectionManager().on('connectionClosed', function () {
+                connectionClosedEventCount++;
+            });
+            return HazelcastClient.newHazelcastClient(clientConfig);
+        }).then(function (c) {
+            client2 = c;
+            return client1.getMap(MAP_NAME);
+        }).then(function (m) {
+            mapFromClient1 = m;
+            return mapFromClient1.addEntryListener({
```
SaitTalhaNisanci: if there is no op why is this being added?

mustafaiman: We try to get the server to send messages to this client even though this client does not send any messages to the server. Having registered a listener, server sends updates to client all the time.

PR #456, opened 2019-01-21
update hazelcast version to use the renewed certificates

mdumandag:
Renewed Let's Encrypt certificates are in the 3.11.2-Snapshot jar. This pr updates the hazelcast version used during the tests to get these certificates


mdumandag:
verify

PR #465, opened 2019-02-26
update readme and code samples

mdumandag:
Updated Readme to get rid of the repetitive names. Renamed `hazelcast-client-sample.json` as `hazelcast-client-full.json` and  added missing properties. Also, changed some single quotes to double quotes in the code samples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,14 +79,11 @@
       * [7.5.2.4. Message Listener](#7524-message-listener)
   * [7.6. Distributed Computing](#76-distributed-computing)
     * [7.6.1. Using EntryProcessor](#761-using-entryprocessor)
-     * [Processing Entries](#processing-entries)
   * [7.7. Distributed Query](#77-distributed-query)
     * [7.7.1. How Distributed Query Works](#771-how-distributed-query-works)
       * [7.7.1.1. Employee Map Query Example](#7711-employee-map-query-example)
       * [7.7.1.2. Querying by Combining Predicates with AND, OR, NOT](#7712-querying-by-combining-predicates-with-and-or-not)
       * [7.7.1.3. Querying with SQL](#7713-querying-with-sql)
-         * [Supported SQL Syntax](#supported-sql-syntax)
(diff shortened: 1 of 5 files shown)


mdumandag:
verify

README.md
```diff
@@ -2152,13 +2157,14 @@ topic.addMessageListener(function (message) {
     console.log(message.messageObject);
 });
 
-var engineer = {
-    name: 'Furkan Senharputlu',
-    age: 23,
-    field: 'Computer Engineering',
-    university: 'Bogazici University'
+var movie = {
```
furkansenharputlu: I wrote this documentation, lots of engineers use their names in tests, docs etc. There is no problem with this. Also, this change doesn't add extra thing. Is it a must to use foreign names ? I wanted to use TURKISH names. See hazelcast docs using `ali`, `veli`: https://docs.hazelcast.org/docs/3.11.2/manual/html-single/index.html#using-map-with-java-client. 

The purpose of this PR is not an enhancement or fix or something..

mdumandag: This pr is meant to check the readme and update the inputs used in examples. I mean, there were more than 10 occurrences of `Furkan` so I changed them with the inputs used in other clients (mixture of python and cpp). I didn't mean to disregard your efforts

furkansenharputlu: Don't forget that this is the base documentation that is copy-pasted to all clients and C++ mostly uses Node.js client examples. Also, these examples aren't used in Python too. For example, I couldn't see `Christopher Nolan`. Anyway, I didn't mean my efforts are disregarded. My commits are still there. I just wanted to say there is no problem with TURKISH names.

mdumandag: I also think that there is no problem with the Turkish names. This sample came to mine mind while editing the readme for the reason I wrote above and I just changed it to that.

As long as the readme serves it purpose, I think it is okay. Whether we used Turkish names or English names does not matter

furkansenharputlu: Yes, you're right so there is no need to make this change because the existing one does the job.

PR #467, opened 2019-03-07
Improve simple map benchmark

puzpuzpuz:
* Replaces existing `SimpleMapBenchmark.js` benchmark with two simple parallel ops benchmark.
* Generic benchmark logic was moved into a common module.
* The common benchmark module runs multiple operations in parallel. Thus it loads the client and the server more effectively.
* A couple of runner scripts were introduced.
* `MapRandomOpRunner` runner is using the original random operation generator.
* `MapPutRunner` runner removes randomness factor and dependency on server-side from the benchmark: all operations now are `map.put('foo', 'bar')`.
* Also includes a readme to simplify running benchmarks.
* Tested on node 4 & 10.

Old benchmark showed about 16-17K ops/sec on my laptop, while the new one (namely `MapPutRunner`) shows about 50-60K ops/sec.


mustafaiman:
I liked the new benchmark using the client to its fullest performance. Thanks for noticing the old benchmark was self throttling.

About this new one, is it possible to re-introduce the parameterized put-get-remove options? We can set PUT_PERCENTAGE to 100 to consistently test put throughput. Also, I think having a key space and some randomness makes more sense because we would like to test it against a cluster of multiple instances. Having a single key makes the client use only a single instance. Even with a single instance scenario, we are limiting hazelcast member to utilize only a single partition. Therefore, we remove parallelization from member by using a single key.

Also, I would like to have benchmark code compatible with node.js 4 as long as we support it.

puzpuzpuz:
@mustafaiman 
I could extract generic benchmark logic into a common module and introduce a couple of runner scripts. The first one would be using the original random operation generator, while the second one would be doing `map.put('foo', 'bar')`.
WDYT?

As for the node 4, I'll downgrade syntax to support it.

mustafaiman:
@puzpuzpuz sounds awesome

puzpuzpuz:
@mustafaiman 
I've updated the code according to our discussions. Could review it once again?

PR #471, opened 2019-04-05
log lifecycle events with the info level

### LINKED ISSUE #462: client shutdown logs nothing
If I do not miss something, shutdown is not being logged. I think it should be logged.

mdumandag:
Formerly, LifecycleService was not logging the state changes. This pr fixes that issue and makes the logging behaviour of the LifecycleService same as the Java client by logging state changes with the info log level.

fixes #462


mdumandag:
verify

PR #475, opened 2019-04-24
add server version check to hazelcast json value tests and increase timeout on listener tests for old server versions

mdumandag:
Server version check was missing from the query related HazelcastJsonValue tests.
Also for old server versions (like 3.6) 30 seconds timeout was not enough for some listener reconnect tests. It is increased to 40 seconds. 

Finally, updated HZ version to 3.12.1-SNAPSHOT. On 3.12-SNAPSHOT some SSL tests that shouldn't fail (like mutual authentication tests) was failing (certificates are not updated yet so I expect the default certificate test to fail again but others should pass)


test/hazelcast_json_value/HJVJsonSerializerTest.js
```diff
@@ -130,11 +107,73 @@ describe('HazelcastJsonValue with JsonSerializer', function () {
             expect(value).to.deep.equal(object);
         });
     });
+});
+
+describe('Querying with HazelcastJsonValue', function () {
```
mustafaiman: This is my personal opinion but having multiple describe blocks in a single test file makes it harder to find failed tests in PR builder and nightly runs. I would prefer each describe block have its own ...Test.js file with a similar name to its label.

mdumandag: Separated them

mdumandag:
verify

mdumandag:
verify

mdumandag:
verify

mdumandag:
verify

PR #478, opened 2019-05-10
implement pipelining

mdumandag:
This pr includes documentation, code samples, implementation and unit tests for the pipelining.

Pipelining is implemented differently than other clients due to the nature of the Node.js

Instead of adding async operations to pipeline, users provide a generator-like function called loadGenerator to the pipeline. When called, load generator returns a promise that represents a request. When the generator is exhausted, it returns null.

Also, the pipelining has an extra option in which you can choose to store the results of the operations within the pipeline or not. Depending on this setting, `run()` method will either resolve with `Promise<void>` or `Promise<E[]>`


src/core/Pipelining.ts
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @param depth the maximum number of concurrent calls allowed in this Pipelining.
+     * @param loadGenerator a generator-like function that returns a promise
+     *  or null if the generator is exhausted.
+     * @param storeResults a boolean value indicating that whether the results of the requests
+     *  should be stored in the Pipelining or not. If this value is set to false, it is expected
+     *  from the user to handle the results of the requests within the load generator. It is false
+     *  by default.
+     * @throws {AssertionError} if the depth is smaller than 1.
+     * @throws {AssertionError} if the load generator is null.
```
mustafaiman: I think `RangeError` would be more appropriate for both throws statements. See the definition of RangeError in Nodejs documentation:
```
A subclass of Error that indicates that a provided argument was not within the set or range of acceptable values for a function; whether that is a numeric range, or outside the set of options for a given function parameter.
```

src/core/Pipelining.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * If the Pipelining is constructed to store the results, results will be
+     * returned with order the requests were done when all of the requests
+     * are completed.
+     *
+     * If not, it will return void when all of the requests are completed.
+     *
+     * @throws {Error} if any of the requests fails
```
mustafaiman: Does it throw the related error or just a generic error instance like "Pipeline failed"?

src/core/Pipelining.ts
```diff
@@ -0,0 +1,181 @@
...
+        const promise = this.loadGenerator();
+        if (promise == null) {
+            this.done = true;
+            if (this.resolvingCount === 0) {
+                this.resolve();
+            }
+            return;
+        }
+
+        this.resolvingCount++;
```
mustafaiman: Is this the number of in-flight operations? If so, better name it that way.

src/core/Pipelining.ts
```diff
@@ -0,0 +1,181 @@
...
+            this.done = true;
+            if (this.resolvingCount === 0) {
+                this.resolve(this.results);
+            }
+            return;
+        }
+
+        this.resolvingCount++;
+        promise.then((value) => {
+           this.results[i] = value;
```
mustafaiman: `nextWithResults` and `nextWithoutResults` are the same except this line. Can you reduce code duplication more?

src/core/Pipelining.ts
```diff
@@ -0,0 +1,181 @@
...
+     * If the Pipelining is constructed to store the results, results will be
+     * returned with order the requests were done when all of the requests
+     * are completed.
+     *
+     * If not, it will return void when all of the requests are completed.
+     *
+     * @throws {Error} if any of the requests fails
+     * @returns array of the results or void
+     */
+    run(): Promise<void | E[]> {
```
mustafaiman: What happens if i do
```
let pipe = new Pipelining(...);
pipe.run();
pipe.run();
```
Looks like all internal variables are going to be messed up. We need to prevent concurrent restarts.

test/pipelining/PipeliningTest.js
```diff
@@ -0,0 +1,237 @@
...
+    it('should not do more operations when the load generator is exhausted with the storage of results ', function () {
+        var pipelining = new Pipelining(100, createLoadGenerator(map), true);
+        return pipelining.run().then(function (result) {
+            expect(result).to.deep.equal(expected);
+            return pipelining.run();
+        }).then(function (result) {
+            expect(result).to.deep.equal(expected);
+        })
+    });
+
```
mustafaiman: Can you write a test that actually verifies that "depth" is respected and there is not more than "depth" operations any given time? I know it is not trivial to write something like that.

Maybe we can write a test with a load generator that verifies internal in-flight requests counter of the pipelining object.

puzpuzpuz: A follow-up to this discussion.

I think that a running IMDG member isn't required for most of the tests in this module. As `Pipelining` isn't coupled with other objects in the client library, we can use any generator-like function that would return promises with `Promise.resolve`/`.reject`/`.delay`/etc. With this approach it'll be more or less trivial to implement a test that verifies depth limit guarantees.

So, I suggest to implement a separate suite of "quick" tests for pipelining. WDYT?

README.md
```diff
@@ -3194,6 +3195,63 @@ You can configure eventual consistency with the `ClientConfig.properties` below:
 
 - `hazelcast.invalidation.reconciliation.interval.seconds`: Default value is `60` seconds. This is a periodic task that scans cluster members periodically to compare generated invalidation events with the received ones from the client Near Cache.
 
+### 7.8.3. Pipelining
+
+With the pipelining, you can send multiple concurrent requests without waiting for their results with a concurrency limit.
```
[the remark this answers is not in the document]
mustafaiman: I think @Serdaro should take a look at all Readme change here. I believe some parts of this documentation is the same as the one in reference manual. If there is something to improve, we can improve both

src/core/Pipelining.ts
mustafaiman: I think the definition from node.js manual that I quoted perfectly fits the use case here. It is not limited to number intervals but it explicitly states that `whether that is a numeric range, or outside the set of options for a given function parameter`.

I looked at the article but couldn't find anything about why we shouldn't use builtin Errors. If anything, I am all for using builtin API rather than introducing new.

test/pipelining/PipeliningTest.js
mustafaiman: +1 to testing Pipelining throughly in isolation. I still would like to see one or two tests involving some map operations to make sure they integrate well.

src/core/Pipelining.ts
puzpuzpuz: I don't see how `RangeError` fits into the `if the load generator is null` case. The doc fragment that you have cited says: `a numeric range, or outside the set of options for a given function parameter`. As I see it, an argument being `null` has nothing to do with ranges or set of options, so this error would be misleading. I also tried to find `null` check cases in the Node source code, but a brief check gave me nothing.

Still, I agree that `RangeError` fits both cases much better than `AssertionError` and if we have to choose from standard errors that the one that we should use.

As for using custom error hierarchy, it has the following benefits:
* Client library users will be able to determine whether a thrown error is a library level error or it's a system level one
* Additional types of errors can be easily added, thus allowing to cover cases where standard errors don't fit in.

mdumandag: What about the `TypeError` for the null check ? It is thrown when you try to access the keys of a null object. If that check does not exists, we would get `TypeError` when they call the `run` method. So, can we think this error check as a prevention for a future error and throw `TypeError` before the future error happens? I think it is a similar approach used within the Java community where `NullPointerException` is thrown when a null value is passed in to a parameter when non-null values are expected.

src/core/Pipelining.ts
mdumandag: It throws the first error it encountered. Should we wrap the error in a specific error related to Pipelining and set the error it encountered as the cause ?

src/core/Pipelining.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    private nextWithoutResults(): void {
+        if (this.rejected) {
+            return;
+        }
+
+        const promise = this.loadGenerator();
+        if (promise == null) {
+            this.done = true;
+            if (this.resolvingCount === 0) {
```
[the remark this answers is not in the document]
mdumandag: The `run` method has a break statement before it calls the next `next`. If the pipelining is `done`, it breaks the loop without calling the `next`. So, resolve is called only once in this case.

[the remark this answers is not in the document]
puzpuzpuz: Yes, you're right. I've missed that `break` in the starting loop. Thanks for the clarification!

README.md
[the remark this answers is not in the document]
Serdaro: ToDo`ing this Mustafa.

src/core/Pipelining.ts
puzpuzpuz: > It throws the first error it encountered. Should we return a specific error related to Pipelining and set the error it encountered as the cause ?

As we don't have library-specific error hierarchy, I vote for returning an unwrapped error, as is. But we should mention this in the JSDoc.

src/core/Pipelining.ts
puzpuzpuz: +1 for `TypeError` in the `null` case.

src/core/Pipelining.ts
```diff
@@ -90,11 +83,16 @@ export class Pipelining<E> {
...
      *
-     * @throws {Error} if any of the requests fails
+     * @throws {Error} with the error of the failing request if any of the requests fail. Also,
+     *  subsequent calls to this method with the same pipelining instance rejects.
      * @returns array of the results or void
      */
     run(): Promise<void | E[]> {
-        this.reset();
+        if (this.isPipeliningRan) {
+            return Promise.reject('This pipelining instance was already run. ' +
```
puzpuzpuz: It seems to me like a good idea to mention this behavior (rejection on non-first `run()`) in JSDoc and readme. WDYT?

mdumandag: Thanks for the review. I will clarify this behaviour on documentations

README.md
```diff
@@ -3194,6 +3195,63 @@ You can configure eventual consistency with the `ClientConfig.properties` below:
 
 - `hazelcast.invalidation.reconciliation.interval.seconds`: Default value is `60` seconds. This is a periodic task that scans cluster members periodically to compare generated invalidation events with the received ones from the client Near Cache.
 
+### 7.8.3. Pipelining
+
+With the pipelining, you can send multiple parallel requests with a defined concurrency limit without waiting for their results.
```
Serdaro: No need to use the article ("the") for the act of pipelining throughout this content.

README.md
```diff
@@ -3194,6 +3195,63 @@ You can configure eventual consistency with the `ClientConfig.properties` below:
...
+
+The pipelining allows to control back pressure, i.e., to limit
```
Serdaro: Pipelining allows controlling the back pressure, i.e., limiting ...

README.md
```diff
@@ -3194,6 +3195,63 @@ You can configure eventual consistency with the `ClientConfig.properties` below:
...
+        return pipelining.run();
+    }).then(function (results) {
+        // Wait for completion
+        console.log(results);
+        return client.shutdown();
+    });
+});
+```
+
+In the above example, we make 100 concurrent `map.get()` calls, but the maximum number of inflight calls is 10.
```
Serdaro: The above example makes 100 concurrent ...

README.md
```diff
@@ -3194,6 +3195,63 @@ You can configure eventual consistency with the `ClientConfig.properties` below:
...
+
+By increasing the depth of the pipelining, throughput can be increased.
```
Serdaro: You can increase the throughput by increasing the depth of pipelining.

README.md
```diff
@@ -3194,6 +3195,63 @@ You can configure eventual consistency with the `ClientConfig.properties` below:
...
+        return client.shutdown();
+    });
+});
+```
+
+In the above example, we make 100 concurrent `map.get()` calls, but the maximum number of inflight calls is 10.
+
+By increasing the depth of the pipelining, throughput can be increased. 
+
+The pipelining has its own back pressure, you do not need to enable the back pressure on the member to have this feature on the pipelining.
```
Serdaro: You can simplify the whole sentence as:

Pipelining has its own back pressure; you don't need to enable it on the member.

README.md
```diff
@@ -3194,6 +3195,63 @@ You can configure eventual consistency with the `ClientConfig.properties` below:
...
+    });
+});
+```
+
+In the above example, we make 100 concurrent `map.get()` calls, but the maximum number of inflight calls is 10.
+
+By increasing the depth of the pipelining, throughput can be increased. 
+
+The pipelining has its own back pressure, you do not need to enable the back pressure on the member to have this feature on the pipelining. 
+However, if you have many pipelines, you may still need to enable the member back pressure because it is possible to overwhelm the system with requests in that situation.
```
Serdaro: .... to enable the back pressure on the member since the system may become overwhelmed by the requests in that situation.

README.md
```diff
@@ -3194,6 +3195,63 @@ You can configure eventual consistency with the `ClientConfig.properties` below:
...
+See the [Back Pressure section](https://docs.hazelcast.org/docs/latest/manual/html-single/index.html#back-pressure) of the Hazelcast IMDG Reference Manual to learn how to enable it on the member.
+
+You can use the pipelining without a special configuration, it works out-of-the-box.
```
Serdaro: ... special configuration; it works ...

srknzl:
closing due to being stale, open if you want to continue on this

PR #491, opened 2019-07-10
write byte length of strings instead of char length on the client mes…

### LINKED ISSUE #474: Non-ASCII chars in data structure names lead to invalid names on server
String serialization for data structure names incorrectly places length of the string (in chars) into the first 4 bytes:
https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/ClientMessage.ts#L157

Instead it should use bytes length of the string. See `SafeBuffer#putStringUtf8` in the Java client for the correct implementation.

The easiest way to reproduce it is to try to access the same data structure from Node.js and Java client.

First, run this Node.js code:
```javascript
'use strict';

const Client = require('../.').Client;

Client.newHazelcastClient()
    .then((client) => client.getMap('тестовое-отображение'))
    .then((map) => {
        return map.set('key', 'value')
            .then(() => map.get('key'))
            .then(console.log)
            .then(() => map);
    })
    .then((map) => map.client.shutdown());
```

Now run this Java client:
```java
ClientConfig clientConfig = new ClientConfig();

List<String> memberAddresses = Arrays.asList("127.0.0.1:5701");
clientConfig.getNetworkConfig().setAddresses(memberAddresses);

HazelcastInstance client = HazelcastClient.newHazelcastClient(clientConfig);

IMap map = client.getMap("тестовое-отображение");
System.out.println(map);
// this gives a `0`:
System.out.println(map.size());
// this gives a `null`:
System.out.println(map.get("key"));
```

mdumandag:
write byte length of strings instead of char length on the client message.

This was causing non-ascii strings to have wrong size written to client message and was resulting in the issue described in #474 .

fixes #474


src/codec/Utils.ts
```diff
@@ -23,7 +23,7 @@ export class Utils {
             size += 1;
         }
 
-        size += value == null ? 0 : value.length;
+        size += value == null ? 0 : Buffer.byteLength(value, 'utf8');
```
puzpuzpuz: Is this `Utils` class used anywhere? It duplicates `BitsUtil`'s functionality. Maybe we should remove it?

mdumandag: Thanks for the remark, this file is duplicate and unused as you said. I removed it.

mdumandag:
verify

PR #493, opened 2019-07-19
a sample about querying with SQL

gulcesirvanci:


code_samples/querying_with_SQL.js
```diff
@@ -0,0 +1,49 @@
...
+            ['6', new Customer('Ragnar', true, 30)],
+            ['7', new Customer('Hilary', true, 19)],
+        ]);
+    }).then(function () {
+        const predicate = new Predicates.sql('active AND age < 30');
+        return personMap.valuesWithPredicate(predicate);
+    }).then(function (values) {
+        values.toArray().forEach(function (value) {
+            console.log(value);
+            return client.shutdown();
```
mdumandag: This return statement should be in the outer scope.

code_samples/querying_with_SQL.js
```diff
@@ -0,0 +1,49 @@
...
+var Predicates = require('hazelcast-client').Predicates;
+
+function Customer(name, active, age) {
+    this.active = active;
+    this.age = age;
+    this.name = name;
+};
+
+Client.newHazelcastClient().then(function (hazelcastClient) {
+    var client = hazelcastClient;
```
mdumandag: no need to define a new variable for the client. Delete this and change `function (hazelcastClient)` to `function (client)`

code_samples/querying_with_SQL.js
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+var Client = require('hazelcast-client').Client;
+var Predicates = require('hazelcast-client').Predicates;
+
+function Customer(name, active, age) {
+    this.active = active;
+    this.age = age;
+    this.name = name;
```
mdumandag: It would be better for readability if you assign variables in the same order with the constructor

code_samples/querying_with_SQL.js
```diff
@@ -0,0 +1,49 @@
...
+};
```
mdumandag: unnecessary `;`

mdumandag:
verify

PR #495, opened 2019-07-22
add missing expired and loaded events

### LINKED ISSUE #483: Missing "LOADED" event
See https://github.com/hazelcast/hazelcast-go-client/issues/464

IMDG 3.11 added LOADED event - https://docs.hazelcast.org/docs/3.12.1/javadoc/com/hazelcast/core/EntryEventType.html#LOADED

This is not handled by this client

mdumandag:
This pr adds the missing two events: expired and loaded. The test for the loaded event uses `SampleMapStore` in the `com.hazelcast.client.test` package. It firsts puts a data to map with TTL. After 1 second, data is not available on the map but available on the map store. So, the get method call causes the event to fire by loading the event from the underlying MapStore

fixes #483


mdumandag:
verify

PR #502, opened 2019-08-05
Fixes Client member list state gets broken with hot restart

### LINKED ISSUE #472: Client member list state gets broken with hot restart
This behavior has changed in hazelcast 3.12. For clients to be able to work with 3.12 members and hot-restart feature, the fix should be applied to all clients. 

issue https://github.com/hazelcast/hazelcast/issues/14839
fix https://github.com/hazelcast/hazelcast/pull/14844
test https://github.com/hazelcast/hazelcast/issues/14839

gulcesirvanci:
In hotrestart feature, members will preserve their uuid's, but can
start with different addresses. Client was relying on only uuid,
and when a member restarted, it was assuming nothing has changed.
When it does not change its local membership view in this case, it
is trying to continue with old address which is wrong.

Also for these cases, we were not firing any member
removed,added events. With this fix, for a restarted member
with different address we will wire removed,added events.

Note that, if member comes back with same address and uuid, from
client point of view there is no way to detect the restart. So
no events will be fired in that case.

 Fixes #472

### DIFF
--- a/src/core/MembershipEvent.ts
+++ b/src/core/MembershipEvent.ts
@@ -24,17 +24,17 @@ export class MembershipEvent {
     /**
      * the removed or added member.
      */
-    private member: Member;
+    public member: Member;
 
     /**
      * the membership event type.
      */
-    private eventType: number;
+    public eventType: number;
 
     /**
      * the members at the moment after this event.
      */
-    private members: Member[];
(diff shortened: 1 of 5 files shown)


puzpuzpuz:
Closing this one in favor of #610

PR #526, opened 2020-04-21
Simplify readme and remove package-lock.json

mdumandag:
This PR moves the README.MD to a file named DOCUMENTATION.MD
and simplifies the README.MD by shortening it to display the
intent and capabilities of the client better.

Also, removed the package-lock.json since it is never published to npm and that defeats the purpose of it.

I couldn't add the badges for our private Jenkins build yet due to a problem in my computer. I cannot access the jenkins.hazelcast.com to setup the badges. I will try with different configurations to solve to problem and add the badges afterwards.


README.md
```diff
@@ -1,3470 +1,188 @@
...
-Hazelcast Node.js client performs automated pipelining of operations. It means that the library pushes all operations into an internal queue and tries to send them in batches. This reduces the count of executed `Socket.write()` calls and significantly improves throughtput for read operations.
+### Usage
 
-You can configure automated operation pipelining with the `ClientConfig.properties` below:
+```js
+const Client = require('hazelcast-client').Client;
 
-- `hazelcast.client.autopipelining.enabled`: Default value is `true`. Turns automated pipelining feature on/off. If your application does only writes operations, like `IMap.set()`, you can try disabling automated pipelining to get a slightly better throughtput.
+// Connect to Hazelcast cluster
+Client.newHazelcastClient()
```
puzpuzpuz, line 62: It's better to use async/await syntax here for the sake of readability.

The same suggestion applies to the config snippet.

README.md
```diff
@@ -1,3470 +1,188 @@
...
-    },
-
-    warn: function (objectName, message, furtherInfo) {
-        this.log(LogLevel.WARN, objectName, message, furtherInfo);
-    },
 
-    info: function (objectName, message, furtherInfo) {
-        this.log(LogLevel.INFO, objectName, message, furtherInfo);
-    },
+You can also configure the client
```
puzpuzpuz, line 114: I'd also leave a link to the full doc's config section somewhere after this paragraph.

README.md
```diff
@@ -1,3470 +1,188 @@
...
+[declaratively](https://github.com/hazelcast/hazelcast-nodejs-client#312-declarative-configuration-json) using a JSON file.
```
puzpuzpuz, line 115: I don't think `https://github.com/hazelcast/hazelcast-nodejs-client#312-declarative-configuration-json` link will work once this PR is merged.

README.md
```diff
@@ -1,3470 +1,188 @@
...
-    trace: function (objectName, message, furtherInfo) {
-        this.log(LogLevel.TRACE, objectName, message, furtherInfo);
-    }
+* Distributed, partitioned and queryable in-memory key-value store implementation, called **Map**
+* Eventually consistent cache implementation to store a subset of the Map data locally in the memory of the client, called **Near Cache**
+* Additional data structures and simple messaging constructs such as **Set**, **MultiMap**, **Queue**, **Topic**
+* Cluster-wide unique ID generator, called **FlakeIdGenerator**
+* Distributed, CRDT based counter, called **PNCounter**
+* Primitives for distributed computing such as **Lock**, **Semaphore**, **Atomic Long**
+* Integration with [Hazelcast.cloud](https://cloud.hazelcast.com/)
```
puzpuzpuz, line 125: Don't we call it `Hazelcast Cloud` (not `Hazelcast.cloud`)?

README.md
```diff
@@ -1,3470 +1,188 @@
...
-    info: function (objectName, message, furtherInfo) {
-        this.log(LogLevel.INFO, objectName, message, furtherInfo);
-    },
+You can also configure the client
+[declaratively](https://github.com/hazelcast/hazelcast-nodejs-client#312-declarative-configuration-json) using a JSON file.
 
-    debug: function (objectName, message, furtherInfo) {
-        this.log(LogLevel.DEBUG, objectName, message, furtherInfo);
-    },
+## Features
```
puzpuzpuz, line 117: I'd put a link to https://hazelcast.org/imdg/clients-languages/node-js/ at the end of this section, as this page has a list of supported features.

puzpuzpuz, line 117: Users seem to be interested in various events exposed via client API, so I'd also mention this one in the features list

README.md
```diff
@@ -1,3470 +1,188 @@
...
-Note that it is not possible to configure custom logging via declarative configuration.
+## Getting Help
 
-# 8. Development and Testing
+You can use the following channels for your questions and development/usage issues:
 
-Hazelcast Node.js client is developed using TypeScript. If you want to help with bug fixes, develop new features or
-tweak the implementation to your application's needs, you can follow the steps in this section.
+* [GitHub repository](https://github.com/hazelcast/hazelcast-nodejs-client)
+* [Complete documentation](https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/README.md)
```
puzpuzpuz, line 134: Is it the correct link? Should be something like `*/COMPLETE_DOCUMENTATION.md`

COMPLETE_DOCUMENTATION.md
```diff
@@ -0,0 +1,3471 @@
...
+        this.log(LogLevel.TRACE, objectName, message, furtherInfo);
+    }
+
+};
+cfg.customLogger = winstonAdapter;
+```
+
+Note that it is not possible to configure custom logging via declarative configuration.
+
+# 8. Development and Testing
```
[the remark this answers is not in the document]
mdumandag: The development section in the DOCUMENTATION.md talks about linking the client package so that it can be used in other projects. Do you think that we should put these changes into README before removing them ?

README.md
mdumandag: Should we put the code samples into an async function or should they be in the top-level for demonstration purposes ?

README.md
mdumandag: The configuration section does not offer more for the programmatic configuration than README. It contains somewhat advanced usages for the declarative configuration. Should we do it ?

README.md
mdumandag: We were calling it Hazelcast.cloud while implementing the discovery feature but it is called as Hazelcast Cloud in other places now. Updated it

README.md
mdumandag: I changed the link of `many more` with the features list in this page. Would that be enough ?

COMPLETE_DOCUMENTATION.md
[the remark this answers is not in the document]
puzpuzpuz: I think we can remove `npm link` part completely, as it's not that important.

README.md
puzpuzpuz: I'd prefer top-level as it'll make snippets more readable. Any reader who is familiar with async/await knows that they need to wrap the snippet into an async function.

README.md
puzpuzpuz: Not sure if I understand the question correctly, but I'd leave a link to the following section: https://github.com/hazelcast/hazelcast-nodejs-client#3-configuration-overview

README.md
puzpuzpuz: Sounds good

README.md
mdumandag: I mean in this section, we talk about programmatic configuration as much as we do in the new README. I don't think it offers more for the reader for this configuration type.

For the declarative configuration, the part I linked summarizes the concept well and the remaining parts (which are needed just for the declarative configuration) are easily reachable from there if they are needed.

puzpuzpuz: Got it. In this case, we don't need a link here.

package-lock.json
```diff
@@ -1,2650 +1,2644 @@
...
-            "integrity": "sha1-D2RYS6ZC8PKs4oBiefTwbKI61zw=",
-            "dev": true,
-            "requires": {
-                "assertion-error": "^1.0.1",
-                "check-error": "^1.0.1",
-                "deep-eql": "^3.0.0",
-                "get-func-name": "^2.0.0",
-                "pathval": "^1.0.0",
-                "type-detect": "^4.0.0"
+  "name": "hazelcast-client",
```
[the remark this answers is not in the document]
puzpuzpuz: @mdumandag could you mention this change in the PR description, so there would be a quick way to understand the changes if we ever need to return to this PR?

srknzl:
> Also, removed the package-lock.json since it is never published to npm and that defeats the purpose of it.

Is this really the case btw? I mean according to https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json it is not

puzpuzpuz:
> Is this really the case btw? I mean according to https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json it is not

`package-lock.json` shouldn't be included into libraries, as npm ignores `package-lock.json` from `./node_modules` (at least, it used to do that; but as we don't control Node.js/npm version used by the end user, it's the same for us).

srknzl:
Yeah but keeping it will help us as developers to use same dependencies when we debug, test etc. It is of course not useful for the library users

puzpuzpuz:
The thing is that we should avoid keeping it even for ourselves (i.e. in `.gitignore`). Otherwise, each CI run will use the same set of dependency versions, while the end user might get a different version of one of our transient dependencies which may result in a broken functionality. It's a common advice to avoid package-lock in libraries and most (if not all of) popular Node.js libraries follow it.

srknzl:
Ok can we have this kind of functionality:

fixed versions for devDependencies 
non-fixed versions for dependencies 

see my comment https://github.com/hazelcast/hazelcast-nodejs-client/pull/825#issuecomment-803878371

puzpuzpuz:
> Best of both worlds:
> 
> fixed versions for devDependencies
> non-fixed versions for dependencies

Does npm allow you to generate `package-lock.json` for dev dependencies only? If it does, then we need to somehow make sure to restrict the lock file to contain dev dependencies only.

srknzl:
Yeah I will look for that option it would be better than not having one

srknzl:
Could not find the option. What about running the ci with `--no-package-lock` and including the lock file in git. Ci will ignore package-lock.json file. And we will be sure that we work with same dependencies as developers(especially the devDependencies are important, like eslint, typescript etc)

cc @mdumandag

puzpuzpuz:
> devDependencies are important, like eslint, typescript etc

If we want to avoid problems related with dev dependencies, it should be enough to freeze their minor version in the `package.json` (`~x.y.z` syntax). Popular libraries follow semver and in rare cases when they don't, it's unlikely to happen in a patch release.

PR #528, opened 2020-04-30
Add RestValue serializer

### LINKED ISSUE #518: Hazelcast queue - Write in queue with REST CLIENT and poll entry with NodeJs client
Moving https://github.com/hazelcast/hazelcast/issues/16713 by @NutellaTorte to here.
__

  
**Describe the bug**
Hello,
I want to write entries to a queue using REST call. This is no problem, this works fine.
Next step is to create a item listener on this queue using the NodeJs client. This item listener should poll the entries which I added by the REST call. If I try to start the NodeJs client, there will be an error: 

RangeError: There is no Identified Data Serializer factory with id -37.

I read a lot of your documentation and tried to create a custom serializer, identified data serializer. Nothing works. I dont know what to do next.

**Expected behavior**
I expect the NodeJs Client to receive the JSON string that was written to the queue via the REST call.

**To Reproduce**

1. Curl 
```
 curl -v -H "Content-Type: text/plain" -d '{"name":"example"}' http://<host>/hazelcast/rest/queues/myqueue
```
2. NodeJs Files
index.ts
```
import { HazelcastClient } from './hazelcast';

const client = new HazelcastClient();
client.startClient().then(client => {
    client.getQueue('myqueue').then(queue => {
        const itemListener = {
            itemAdded: (itemEvent) => {
                console.log('item added', itemEvent.item);
            } ,
            itemRemoved: (itemEvent) => {
                console.log('item removed', itemEvent.item);
            }
        };
        queue.addItemListener(itemListener, true).then(value => {
            console.log(value);
        });
    });
}); 
```
hazelcast.ts
```
import { Client, Config } from 'hazelcast-client';
export class HazelcastClient {
    config = new Config.ClientConfig();

    constructor() {
        this.config.networkConfig.addresses.push(
               'myaddresses:5701
        );
    }

    startClient() {
        return Client.newHazelcastClient(this.config);
    }
}
```
3. Error Message: 
```
RangeError: There is no Identified Data Serializer factory with id -37.
```

**Additional context**
```
"dependencies": {
    "@types/bluebird": "^3.5.29",
    "@types/long": "^4.0.1",
    "@types/node": "^13.7.0",
    "hazelcast-client": "^3.12.2",
    "typescript": "^3.7.5"
}
``` 
Hazelcast Version: 3.12.7

mdumandag:
This PR adds the RestValue serializer to be able deserialize data put over the REST endpoints.

fixes #518

### DIFF
--- a/src/core/RestValue.ts
+++ b/src/core/RestValue.ts
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2008-2020, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {IdentifiedDataSerializable, IdentifiedDataSerializableFactory} from '../serialization/Serializable';
+import {DataInput, DataOutput} from '../serialization/Data';
+
+export const REST_VALUE_FACTORY_ID = -37;
+export const REST_VALUE_CLASS_ID = 1;
+
+export class RestValue implements IdentifiedDataSerializable {
+    value: number[];
+    contentType: number[];
+
+    getClassId(): number {
+        return REST_VALUE_CLASS_ID;
+    }
+
+    getFactoryId(): number {
+        return REST_VALUE_FACTORY_ID;
+    }
+
+    readData(input: DataInput): any {
+        this.value = input.readByteArray();
+        this.contentType = input.readByteArray();
+    }
+
+    writeData(output: DataOutput): void {
+        output.writeByteArray(this.value);
+        output.writeByteArray(this.contentType);
+    }
+}
+
+export class RestValueFactory implements IdentifiedDataSerializableFactory {
+    create(type: number): IdentifiedDataSerializable {
+        if (type === REST_VALUE_CLASS_ID) {
+            return new RestValue();
+        }
+        return null;
+    }
+}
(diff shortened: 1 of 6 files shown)


src/core/RestValue.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+import {IdentifiedDataSerializable, IdentifiedDataSerializableFactory} from '../serialization/Serializable';
+import {DataInput, DataOutput} from '../serialization/Data';
+
+export const REST_VALUE_FACTORY_ID = -37;
+export const REST_VALUE_CLASS_ID = 1;
+
+export class RestValue implements IdentifiedDataSerializable {
+    value: number[];
```
puzpuzpuz, line 24: For both `value` and `contentType` fields, it's better to use native type for binary data (`Buffer`) instead of an array of numbers. But in this case we could also use strings instead of numbers, as they come from HTTP. It's not the way we do it in Java client, but over there it's used also in REST API, so that should be the reason why it operates binary data instead of strings. WDYT?

BTW we have a PRD to improve the client in this way:
https://hazelcast.atlassian.net/wiki/spaces/PM/pages/1749254626/Node.js+Client+Support+native+binary+data+type

@burakcelebi maybe it makes sense to include the above PRD into the 4.0 plan, as it assumes backwards incompatible changes. WDYT?

test/serialization/DefaultSerializersLiveTest.js
```diff
@@ -160,5 +162,30 @@ stringSerializationPolicies.forEach(function(stringSerializationPolicy) {
...
             });
         });
+
+        it('rest value', function () {
+            // Making sure that the object is properly de-serialized at the server
+            var restValue = new RestValue();
+            restValue.value = Buffer.from('{"test":"data"}').toJSON().data;
+            restValue.contentType = Buffer.from('text/plain').toJSON().data;
+
+            var script = 'var map = instance_0.getMap("' + map.getName() + '");\n' +
```
puzpuzpuz, line 172: nit: if we want to make sure that the stored `RestValue` is correctly deserialized on member side, wouldn't it be simpler to read the value via IMDG REST API after we put it in the map?

src/core/RestValue.ts
mdumandag: Changed field types to string

test/serialization/DefaultSerializersLiveTest.js
mdumandag: Yes thats also an option, but I found it simpler to write using remote controller script. I could change that if you want

mdumandag:
verify

test/serialization/DefaultSerializersLiveTest.js
puzpuzpuz: In my opinion, an HTTP request instead of a script will make this test more readable. But that's a matter of taste, so it's up to you to change it or not.

PR #533, opened 2020-05-15
Remove safe buffer dependency

### LINKED ISSUE #519: Change minimal Node.js version to v8.0 and remove safe-buffer dependency
Node.js v4.0-7.0 are not supported for a long time (see https://nodejs.org/en/about/releases/), so we should change minimal Node.js version to v8.0 as a part of 4.0 release.

Once we do it, we can remove `safe-buffer` library from deps and use `Buffer` API directly.

mdumandag:
We are going to pick Node.js 8 as the minimum supported version,
which contains polyfill methods implemented by the safe-buffer.
Hence, there is no need for an extra dependency for these.

Protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/321

fixes #519

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,7 @@
   "main": "lib/index.js",
   "dependencies": {
     "bluebird": "3.5.4",
-    "long": "3.2.0",
-    "safe-buffer": "5.1.2"
+    "long": "3.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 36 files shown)


puzpuzpuz:
Note: this PR seems to close #519

PR #536, opened 2020-05-26
Add missing properties and schemas

mdumandag:
There were a few missing properties in the hazelcast-client-default.json
along with some non-valid values. Also, hazelcastCloud element was
missing from the configuration validation schema.


mdumandag:
verify

PR #546, opened 2020-06-22
Fix ListenerServiceTest

mdumandag:
This test fails time to time. I fixed the issue with this test in
master branch. This PR backports the changes done in master to
3.12.x branch.

The problem is the following. In tests, after asserting the
values received from the events, we remove the distributed object
listener and call done(). However, done is called in the next
statment instead of a then block. Because of that, sometimes,
the listeners from the previous test runs stay active in other tests,
and that causes the failures. I believe that we should not use the 
same client for every test. Each of them tests a separate logic. 
Therefore, in this fix I changed the tests so that each of them uses 
a separate client.

Also, updated the scripts with the recent changes from master. There were
small improvments to some internal logic.

### DIFF
--- a/scripts/dev-test-rc.sh
+++ b/scripts/dev-test-rc.sh
@@ -18,6 +18,15 @@ else
 	ENTERPRISE_REPO=${ENTERPRISE_RELEASE_REPO}
 fi
 
+if [[ ${HZ_TEST_VERSION} == *-SNAPSHOT ]]
+then
+    TEST_REPO=${SNAPSHOT_REPO}
+    ENTRERPRISE_TEST_REPO=${ENTERPRISE_SNAPSHOT_REPO}
+else
+    TEST_REPO=${RELEASE_REPO}
+    ENTRERPRISE_TEST_REPO=${ENTERPRISE_RELEASE_REPO}
+fi
+
 if [ -f "hazelcast-remote-controller-${HAZELCAST_RC_VERSION}.jar" ]; then
     echo "remote controller already exist, not downloading from maven."
 else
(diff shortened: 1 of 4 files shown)


mdumandag:
verify

scripts/dev-test-rc.sh
```diff
@@ -18,6 +18,15 @@ else
 	ENTERPRISE_REPO=${ENTERPRISE_RELEASE_REPO}
 fi
 
+if [[ ${HZ_TEST_VERSION} == *-SNAPSHOT ]]
+then
+    TEST_REPO=${SNAPSHOT_REPO}
+    ENTRERPRISE_TEST_REPO=${ENTERPRISE_SNAPSHOT_REPO}
+else
+    TEST_REPO=${RELEASE_REPO}
+    ENTRERPRISE_TEST_REPO=${ENTERPRISE_RELEASE_REPO}
```
puzpuzpuz, line 27: nit: misprint in `ENTRERPRISE_TEST_REPO`. Should be `ENTERPRISE_TEST_REPO`

scripts/download-rc.sh
```diff
@@ -69,7 +69,7 @@ if [ -n "${HAZELCAST_ENTERPRISE_KEY}" ]; then
         echo "hazelcast-enterprise-tests.jar already exists, not downloading from maven."
     else
         echo "Downloading: hazelcast enterprise test jar com.hazelcast:hazelcast-enterprise:${HAZELCAST_TEST_VERSION}:jar:tests"
-        mvn -q org.apache.maven.plugins:maven-dependency-plugin:2.8:get -DrepoUrl=${ENTERPRISE_REPO} -Dartifact=com.hazelcast:hazelcast-enterprise:${HAZELCAST_TEST_VERSION}:jar:tests -Ddest=hazelcast-enterprise-${HAZELCAST_TEST_VERSION}-tests.jar
+        mvn -q org.apache.maven.plugins:maven-dependency-plugin:2.8:get -DrepoUrl=${ENTRERPRISE_TEST_REPO} -Dartifact=com.hazelcast:hazelcast-enterprise:${HAZELCAST_TEST_VERSION}:jar:tests -Ddest=hazelcast-enterprise-${HAZELCAST_TEST_VERSION}-tests.jar
```
puzpuzpuz, line 72: nit: the same misprint in `ENTRERPRISE_TEST_REPO` here

scripts/dev-test-rc.sh
mdumandag: Thanks for that, I forget to check these two scripts for this typo initially

mdumandag:
verify

PR #549, opened 2020-06-26
Support protocol extension

mdumandag:
This PR adds support for protocol extension for the Node.js client. 

Main changes are in the protocol repo for now since we don't have that many additions to protocol yet. There is,

- https://github.com/hazelcast/hazelcast-client-protocol/pull/312 which is request only. So, no changes are needed for non-java protocol-wise.
- https://github.com/hazelcast/hazelcast-client-protocol/pull/313 all params has `2.1` since values so protocol-wise, there should be no changes.
- https://github.com/hazelcast/hazelcast-client-protocol/pull/316 which is specific to Java client. We don't have advanced network configuration. Hence this is not needed. For now, I added simple `!skip` type to the protocol template of Node.js client that we can use to skip encoding/decoding parameters of this types. With this approach, there is a problem if another var-sized type is added to protocol definition after the skipped parameters but I believe it is okay for now. It feels wrong to add `EndpointQualifier` class to client just because protocol requires it. We can discuss whether we should do workarounds or add `EndpointQualifier`.

However, considering future additions to protocol, I changed to way we decode responses. If there is an addition that requires conditional checks (like https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/codec/MapPutAllCodec.java#L101) we cannot return the way we are using now. So, it is changed for all codecs.

This PR also contains implementation for `map#setAll` and `multimap#putAll` but I am not sure about them. Looking at the codecs, there are lots of unimplemented features. Since there is no way to tell protocol code generator to skip certain operations of a certain service (like skip `fetchKeys` of `Map.yaml`) they are auto-generated like these two. Maybe we should create an issue to track all of them and create PRs for them separately.

Protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/327


puzpuzpuz:
> Maybe we should create an issue to track all of them and create PRs for them separately.

Sounds like a good idea to me.

puzpuzpuz:
> It feels wrong to add `EndpointQualifier` class to client just because protocol requires it. We can discuss whether we should do workarounds or add `EndpointQualifier`.

I'm ok with skipping this parameter, however I think that we should at least create a GH issue to track this workaround. The issue could describe advanced network configuration support in Node.js client.

test/map/MapProxyTest.js
```diff
@@ -202,36 +202,44 @@ describe('MapProxy', function () {
...
+                    ];
+                    var returnedCorrectly = 0;
+                    var verify = function (expected) {
+                        return function (val) {
+                            try {
+                                expect(val).to.equal(expected);
+                                returnedCorrectly++;
+                                if (returnedCorrectly === 5) {
+                                    done();
+
```
puzpuzpuz, line 222: nit: redundant new line here.

mdumandag:
@puzpuzpuz While creating issues I just saw that we are generating codecs for `Transacion*` services which is not supported by this client. We can skip generating them since they are defined as different services. So, I also did this

puzpuzpuz:
> @puzpuzpuz While creating issues I just saw that we are generating codecs for `Transaction*` services which is not supported by this client. We can skip generating them since they are defined as different services.

+1 from me for skipping those codecs.

puzpuzpuz:
Could you update and merge this one when you're back from the vacation?

Issue #523, opened 2020-04-10
Convert tests to ES6 syntax


puzpuzpuz:
Currently tests (both unit and integration ones) are written with ES5 syntax which may be error-prone under certain conditions (say, due to variable hoisting). We should start using ES6 syntax (`const`/`let`, arrow functions, etc.) in tests.

Note: certain language features may be not available in Node v4. So, this task potentially depends on #519.

smitchell41:
I'd like to work on this, if I can!

puzpuzpuz:
@smitchell41 first of all, many thanks for your interest! For sure you can pick this issue. I suggest you to wait for #522 to be merged before working on it. That PR involves many changes in tests. I'll post an update here once it's merged, which should happen soon.

puzpuzpuz:
@smitchell41 as #522 is merged now, there are no blockers for this improvement anymore. Please let me know if you need any help on this one. Looking forward for your contributions! 👍

smitchell41:
@puzpuzpuz I'm so sorry, but I've ended up with a lot more to do this summer than I expected, and I don't think I have time to work on this. Thank you for your help, though.

puzpuzpuz:
@smitchell41 no problem at all. Thanks for your interest in the project.

EaswarDSharma:
can i take this one up? is it assigned to someone?
and these are all the files to convert right?
[https://github.com/hazelcast/hazelcast-nodejs-client/tree/master/test](url)

puzpuzpuz:
@EaswarDSharma I'm currently updating many tests as a part of #554. In the meantime you could pick up any other issue labeled with `good first issue`.

As the the directory with tests, then yes, you have pointed at the right one.

Issue #496, opened 2019-07-24
Documentation for the custom SSLOptionsFactory is misleading


mdumandag:
`SSLOptionsFactory` implementations should return `Promise<void>` for its init method, however in the sample code provided for the readme, it returns just `void`.

Issue #418, opened 2018-11-22
Update winston example


jcaprice:
Winston has a new factory method for creating a logger, `winston.createLogger`.

`new (winston.Logger)` is not working with the latest winston release. This updated example works:

```
var winstonAdapter = {
    logger: winston.createLogger({
        transports: [
            new (winston.transports.Console)()
        ]
    }),

    levels: [
        'error',
        'warn',
        'info',
        'debug',
        'silly'
    ],

    log: function (level, className, message, furtherInfo) {
        this.logger.log(this.levels[level], className + ' ' + message);
    }
};
```

furkansenharputlu:
Which latest version? `3.x-latest` or `2.x-latest`?

jcaprice:
3.x-latest. I'm currently running 3.1.

PR #554, opened 2020-07-21
Redesign Configuration

puzpuzpuz:
* Simplifies client configuration in order to make it more "Node.js native"
  - Includes documentation changes
* Migrates tests to ES6 syntax (in future we should also migrate them to async/await)
* Migrates documentation snippets and code samples to ES6 syntax + async/await
* Converts all enum values used in config options to strings
* Changes default `evictionPolicy` to `LRU`
* Removes unused fields from `SerializationServiceV1`
* Uses `Serializer` interface in `serializer` config instead of `any`
* Closes: #523
* Closes: #496
* Closes: #418
* Also reverts IMDG version to 4.0.2 due to test failures that I had on my machine

Client protocol counterpart PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/333

Main changes are:
* No more imports and constructor calls (i.e. `new Config.ClientConfig()`) are required now in vanilla JS.
* Declarative Configuration (JSON) feature has gone. That's because loading a JSON in Node.js is a matter of `require('hazelcast-client.json')` and, thus, can be easily done on the end application side (as well as any other advanced config logic, like separate loading different configurations for dev/test/production environments and so on).
* This also means that all listeners/factories/etc. have to be configured programmatically instead of the old path resolution way. That's because there is no such thing as class path in Node.js and paths (say, to a listener function) specified in JSON config have to be either relative or absolute. All of that make this feature error-prone and inconvenient for the end users.

Corresponding PRD: https://hazelcast.atlassian.net/wiki/spaces/PM/pages/2016444617/Node.js+Client+Redesign+Configuration

### Programmatic configuration (old way)

```js
const { Client, Config } = require('hazelcast-client');

// Create a configuration object
const clientConfig = new Config.ClientConfig();

// Customize the client configuration
clientConfig.clusterName = 'cluster-name';
clientConfig.networkConfig.addresses.push('10.90.0.2:5701');
clientConfig.networkConfig.addresses.push('10.90.0.3:5701');
clientConfig.listeners.addLifecycleListener(function (state) {
    console.log('Lifecycle Event >>> ' + state);
});

// Initialize the client with the given configuration
const client = await Client.newHazelcastClient(clientConfig);
```

### Programmatic configuration (new way)

```js
const { Client } = require('hazelcast-client');

// Initialize the client with the given configuration
const client = await Client.newHazelcastClient({
    clusterName: 'cluster-name',
    network: {
        addresses: [
            '10.90.0.2:5701',
            '10.90.0.3:5701'
        ]
    },
    lifecycleListeners: [
        (state) => {
            console.log('Lifecycle Event >>> ' + state);
        }
    ]
});
```

#### Outstanding questions

* We could remove support for multiple listeners of the same type. This would help with flattening the config object. For instance, users will be able to register only a single lifecycle listener:
```js
    lifecycleListener: (state) => {
        console.log('Lifecycle Event >>> ' + state);
    }
```


puzpuzpuz:
@huseyinbabal you maintain one of libraries integrated with the Node.js client. So, I would be glad to hear your opinion on the proposed changes.

huseyinbabal:
> @huseyinbabal you maintain one of libraries integrated with the Node.js client. So, I would be glad to hear your opinion on the proposed changes.

@puzpuzpuz I really liked the new notation 🥰

puzpuzpuz:
@huseyinbabal many thanks for your feedback 🙏🏻

puzpuzpuz:
Squashed commits and updated the description.

puzpuzpuz:
@mdumandag we have some conflicts in this one as #549 was merged. I'm going to resolve the conflicts.

puzpuzpuz:
@mdumandag I've faced some test failures on my machine and had to revert IMDG version to 4.0.2 to mitigate them. Going to change it to `4.1-SNAPSHOT` in future.

Other than that the PR should be ready to review once there is a green build.

src/config/Config.ts
```diff
@@ -15,34 +15,113 @@
...
+
+    /**
+     * Custom credentials to be used as a part of authentication on the cluster.
+     */
+    customCredentials?: any;
+
+    /**
+     * User-defined properties.
+     */
+    properties?: Properties;
```
mdumandag, line 120: What do you think about making available options in the `properties` top-level configuration elements? Like `heartbeatInterval`, `heartbeatTimeout` etc. I believe they are hard to use now (requires lots of typing and hard to remember) but moving them to top-level may pollute the config options.

DOCUMENTATION.md
```diff
@@ -296,94 +296,55 @@ These configuration elements are enough for most connection scenarios. Now we wi
...
-**Declarative Configuration:**
-
-```json
-{
-    "network": {
-        "clusterMembers": [
-            "some-ip-address:port"
+const cfg = {
+    networkConfig: {
+        addresses: [
```
mdumandag, line 343: This is referred as `clusterMembers` in other parts

DOCUMENTATION.md
```diff
@@ -576,150 +525,28 @@ Hazelcast Node.js client supports the following data structures and features:
...
-
-```json
-{
-    "network": {
-        "clusterMembers": [
-            "127.0.0.10:4001",
-            "127.0.0.11:4001"
+const cfg = {
+    networkConfig: {
+        addresses: [
```
mdumandag, line 540: same as above. `addresses` -> `clusterMembers`

DOCUMENTATION.md
```diff
@@ -1089,108 +874,71 @@ By default, JSON serialization is used if the object is not `IdentifiedDataSeria
...
-        "smartRouting": true,
-        "redoOperation": true,
-        "connectionTimeout": 6000
-    }
-}
-```
+# 5. Setting Up Client Network
 
-**Programmatic Configuration:**
+All network related configuration of Hazelcast Node.js client is performed via the `network` element in the configuration. Let's first give the examples for these two approaches. Then we will look at its sub-elements and attributes.
```
mdumandag, line 929: This may be a bit misleading right now since we have `connectionStrategy` element now.

Also, for the second sentence, we don't have two ways anymore.

DOCUMENTATION.md
```diff
@@ -1089,108 +874,71 @@ By default, JSON serialization is used if the object is not `IdentifiedDataSeria
...
 
 ```javascript
-var clientConfig = new Config.ClientConfig();
-clientConfig.networkConfig.addresses.push('10.1.1.21', '10.1.1.22:5703');
-clientConfig.networkConfig.smartRouting = true;
-clientConfig.networkConfig.redoOperation = true;
-clientConfig.networkConfig.connectionTimeout = 6000;
+const cfg = {
+    network: {
+        addresses: ['10.1.1.21', '10.1.1.22:5703'],
```
mdumandag, line 936: same as above. `addresses` -> `clusterMembers`

DOCUMENTATION.md
```diff
@@ -1199,52 +947,31 @@ Address list is the initial list of cluster addresses which the client will conn
...
-**Declarative Configuration:**
-
-```json
-{
-    "network": {
-        "clusterMembers": [
+```javascript
+const cfg = {
+    network: {
+        addresses: [
```
mdumandag, line 953: same as above. `addresses` -> `clusterMembers`

DOCUMENTATION.md
```diff
@@ -1441,109 +1099,83 @@ while (tryConnect(connectionTimeout)) != SUCCESS) {
...
-The second step is initializing the `HazelcastClient` to be connected to the cluster:
+The fist step is to define configuration and initialize the `HazelcastClient` to be connected to the cluster:
 
 ```javascript
-Client.newHazelcastClient(clientConfig).then(function (client) {
-    // Some operation
+const client = await Client.newHazelcastClient({
+    clusterName: 'dev',
+    network: {
+        addresses: ['10.90.0.1', '10.90.0.2:5702']
```
mdumandag, line 1125: same as above. `addresses` -> `clusterMembers`

src/config/Config.ts
```diff
@@ -15,34 +15,113 @@
...
+    /**
+     * Name of the cluster to connect to. By default, set to `dev`.
+     */
+    clusterName?: string;
+
+    /**
+     * Name of the client instance. By default, set to `hz.client_${CLIENT_ID}`, where
+     * `CLIENT_ID` starts from `0` and it is incremented by `1` for each new client.
+     */
+    instanceName?: string;
```
mdumandag, line 55: What do you think about changing this to `clientName`. I believe the latter one is more intuitive but it is different from the Java counterpart.

config-schema.json
```diff
@@ -88,7 +39,7 @@
         "properties": {
             "$ref": "#/definitions/propertiesObject"
         },
-        "clientLabels": {
+        "labels": {
```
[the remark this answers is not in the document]
mdumandag: By the way, do we use this file somewhere apart from the tests? Does this provide IDE support (like highlighting, code completion) for `JSON` files? Maybe we can remove it ?

DOCUMENTATION.md
mdumandag: I just saw that this problem exists in the `README.md`

src/config/ClientCloudConfig.ts
```diff
@@ -15,15 +15,25 @@
...
+     * Discovery token of your cluster.
      */
-    discoveryToken: string = null;
+    discoveryToken?: string;
+
     /**
-     * enabled true to use hazelcast.cloud
+     * If it is true, cloud discovery is enabled.
      */
+    enabled?: boolean;
```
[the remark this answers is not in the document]
puzpuzpuz: I also see no value in this property. Let me remove it.

config-schema.json
```diff
@@ -115,7 +66,7 @@
                     "minimum": 1000,
                     "default": 5000
                 },
-                "hazelcastCloud": {
+                "cloudConfig": {
```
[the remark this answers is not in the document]
puzpuzpuz: Thanks for spotting this. Fixed in 9bc863361296174f37237294604f814d914dc002

config-schema.json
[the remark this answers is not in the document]
puzpuzpuz: I'm going to move this file into `./test` directory in #563. So, we'll use it only for tests. In future we may completely remove it.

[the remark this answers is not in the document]
puzpuzpuz: Fixed in 7915c7337cfae65250b47fd7c3dbd6774d34968d

src/config/Config.ts
puzpuzpuz: I'd rather leave them as is, as it's a rare case when a user needs to change one of those. Of course, they should be documented well.

DOCUMENTATION.md
puzpuzpuz: Good catch. Fixed in e3042e71bf4d3c6ff0db4b69a31ac871d5d6ae48

src/network/ClientConnectionManager.ts
```diff
@@ -127,15 +125,15 @@ export class ClientConnectionManager extends EventEmitter {
         super();
         this.client = client;
         this.loadBalancer = client.getLoadBalancer();
-        this.labels = Array.from(client.getConfig().labels);
+        this.labels = Array.from(client.getConfig().clientLabels);
```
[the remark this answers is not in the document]
puzpuzpuz: Makes sense. Remove the copy in e3042e71bf4d3c6ff0db4b69a31ac871d5d6ae48

src/config/Config.ts
puzpuzpuz: I think we should try to keep naming close enough to Java client when possible. So, I'd rather keep it as is.

DOCUMENTATION.md
puzpuzpuz: Nice catch. Changed to `Main parts of network related configuration for Hazelcast Node.js client is performed via the `network` configuration option.` in ab63f5f90cffccfd14871b3d175821732ebd2783

src/config/ClientCloudConfig.ts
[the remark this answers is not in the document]
puzpuzpuz: Removed in f694875de0c8dee8feecbeb8961ebadcd1d8ff25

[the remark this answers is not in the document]
puzpuzpuz: I've also noticed and fixed unused `HazelcastCloudDiscovery.connectionTimeoutInMillis` field. See ee86534732abfc8fad4daaa9a32f57a27d6fc713

src/codec/custom/BitmapIndexOptionsCodec.ts
```diff
@@ -19,36 +19,41 @@ import {FixSizedTypesCodec} from '../builtin/FixSizedTypesCodec';
...
 const INITIAL_FRAME_SIZE = UNIQUE_KEY_TRANSFORMATION_OFFSET + BitsUtil.INT_SIZE_IN_BYTES;
 
 export class BitmapIndexOptionsCodec {
-    static encode(clientMessage: ClientMessage, bitmapIndexOptions: BitmapIndexOptions): void {
+    static encode(clientMessage: ClientMessage, bitmapIndexOptions: BitmapIndexOptionsImpl): void {
         clientMessage.addFrame(BEGIN_FRAME.copy());
 
         const initialFrame = Frame.createInitialFrame(INITIAL_FRAME_SIZE, DEFAULT_FLAGS);
-        FixSizedTypesCodec.encodeInt(initialFrame.content, UNIQUE_KEY_TRANSFORMATION_OFFSET, bitmapIndexOptions.uniqueKeyTransformation);
+        const transformationId = uniqueKeyTransformationToId(bitmapIndexOptions.uniqueKeyTransformation);
```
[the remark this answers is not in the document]
puzpuzpuz: I'm thinking of converting `BitmapIndexOptionsCodec` and `IndexConfigCodec` codecs into builtin codecs, so that they're not generated. Do you think it's a bad idea to do so?

puzpuzpuz:
verify

puzpuzpuz:
verify

puzpuzpuz:
verify

puzpuzpuz:
verify

src/codec/custom/BitmapIndexOptionsCodec.ts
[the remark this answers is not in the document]
puzpuzpuz: Implement it in a way that doesn't require changes in codecs. The only change to be made is in https://github.com/hazelcast/hazelcast-client-protocol/pull/333

puzpuzpuz:
@mdumandag could you take another look at the PR? It seems that I have addressed all pending comments.

P.S. CI build fails due to npm mirror returning an HTTP 404 for one of dev dependencies, but `npm test` went successfully on my machine. I'm going to investigate the CI issue, but this shouldn't block the review.

puzpuzpuz:
@mdumandag thanks for the quick and helpful review!

puzpuzpuz:
verify

Issue #566, opened 2020-08-07
Implement ReliableMessageListener functionality for Reliable Topic


puzpuzpuz:
Java client has [`ReliableMessageListener`](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/topic/ReliableMessageListener.java) interface which describes additional methods, like `boolean isLossTolerant()`. On the other hand, Node.js client only has [`MessageListener`](https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/proxy/topic/MessageListener.ts) type which describes a function. At the moment the behavior of listeners in Node.js client more or less match loss tolerant listeners in Java, but that means less flexibility for users. We need to port functionality related with those additional methods.

PR #567, opened 2020-08-07
Fix ReliableTopic implementation

### LINKED ISSUE #543: Fix ReliableTopic implementation
In IMDG 4.0, server does not throw `StaleSequenceError` [anymore](https://github.com/hazelcast/hazelcast/pull/16303). We need to update our implementation accordingly and re-activate `Reliable Topic Proxy#continues operating when stale sequence is reached` test.

puzpuzpuz:
Closes: #543
Refs: #569

* Removes `StaleSequenceError` handling from `ReliableTopicListenerRunner`
* Adds timeout error handling into `ReliableTopicListenerRunner`
* Renames test scripts for Reliable Topic and Ringbuffer
* Migrates tests for Reliable Topic and Ringbuffer to async/await syntax
* Also adds new tests for Reliable Topic and Ringbuffer 

I've created #566 as a side effect of investigating this issue.


src/proxy/topic/ReliableTopicListenerRunner.ts
```diff
@@ -59,45 +59,55 @@ export class ReliableTopicListenerRunner<E> {
...
     public cancel(): void {
         this.cancelled = true;
     }
 
+    private scheduleNext(): void {
+        setImmediate(this.next.bind(this));
+    }
+
+    private handleInternalError(err: Error): boolean {
+        if (err instanceof InvocationTimeoutError) {
```
mdumandag, line 105: There was an issue reported to the main repo. https://github.com/hazelcast/hazelcast/issues/17070

Maybe we could also handle `ClientNotActiveError` and log it in `trace` level instead of `warn`.

test/topic/ReliableTopicTest.js
```diff
@@ -130,97 +121,73 @@ describe('TopicTest', function () {
...
+        const ringbuffer =  topic.getRingbuffer();
+
+        const all = generateItems(clientOne, 10);
+        await ringbuffer.addAll(all);
+        await topic.publish(11);
+
+        const seq = await ringbuffer.tailSequence();
+        const item = await ringbuffer.readOne(seq);
+        const obj = clientOne.getSerializationService().toObject(item.payload);
+        expect(obj).to.equal(11);
```
mdumandag, line 191: Not related to this PR, but do we have an issue about migrating tests to `async/await`? It is a lot readable :smile:

src/proxy/topic/ReliableTopicListenerRunner.ts
```diff
@@ -59,45 +59,55 @@ export class ReliableTopicListenerRunner<E> {
...
-                        this.sequenceNumber = newSequence;
-                        setImmediate(this.next.bind(this));
-                    });
 
-                    return;
+                const nextSeq = result.getNextSequenceToReadFrom().toNumber();
+                const lostCount = nextSeq - result.getReadCount() - this.sequenceNumber;
+                if (lostCount !== 0) {
+                    this.logger.trace('ReliableTopicListenerRunner', 'Listener of topic: ' + this.proxy.getName()
+                        + ' lost ' + lostCount + ' messages');
```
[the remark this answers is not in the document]
puzpuzpuz: Good catch. Changed the logic in 9d6b40d274bcbaf29f9d05a40009bdf2c807736b

src/proxy/topic/ReliableTopicListenerRunner.ts
puzpuzpuz: Thanks for the heads-up. Changed the error logging in 9d6b40d274bcbaf29f9d05a40009bdf2c807736b

test/topic/ReliableTopicTest.js
puzpuzpuz: Yeah, it's a lot more readable. I was going to migrate tests incrementally as a part of other PRs. But having this plan defined as a GH issue is a good idea. Going to create one.

puzpuzpuz: Here it goes: #569

puzpuzpuz:
@mdumandag I've addressed pending comments. Could you take another look?

src/proxy/topic/ReliableTopicListenerRunner.ts
```diff
@@ -65,9 +69,13 @@ export class ReliableTopicListenerRunner<E> {
 
                 const nextSeq = result.getNextSequenceToReadFrom().toNumber();
                 const lostCount = nextSeq - result.getReadCount() - this.sequenceNumber;
+                // If messages were lost, behave as a non-loss tolerant listener
                 if (lostCount !== 0) {
-                    this.logger.trace('ReliableTopicListenerRunner', 'Listener of topic: ' + this.proxy.getName()
-                        + ' lost ' + lostCount + ' messages');
+                    this.logger.warn('ReliableTopicListenerRunner', 'Terminating listener of topic: '
+                        + this.proxy.getName() + '. Reason: The listener was too slow lost or the retention'
```
mdumandag, line 75: Minor: `Reason: The listener was too slow lost or the retention...` should be `Reason: The listener was too slow or the retention...` (extra `lost`)

puzpuzpuz: Fixed in ab19d9c259760e62bccbcb020cf913f3e78572f8

puzpuzpuz:
@mdumandag thanks for the quick review!

src/proxy/topic/ReliableTopicListenerRunner.ts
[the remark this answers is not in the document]
puzpuzpuz: I'd like to reiterate on this one.

The previous implementation was handling `StaleSequenceError` as if it was based on a loss tolerant listener. I think we should do the same here. Otherwise we leave users sitting on a time bomb. If producers have a spike in message generation for a single time, the listener will be terminated and users will have no programmatic way to handle the termination and, say, register another listener. I have reverted the logic in cdcf57cfd71c9dc77ee9f9639be046f740d3b04e

WDYT?

Off-topic: I have also removed redundant `setImmediate` from `ReliableTopicListenerRunner`. See: 56a37ca07bf7a3f66474bd137c22b9239b63c565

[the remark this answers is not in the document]
mdumandag: Hmm, okay, that makes sense. With a `warn` level log, I guess, it should be okay.

PR #570, opened 2020-08-12
Fix entry point re-exports

puzpuzpuz:
Depends on #568

* Introduces `index.ts` files in subdirs for the purpose of re-exporting public modules. This way root `index.ts` (the entry point) becomes much simpler and it's simpler to make sure that we export all public APIs from the entry point
* Introduces `Portable` module which contains public API definitions for portable serialization
* Moves modules into subdirs, so the overall structure is more consistent
* Merges certain related modules into a single one
* Adds documentation for wildcard names and default DS configurations
* Changes `MembershipListener` registration id type to `string` (was `UUID`), so it becomes consistent with registration types for other listeners (e.g. `IMap#addEntryListener`)
* Includes a bunch of improvements in tsdoc and code style

Client protocol counterpart PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/336


puzpuzpuz:
verify

test/ringbuffer/RingbufferProxyTest.js
```diff
@@ -15,146 +15,129 @@
  */
 'use strict';
 
-const { expect } = require('chai');
```
mdumandag, line 18: Why are we replacing async/await with promise chains? There are also test removals and some changes in the test logic. Are these intentional or some kind of merging error?

src/core/RestValue.ts
```diff
@@ -35,18 +35,18 @@ export class RestValue implements IdentifiedDataSerializable {
      * HTTP Content-Type specified for the value.
      */
     contentType: string;
-    /** @internal */
+    /** @ignore */
```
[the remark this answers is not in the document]
puzpuzpuz: No. The change is done because of `RestValue` not being exported in the `index.ts` previously. Once I have fixed that, `npm run validate-user-code` started failing. So, user code validation has already proven itself as a valuable test.

test/ringbuffer/RingbufferProxyTest.js
puzpuzpuz: That's not how it's meant to be. Looks like a rebase error. Thanks for noticing this and let me fix it.

src/invocation/ClusterService.ts
```diff
@@ -126,11 +124,7 @@ export class ClusterService implements Cluster {
...
 
-    /**
-     * @param listener The listener to be registered.
-     * @return The registration ID
-     */
-    public addMembershipListener(listener: MembershipListener): UUID {
+    addMembershipListener(listener: MembershipListener): string {
         assertNotNull(listener);
 
         const registrationId = UuidUtil.generate();
```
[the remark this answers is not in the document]
puzpuzpuz: Nice suggestion. Addressed in bd8ee83960cbd49187f850240c8b6cb6edf740d6

test/ringbuffer/RingbufferProxyTest.js
puzpuzpuz: Fixed in deac2c149bd4aa5c0d1e4aba4c5e24917cfb62dd

puzpuzpuz:
> In general, looks good to me. There are lots of `imports can be shortened` warnings shown in WebStorm but I think this is not a big problem.

I don't see such suggestions in VS Code, but yes, imports can be improved further, but that can be improved incrementally in upcoming PRs.

PR #571, opened 2020-08-12
Remove handled fragmented messages from the map

mdumandag:
Depends on #568 (because of CI build changes)

We should remove handled fragmented messages from fragmentedMessages
map to free it up. This PR includes this fix.

Also, FragmentedClientMessageHandlerTest is enhanced to check the
the map's size before and after merging and handling the
end fragment.

### DIFF
--- a/src/network/ClientConnection.ts
+++ b/src/network/ClientConnection.ts
@@ -242,6 +242,11 @@ export class ClientMessageReader {
 /** @internal */
 export class FragmentedClientMessageHandler {
     private readonly fragmentedMessages = new Map<number, ClientMessage>();
+    private readonly logger: ILogger;
+
+    constructor(logger: ILogger) {
+        this.logger = logger;
+    }
 
     handleFragmentedMessage(clientMessage: ClientMessage, callback: Function): void {
         const fragmentationFrame = clientMessage.startFrame;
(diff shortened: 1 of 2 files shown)


puzpuzpuz:
verify

src/network/ClientConnection.ts
```diff
@@ -244,19 +244,15 @@ export class FragmentedClientMessageHandler {
         clientMessage.dropFragmentationFrame();
         if (fragmentationFrame.hasBeginFragmentFlag()) {
             this.fragmentedMessages.set(fragmentationId, clientMessage);
-        } else if (fragmentationFrame.hasEndFragmentFlag()) {
-            const mergedMessage = this.mergeIntoExistingClientMessage(fragmentationId, clientMessage);
-            callback(mergedMessage);
         } else {
-            this.mergeIntoExistingClientMessage(fragmentationId, clientMessage);
+            const existingMessage = this.fragmentedMessages.get(fragmentationId);
```
puzpuzpuz, line 248: Is it theoretically possible that a client didn't receive previous fragment and thus `existingMessage` is `undefined`?

mdumandag: Yes, in fact, this is possible. I added a check before. If it has no begin part, It logs this with the fragmentation id (this is the only thing we could extract from middle or last fragments). Also, enhanced the test case with this scenario

puzpuzpuz: Thanks for the update. The change looks good to me.

mdumandag:
Thanks for the review Andrey, I am merging this.

PR #572, opened 2020-08-17
Add Buffer type support

puzpuzpuz:
Depends on #570

* Adds native support for node.js' `Buffer` type instead of array of numbers used previously
* Also fixes `null` case handling in `ObjectDataInput#readArray`

Corresponding PRD: https://hazelcast.atlassian.net/wiki/spaces/PM/pages/1749254626/Node.js+Client+Native+Binary+Data+Type

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -553,18 +553,22 @@ In the following chapters you will learn the description of all options supporte
 
 Serialization is the process of converting an object into a stream of bytes to store the object in the memory, a file or database, or transmit it through the network. Its main purpose is to save the state of an object in order to be able to recreate it when needed. The reverse process is called deserialization. Hazelcast offers you its own native serialization methods. You will see these methods throughout this chapter.
 
-Hazelcast serializes all your objects before sending them to the server. The `boolean`, `number`,`string` and `Long` types are serialized natively and you cannot override this behavior. The following table is the conversion of types for the Java server side.
+Hazelcast serializes all your objects before sending them to the server. Certain types, like `boolean`, `number`, `string`, and `Long`, are serialized natively and you cannot override this behavior. The following table is the conversion of types for the Java server side.
 
-| Node.js | Java                                |
-|---------|-------------------------------------|
-| boolean | Boolean                             |
-| number  | Byte, Short, Integer, Float, Double |
-| string  | String                              |
-| Long    | Long                                |
+| Node.js         | Java                                  |
+|-----------------|---------------------------------------|
+| boolean         | Boolean                               |
+| number          | Byte, Short, Integer, Float, Double   |
+| string          | String                                |
+| Long            | Long                                  |
+| Buffer          | byte[]                                |
+| Object          | com.hazelcast.core.HazelcastJsonValue |
 
-> **NOTE: A `number` type is serialized as `Double` by default. You can configure this behavior using the `serialization.defaultNumberType` config option.**
+> **NOTE: The `Long` type means the type provided by [long.js library](https://github.com/dcodeIO/long.js).**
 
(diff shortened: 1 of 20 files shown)


puzpuzpuz:
@mdumandag I've just rebased this one. Could you take a look at it when you have some time?

src/serialization/ObjectData.ts
```diff
@@ -338,8 +342,22 @@ export class ObjectDataInput implements DataInput {
...
-    readByteArray(pos?: number): number[] {
-        return this.readArray<number>(this.readByte, pos);
+    readByteArray(pos?: number): Buffer {
+        const backupPos = this.pos;
+        if (pos !== undefined) {
+            this.pos = pos;
+        }
+        const len = this.readInt();
+        if (len === BitsUtil.NULL_ARRAY_LENGTH) {
+            return null;
```
mdumandag, line 352: If the `pos` (the argument one) is not `undefined` shouldn't we set back `this.pos` to `backupPos` before returning?. I checked the `readArray` method and it seems it is not doing it also.

By the way, Java side does not expect `pos` argument for `readXArray` methods. I cannot recall whether or not this argument is needed. Maybe we could also check that.

DOCUMENTATION.md
```diff
@@ -553,18 +553,22 @@ In the following chapters you will learn the description of all options supporte
...
-| string  | String                              |
-| Long    | Long                                |
+| Node.js         | Java                                  |
+|-----------------|---------------------------------------|
+| boolean         | Boolean                               |
+| number          | Byte, Short, Integer, Float, Double   |
+| string          | String                                |
+| Long            | Long                                  |
+| Buffer          | byte[]                                |
+| Object          | com.hazelcast.core.HazelcastJsonValue |
```
mdumandag, line 565: Should we also put `HazelcastJsonValue` (the client's class) here?

mdumandag (commented):
Looks good to me with a question.

Also, although I see that we are returning a slice in the `readByteArray` method, should we also add a test that verifies that?

src/serialization/ObjectData.ts
puzpuzpuz: > If the `pos` (the argument one) is not `undefined` shouldn't we set back `this.pos` to `backupPos` before returning?. I checked the `readArray` method and it seems it is not doing it also.

We restore `this.pos`, just like it's done in `readArray`. See L356.

DOCUMENTATION.md
puzpuzpuz: It's described in the `JSON Serialization` section and most users won't have to deal with it. Only those who want to avoid eager JSON parsing will set `jsonStringDeserializationPolicy` to `NO_DESERIALIZATION` and then work with instances of the class.

So, I've decided to keep the table as small as possible. We also have `RestValue` which I didn't mention in the table for the same reason.

WDYT?

src/serialization/ObjectData.ts
mdumandag: I meant the early return

```js
if (len === BitsUtil.NULL_ARRAY_LENGTH) {
    // this.pos is corrupt now if the pos is undefined
    return null;
}
```

DOCUMENTATION.md
mdumandag: Okay then, it seems reasonable

src/serialization/ObjectData.ts
puzpuzpuz: Oups, missed that point. Yes, it should be restored. Let me fix that.

puzpuzpuz: Fixed in 71d242795651a102cea7dc54ae94d3d6a806003b

mdumandag (approved):

PR #573, opened 2020-08-18
Add boomerang backups support

puzpuzpuz:
* Implements support of backup acks to client (a.k.a. boomerang backups)
* Adds related unit tests for `InvocationService` and `Invocation`
* Slightly improves code quality in `InvocationService`, `Invocation` and `ClientMessage`
* Also migrates `ClientLabelTest` to async/await syntax

Corresponding PRD: https://hazelcast.atlassian.net/wiki/spaces/PM/pages/1764032832/Node.js+Client+Boomerang+Backups

### Feature Overview

So far clients were waiting for the sync backups to complete on the member. This was causing 4 network hops to complete a client operation with sync backup. Since sync backup configuration is our out-of-box experience, we would like to improve it. Boomerang backup design decreases network hops to 3, thus improving the performance. 

boomerang-backups

### DIFF
--- a/src/invocation/InvocationService.ts
+++ b/src/invocation/InvocationService.ts
@@ -22,6 +22,7 @@ import {
     ClientNotActiveError,
     HazelcastInstanceNotActiveError,
     InvocationTimeoutError,
+    IndeterminateOperationStateError,
     IOError,
     RetryableHazelcastError,
     TargetDisconnectedError,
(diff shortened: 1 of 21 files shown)


src/invocation/InvocationService.ts
```diff
@@ -237,49 +370,50 @@ export class InvocationService {
...
-
     /**
      * Removes the handler for all event handlers with a specific correlation id.
-     * @param id correlation id
      */
-    removeEventHandler(id: number): void {
-        if (this.eventHandlers.hasOwnProperty('' + id)) {
-            delete this.eventHandlers[id];
+    removeEventHandler(correlationId: number): void {
+        if (this.eventHandlers.has(correlationId)) {
```
mdumandag, line 377: I think we can call `delete` without checking for the existence of `correlationId` as key

src/invocation/InvocationService.ts
```diff
@@ -132,6 +163,73 @@ export class Invocation {
 
         return false;
     }
+
+    notify(clientMessage: ClientMessage): void {
+        assert(clientMessage != null, 'Response can not be null');
+        const expectedBackups = clientMessage.getNumberOfBackupAcks();
+        if (expectedBackups > this.backupsAcksReceived) {
+            this.pendingResponseReceivedMillis = Date.now();
```
mdumandag, line 171: This block is executed once per the invocation, right?

mdumandag (commented):
Looks good to me. I have a question and a minor comment

src/invocation/InvocationService.ts
puzpuzpuz: Good point. Fixed in 615278a59d65da7672c6120d9b7156e0f2726259

src/invocation/InvocationService.ts
puzpuzpuz: Yes, it's executed once per response from the target member.

mdumandag (approved):

PR #576, opened 2020-08-24
Add serverStackTrace property in HazelcastError class

### LINKED ISSUE #542: Display stack frames of the server exceptions
In #522, we are getting stack frames of server side errors but not displaying it. 

I am not sure how to do that. Official documentation states [that](https://nodejs.org/api/errors.html#errors_error_stack) stack frames for errors are generated lazily. So, I am not confident about using `error.stack` for that purposes. Maybe we can add a new property for server side stack. Anyways, we need to decide on how to display them.

puzpuzpuz:
Closes #542

* Adds `serverError` property in `HazelcastError` class, so it now contains full information on the server-side error
* Introduces new error classes to be used in `ClientErrorFactory`
* Adds new `ClientErrorFactoryTest`

### DIFF
--- a/src/core/HazelcastError.ts
+++ b/src/core/HazelcastError.ts
@@ -22,45 +22,101 @@ export interface HazelcastErrorConstructor {
 }
 
 /**
- * Base class for all specific errors thrown by Hazelcast client.
+ * Represents a stack trace element of server-side exception.
+ */
+export interface ServerErrorStackElement {
+
+    /**
+     * Class name.
+     */
+    className: string;
+
+    /**
+     * Method name.
+     */
+    methodName: string;
+
+    /**
+     * Name of the file containing the class.
+     */
+    fileName: string;
+
+    /**
+     * Line number from the class.
+     */
+    lineNumber: number;
+
+}
+
+/**
+ * Represents server-side exception.
+ */
+export interface ServerError {
+
+    /**
+     * Error code.
+     */
+    errorCode: number;
+
+    /**
+     * Name of the class responsible for throwing the exception.
+     */
+    className: string;
+
+    /**
+     * Error message.
+     */
+    message: string;
+
+    /**
+     * Server-side stack trace.
+     */
+    stackTraceElements: ServerErrorStackElement[];
+
+}
+
+/**
+ * Base class for all specific exceptions thrown by Hazelcast client.
  */
 export class HazelcastError extends Error {
 
     cause: Error;
     stack: string;
+    serverError?: ServerError;
 
-    constructor(msg: string, cause?: Error) {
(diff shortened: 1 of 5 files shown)


src/core/HazelcastError.ts
```diff
@@ -22,45 +22,101 @@ export interface HazelcastErrorConstructor {
...
+     * Line number from the class.
+     */
+    lineNumber: number;
+
+}
+
+/**
+ * Represents server-side exception.
+ */
+export interface ServerError {
```
mdumandag, line 54: I think there are some duplication and unnecessary information here. 

`errorCode` does not mean a lot to the users. It is already used to instantiate the error.
`message` is the same as the error's message.
`className`  is almost the same as the error's class name.

I believe the only interesting part here is the `stackTraceElements`. Maybe, we could rename `HazelcastError`s `serverError` property with something like `serverStackTrace`. What do you think about it?

src/protocol/ErrorFactory.ts
```diff
@@ -66,119 +76,119 @@ export class ClientErrorFactory {
...
             return null;
         }
         const errorHolder = errorHolders[errorHolderIdx];
         const factoryFn = this.codeToErrorConstructor.get(errorHolder.errorCode);
         let error: Error;
         if (factoryFn != null) {
-            error = factoryFn(errorHolder.message, this.createError(errorHolders, errorHolderIdx + 1));
+            error = factoryFn(errorHolder.message, this.createError(errorHolders, errorHolderIdx + 1), errorHolder);
         } else {
             error = new UndefinedErrorCodeError(errorHolder.message, errorHolder.className);
```
mdumandag, line 193: I see that once we are faced with an unknown error code, we stop iterating over the error holders. I know that it's a long shot but there may be other elements in the error tree that we can extract something. Maybe, we could continue iterating while setting the `cause` of `UndefinedErrorCodeError` to the next element. What do you think about it?

src/core/HazelcastError.ts
puzpuzpuz: Fair point. I've removed `ServerError` interface and left only the `serverStackTrace: ServerErrorStackElement[]` property `HazelcastError`. Done in 08d69ff0e778006bc18f01d447d78a4c6464238c

src/protocol/ErrorFactory.ts
puzpuzpuz: I didn't touch the old logic, but that's a nice suggestion. Implemented it in b947edaf337744e1a0ab6b3976abd6ca5ade386f

mdumandag (approved):
Thanks for the changes, Andrey. Looks good to me

puzpuzpuz:
@mdumandag thanks for the helpful review!

PR #577, opened 2020-08-25
[CP Subsystem] Add AtomicLong support

puzpuzpuz:
* Adds `AtomicLong` as the first part of CP Subsystem support
* Includes code samples and documentation
* Also includes minor tsdoc and code style improvements

Corresponding client protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/339

Corresponding PRD: https://hazelcast.atlassian.net/wiki/spaces/PM/pages/1457030796/Node.js+Client+CP+Sub+System

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -60,7 +60,9 @@
     * [7.4.9. Using PN Counter](#749-using-pn-counter)
     * [7.4.10. Using Flake ID Generator](#7410-using-flake-id-generator)
       * [7.4.10.1. Configuring Flake ID Generator](#74101-configuring-flake-id-generator)
-    * [7.4.11. Using Lock, Semaphore and Atomic Long](#7411-using-lock-semaphore-and-atomic-long)
+    * [7.4.11. CP Subsystem](#7411-cp-subsystem)
+      * [7.4.11.1. Using Atomic Long](#74111-using-atomic-long)
+      * [7.4.11.2. Using Lock and Semaphore](#74112-using-lock-and-semaphore)
   * [7.5. Distributed Events](#75-distributed-events)
     * [7.5.1. Listening for Cluster Events](#751-listening-for-cluster-events)
       * [7.5.1.1. Membership Listener](#7511-membership-listener)
@@ -1247,11 +1249,11 @@ const map = await client.getReplicatedMap('my-replicated-map');
 // Put and get a value from the Replicated Map
 // (key/value is replicated to all members)
 const replacedValue = await map.put('key', 'value');
-// Will print 'replaced value = null' as it's the first update
-console.log('replaced value = ' + replacedValue);
+// Will print 'Replaced value: null' as it's the first update
+console.log('Replaced value:', replacedValue);
 const value = map.get('key');
 // The value is retrieved from a random member in the cluster
-console.log('value = ' + value);
+console.log('Value:', value);
 ```
 
 Hazelcast Replicated Map uses `EntryListener` to listen to the events that occur when the entries are added to, updated in or evicted/removed from the Replicated Map. See the [Entry Listener section](#7522-entry-listener) for information on how to create an entry listener object and register it.
@@ -1270,12 +1272,12 @@ await queue.offer('item');
 // Poll the Queue and return the string
 let item = queue.poll();
 // Timed-restricted operations
-await queue.offer('anotheritem', 500);
-await queue.poll(5000);
+await queue.offer('anotheritem', 500); // waits up to 500 millis
+await queue.poll(5000); // waits up to 5000 millis
 // Indefinitely-waiting operations
 await queue.put('yetanotheritem');
-const value = queue.take();
-console.log(value);
+const item = queue.take();
+console.log('Item:', item);
 ```
 
 Hazelcast Queue uses `ItemListener` to listen to the events that occur when the items are added to or removed from the Queue. See the [Item Listener section](#7523-item-listener) for information on how to create an item listener object and register it.
@@ -1298,7 +1300,7 @@ await set.add('item2');
 await set.add('item3');
 // Get the items. Note that there are no duplicates
 const values = await set.toArray();
-console.log(values);
+console.log('Values:', values);
 ```
 
 Hazelcast Set uses `ItemListener` to listen to the events that occur when the items are added to or removed from the Set. See the [Item Listener section](#7523-item-listener) for information on how to create an item listener object and register it.
@@ -1317,10 +1319,10 @@ await list.add('item1');
 await list.add('item2');
 // Remove the first element
 const value = await list.removeAt(0);
-console.log(value);
+console.log('Value:', value);
 // There is only one element left
 const len = await list.size();
-console.log(len);
+console.log('Length:', len);
 // Clear the list
 await list.clear();
 ```
@@ -1343,9 +1345,9 @@ await rb.add(200);
 // If you want to start from the next item, call rb.tailSequence()+1
 const sequence = await rb.headSequence();
 let value = await rb.readOne(sequence);
-console.log(value);
+console.log('Value:', value);
 value = await rb.readOne(sequence.add(1));
-console.log(value);
+console.log('Next value:', value);
 ```
 
 ### 7.4.8. Using Reliable Topic
@@ -1359,7 +1361,7 @@ A Reliable Topic usage example is shown below.
 const topic = await client.getReliableTopic('my-distributed-topic');
 // Add a Listener to the Topic
 topic.addMessageListener((message) => {
-    console.log(message);
+    console.log('Message:', message);
 });
 // Publish a message to the Topic
 await topic.publish('Hello to the distributed world!');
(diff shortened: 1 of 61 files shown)


test/cpsubsystem/AtomicLongProxyTest.js
```diff
@@ -0,0 +1,162 @@
...
+
+        const value1 = await longInAnotherGroup.incrementAndGet();
+        expectLong(1, value1);
+        // the following value has to be 0,
+        // as `long` belongs to the default CP group
+        const value2 = await long.get();
+        expectLong(0, value2);
+    });
+
+    it('destroy: should destroy AtomicLong', async function () {
```
mdumandag, line 71: Minor: Looking at the tests at the Java side, I see that they are testing the cases of multiple destroys and getAtomicLong call after destroy. Maybe we could add those

test/cpsubsystem/AtomicLongProxyTest.js
```diff
@@ -0,0 +1,162 @@
...
+        expectLong(0, value);
+    });
+
+    it('getAndAdd: should add', async function () {
+        await long.getAndAdd(123);
+        const value = await long.get();
+        return expectLong(123, value);
+    });
+
+    it('decrementAndGet: should decrement', async function () {
```
mdumandag, line 109: Looking at the other tests, I see that this one does not have a counterpart in which you test the `get` value after a call to `decrementAndGet` (like `getAndAdd: should return old value` and `getAndAdd: should add`)

test/cpsubsystem/AtomicLongProxyTest.js
```diff
@@ -0,0 +1,162 @@
...
+        expectLong(0, value);
+    });
+
+    it('getAndSet: should set the value', async function () {
+        await long.getAndSet(-123);
+        const value = await long.get();
+        expectLong(-123, value);
+    });
+
+    it('incrementAndGet: should increment', async function () {
```
mdumandag, line 147: Same comment with `decrementAndGet: should decrement`

src/proxy/IAtomicLong.ts
```diff
@@ -0,0 +1,108 @@
...
+     */
+    getAndIncrement(): Promise<Long>;
+
+    /**
+     * Atomically sets the given value.
+     *
+     * @param newValue the new value
+     */
+    set(newValue: Long | number): Promise<void>;
+}
```
mdumandag, line 108: Are we going to add APIs that take `IFunction` types as arguments (like alter and apply)?

src/proxy/cpsubsystem/BaseCPProxy.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as Promise from 'bluebird';
+import {HazelcastClient} from '../../HazelcastClient';
+import {ClientMessage} from '../../protocol/ClientMessage';
+import {UnsupportedOperationError} from '../../core';
+
+/**
+ * Common super class for any CP Subsystem proxy.
+ * @internal
+ */
+export abstract class BaseCPProxy {
```
mdumandag, line 27: This seems very similar to `BaseProxy`, is there a special use case of this?

mdumandag (commented):
Looks pretty good to me, I have some minor comments

src/proxy/IAtomicLong.ts
puzpuzpuz: Do you propose to send JS code and execute it on the cluster side? That may have security complications. The old implementation of AtomicLong didn't support those operations.

src/proxy/cpsubsystem/BaseCPProxy.ts
puzpuzpuz: It's similar, but different, as it has no `destroy` implementation, throws on `getPartitionKey` call, has a different implementation of `encodeInvokeOnRandomTarget` method, and has additional properties. As CP DSes have a different manager class (and a different overall behavior), I don't see any good reason to use `BaseProxy` class here.

test/cpsubsystem/AtomicLongProxyTest.js
puzpuzpuz: That's because `decrementAndGet` returns the new value, not the old one like `getAndAdd`. So, we expect to get the new value in this very test.

test/cpsubsystem/AtomicLongProxyTest.js
puzpuzpuz: See my comment for `decrementAndGet`

test/cpsubsystem/AtomicLongProxyTest.js
puzpuzpuz: Makes sense. Improved in 3e4a97ecbb3e8aaf69d5e1a3fb77097cdf0f398c

src/proxy/cpsubsystem/CPProxyManager.ts
```diff
@@ -0,0 +1,93 @@
...
+/** @internal */
+export function withoutDefaultGroupName(name: string): string {
+    assertString(name);
+    name = name.trim();
+    const i = name.indexOf('@');
+    if (i === -1) {
+        return name;
+    }
+
+    assert(name.slice(i + 1).indexOf('@') === -1, 'Custom group name must be specified at most once');
```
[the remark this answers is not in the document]
puzpuzpuz: Improved in 959b3cf3f221f20a2d1ae9cf9d6de0e42933ba8e

DOCUMENTATION.md
```diff
@@ -1454,11 +1456,38 @@ The following are the descriptions of configuration elements and attributes:
...
+// Increment by 42
+await atomicLong.addAndGet(42);
+// Set to 0 atomically if the current value is 42
+const result = atomicLong.compareAndSet(42, 0);
+console.log('CAS operation result:', result);
+```
+
+`IAtomicLong` implementation does not offer exactly-once / effectively-once execution semantics. It goes with at-least-once execution semantics by default and can cause an API call to be committed multiple times in case of CP member failures. It can be tuned to offer at-most-once execution semantics. Please see [`fail-on-indeterminate-operation-state`](https://docs.hazelcast.org/docs/latest/manual/html-single/index.html#cp-subsystem-configuration) server-side setting.
+
+### 7.4.11.2. Using Lock and Semaphore
```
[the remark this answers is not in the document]
puzpuzpuz: Fixed in 959b3cf3f221f20a2d1ae9cf9d6de0e42933ba8e

DOCUMENTATION.md
```diff
@@ -1454,11 +1456,38 @@ The following are the descriptions of configuration elements and attributes:
...
 
-These new implementations are accessed using the [CP Subsystem](https://docs.hazelcast.org/docs/latest/manual/html-single/#cp-subsystem) which cannot be used with the Node.js client yet. We plan to implement support for the CP Subsystem in the upcoming 4.0 release of Hazelcast Node.js client. In the meantime, since there is no way to access old non-CP primitives using IMDG 4.x, we removed their implementations, code samples and documentations. They will be back once we implement them.
+Before using Atomic Long, Lock, and Semaphore, CP Subsystem has to be enabled on cluster-side. Refer to [CP Subsystem](https://docs.hazelcast.org/docs/latest/manual/html-single/#cp-subsystem) documentation for more information.
+
+> **NOTE: If you call the `DistributedObject.destroy()` method on a CP data structure, that data structure is terminated on the underlying CP group and cannot be reinitialized until the CP group is force-destroyed on the cluster side. For this reason, please make sure that you are completely done with a CP data structure before destroying its proxy.**
+
+### 7.4.11.1. Using Atomic Long
```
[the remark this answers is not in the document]
puzpuzpuz: Thanks for pointing this out. Fixed in 959b3cf3f221f20a2d1ae9cf9d6de0e42933ba8e

test/cpsubsystem/AtomicLongProxyTest.js
mdumandag: I initially thought that but after seeing `addAndGet: should return added value` and `addAndGet: should add`, I thought you may missed these two. 

But, I guess it is okay right now as they are using the `addAndGet` internally.

src/proxy/IAtomicLong.ts
mdumandag: No, I was referring to use the code defined at the Java side (via some Hazelcast serializable format)

puzpuzpuz: `IFunction` is `Serializable` which is not supported by non-Java clients. Even if we allow users to point at server-side function somehow, it will require member-side support. Am I missing something?

test/cpsubsystem/AtomicLongProxyTest.js
puzpuzpuz: > I initially thought that but after seeing `addAndGet: should return added value` and `addAndGet: should add`, I thought you may missed these two.

Those tests are a port of tests from the old AtomicLong test suite. But we should keep them in a good shape. Removed redundant test in 43dd117e433a4f1651ce35a4ab5a4aa9bdf7c240

src/proxy/IAtomicLong.ts
mdumandag: If I am not missing something, we should be able to use classes like https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/client/test/PrefixFilter.java by defining their IdentifiedDataSerializable counterparts for the client. That way, the server should pick up the right implementation on the server-side. Of course, this is hard to use since it requires classes like PrefixFilter on the classpath of the server

puzpuzpuz: Makes sense. I'd prefer not to implement this part of Java-centric functionality, like it was in the old implementation. We could create a GH issue for this potential future enhancement and implement it if someone asks.

mdumandag: That sounds good to me

mdumandag (approved):

puzpuzpuz:

PR #584, opened 2020-09-08
Add drain event handling into PipelinedWriter

puzpuzpuz:
Current implementation of `PipelinedWriter` does not consider the result of `Socket.write()` and `drain` events. This PR fixes that.

### DIFF
--- a/test/connection/PipelinedWriterTest.js
+++ b/test/connection/PipelinedWriterTest.js
@@ -25,22 +25,35 @@ const { DeferredPromise } = require('../../lib/util/Util');
 const { PipelinedWriter } = require('../../lib/network/ClientConnection');
 
 describe('PipelinedWriterTest', function () {
+
     let writer;
     let mockSocket;
 
-    const setUpWriteSuccess = () => {
+    function setUpWriteSuccess() {
         mockSocket = new Socket({});
         sinon.stub(mockSocket, 'write').callsFake((data, cb) => {
-            cb();
-            mockSocket.emit('data', data);
+            process.nextTick(cb);
+            process.nextTick(() => mockSocket.emit('data', data));
+            return true;
         });
         writer = new PipelinedWriter(mockSocket, 8192);
     }
 
-    const setUpWriteFailure = (err) => {
+    function setUpWriteSuccessWaitForDrain() {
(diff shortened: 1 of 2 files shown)


test/connection/PipelinedWriterTest.js
```diff
@@ -25,22 +25,35 @@ const { DeferredPromise } = require('../../lib/util/Util');
 const { PipelinedWriter } = require('../../lib/network/ClientConnection');
 
 describe('PipelinedWriterTest', function () {
+
     let writer;
     let mockSocket;
 
-    const setUpWriteSuccess = () => {
+    function setUpWriteSuccess() {
```
mdumandag, line 32: Minor: I think we can get rid of the duplication from `setUpWriteSuccess` and `setUpWriteSuccessWaitForDrain` by making one of them take a boolean parameter and return it. Something like
```node
    function setUpWrite(shouldWrite) {
        mockSocket = new Socket({});
        sinon.stub(mockSocket, 'write').callsFake((data, cb) => {
            process.nextTick(cb);
            process.nextTick(() => mockSocket.emit('data', data));
            return shouldWrite;
        });
        writer = new PipelinedWriter(mockSocket, 8192);
    }
```
What do you think about it?

mdumandag (approved):
Looks good to me with a minor comment.

By the way, would we gain anything by implementing something similar for the `DirectWriter`? (Start queing write requests instead of writing it to the socket when `socket.write` call returns `false` until we receive the `drain` event)

puzpuzpuz:
> By the way, would we gain anything by implementing something similar for the `DirectWriter`? (Start queing write requests instead of writing it to the socket when `socket.write` call returns `false` until we receive the `drain` event)

Having this in the direct writer would turn it into a simplified pipelined writer, so I'd better avoid doing this and keep it as simple as possible. WDYT?

mdumandag:
@puzpuzpuz okay, it's fine for me :+1:

test/connection/PipelinedWriterTest.js
puzpuzpuz: Good suggestion. Merged the functions in 3aa1ab04a696800623afd5f7d88459c55649097e

PR #586, opened 2020-09-08
Add default serializer for UUID

puzpuzpuz:
This serializer was missing previously.

### DIFF
--- a/src/core/UUID.ts
+++ b/src/core/UUID.ts
@@ -37,12 +37,16 @@ export class UUID {
         this.leastSignificant = leastSig;
     }
 
-    /** @internal */
     equals(other: UUID): boolean {
         if (other == null) {
             return false;
         }
-        return other.mostSignificant.equals(this.mostSignificant) && other.leastSignificant.equals(this.leastSignificant);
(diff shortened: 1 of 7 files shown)


src/core/UUID.ts
```diff
@@ -37,12 +37,16 @@ export class UUID {
...
     equals(other: UUID): boolean {
         if (other == null) {
             return false;
         }
-        return other.mostSignificant.equals(this.mostSignificant) && other.leastSignificant.equals(this.leastSignificant);
+        return other.mostSignificant.equals(this.mostSignificant)
+            && other.leastSignificant.equals(this.leastSignificant);
+    }
+
+    static isUUID(obj: any) {
```
mdumandag, line 48: Minor: Is this for internal use only or do you think that it may be useful for the users? If it is internal, we could mark this as that

mdumandag (approved):
Thanks for the quick fix Andrey

src/core/UUID.ts
puzpuzpuz: I decided to keep this method available for users as it may be handy for them. For the same reason, I've removed `@internal` tag from `equals()`.

PR #587, opened 2020-09-08
Remove bluebird dependency

### LINKED ISSUE #529: Switch to native promises instead bluebird
Native promises have increased performance in Node.js 10+, so [bluebird](https://github.com/petkaantonov/bluebird/) doesn't have a significant advantage anymore. Also, bluebird maintainers advice to switch to native promises, when possible. We should experiment with the switch, do some benchmarks and decide whether we should switch now.

If we do the switch, we don't need #524 at all.

puzpuzpuz:
Closes #529

### DIFF
--- a/src/network/ClientConnectionManager.ts
+++ b/src/network/ClientConnectionManager.ts
@@ -15,7 +15,6 @@
  */
 /** @ignore *//** */
 
-import * as Promise from 'bluebird';
 import {EventEmitter} from 'events';
 import {HazelcastClient} from '../HazelcastClient';
 import {
@@ -42,6 +41,7 @@ import {
     scheduleWithRepetition,
     shuffleArray,
     Task,
+    timedPromise
 } from '../util/Util';
 import {BasicSSLOptionsFactory} from '../connection/BasicSSLOptionsFactory';
 import {ILogger} from '../logging/ILogger';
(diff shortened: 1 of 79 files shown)


src/network/ClientConnectionManager.ts
```diff
@@ -632,39 +639,38 @@ export class ClientConnectionManager extends EventEmitter {
...
+                    case AuthenticationStatus.NOT_ALLOWED_IN_CLUSTER:
+                        error = new ClientNotAllowedInClusterError('Client is not allowed in the cluster');
+                        break;
+                    default:
+                        error = new AuthenticationError('Authentication status code not supported. Status: '
+                            + response.status);
                 }
-            });
+                connection.close('Failed to authenticate connection', error);
+                return Promise.reject(error);
```
mdumandag, line 671: Minor: Can be replaced with `throw error`

mdumandag (approved):
Approving with a minor comment

src/network/ClientConnectionManager.ts
puzpuzpuz: Good point. That return was in the original code and it didn't catch my eye. Addressed in 9994e713ce2c4fae6b1008b218e856f7b428a97e

puzpuzpuz:
verify

puzpuzpuz:
verify

PR #588, opened 2020-09-09
Revert single queue in PipelinedWriter

puzpuzpuz:
Depends on #587
A follow-up for #585

* Partially reverts `PipelinedWriter` to have a single queue array. To do so, updates `Util.copyBuffers` to handle arrays of objects instead of arrays of buffers.
  - The motivation is the following. This change allows to reduce the number of array copying done in `schedule()` by half, which should be beneficial for the hot path performance.
* Also fixes `ClientConnectionManager` shut down order, so that `HeartbeatManager` is shut down before connections are closed. This should fix flaky tests

#### Additional context

Approach with `Array.slice()`, which was introduced in #585, is significantly faster than the old shift-and-push approach. This may be seen in the following microbenchmark:
https://github.com/puzpuzpuz/microbenchmarks/blob/master/src/slice-vs-push.js

On Node.js v14.8.0 (Ubuntu 18.04) it gives the following output:
```
$ node src/slice-vs-push.js 
slice x 819,592 ops/sec ±1.42% (87 runs sampled)
shift and push x 166,645 ops/sec ±0.17% (96 runs sampled)
```

### DIFF
--- a/src/network/ClientConnection.ts
+++ b/src/network/ClientConnection.ts
@@ -30,7 +30,7 @@ import {
     SIZE_OF_FRAME_LENGTH_AND_FLAGS
 } from '../protocol/ClientMessage';
 
-const FROZEN_ARRAY = Object.freeze([]);
+const FROZEN_ARRAY = Object.freeze([]) as OutputQueueItem[];
 const PROPERTY_PIPELINING_ENABLED = 'hazelcast.client.autopipelining.enabled';
 const PROPERTY_PIPELINING_THRESHOLD = 'hazelcast.client.autopipelining.threshold.bytes';
 const PROPERTY_NO_DELAY = 'hazelcast.client.socket.no.delay';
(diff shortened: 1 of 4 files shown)


mdumandag (commented):
When I saw the #585, I thought we removed `OutputQueueItem` so that we don't create litter (an extra `OutputQueueItem` object) on each write request. Considering that we would call `schedule` less frequently than `write`, are we going to gain much from this change? Did you measure the difference?

puzpuzpuz:
> When I saw the #585, I thought we removed `OutputQueueItem` so that we don't create litter (an extra `OutputQueueItem` object) on each write request.

Litter is not that important here (it's "cheap" as the total allocation rate is relatively low) when compared with array copying (it gets "pricey" when the queue size gets bigger). Also, when we compare memory footprint of two arrays vs a single one with objects, it will be not that different.

> Considering that we would call `schedule` less frequently than `write`, are we going to gain much from this change? Did you measure the difference?

As the result of this change, we're going to be much closer to the throughput shown in the mentioned microbenchmark. Currently we literally reduce it by x2 as we have to slice arrays 4 times instead of 2 times.

mdumandag (approved):

PR #590, opened 2020-09-09
Implement redo operation option

### LINKED ISSUE #562: Implement support for redoOperation configuration option
Currently `network.redoOperation` option is documented, yet it's not used in the client implementation (see [this TODO](https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/config/ClientNetworkConfig.ts#L40)). We should implement it.

In case if this issue is moved to a later release, we should at least remove this option from the documentation.

puzpuzpuz:
* Closes #562
* Also migrates `ClientReconnectTest` to async/await syntax

### DIFF
--- a/test/ClientRedoEnabledTest.js
+++ b/test/ClientRedoEnabledTest.js
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2008-2020, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const chai = require('chai');
+chai.should();
+chai.use(require('chai-as-promised'));
+const expect = chai.expect;
+const RC = require('./RC');
+const {
+    Client,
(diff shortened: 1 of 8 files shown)


puzpuzpuz:
verify

puzpuzpuz:
verify

test/ClientRedoEnabledTest.js
```diff
@@ -0,0 +1,120 @@
...
+        cluster = await RC.createCluster(null, null);
+        await RC.startMember(cluster.id);
+    });
+
+    afterEach(async function () {
+        await client.shutdown();
+        return RC.terminateCluster(cluster.id);
+    });
+
+    it('List.get should throw on out of bounds index access', async function () {
```
mdumandag, line 46: Since we consider `redoOperation` only when we receive `TargetDisconnectedError`, what is the purpose of this and `Map.unlock should throw when not locked`?

test/config/configurations/full.json
```diff
@@ -27,6 +27,7 @@
             "127.0.0.9",
             "127.0.0.2:5702"
         ],
+        "redoOperation": true,
```
mdumandag, line 30: I guess we need to add this to `config-schema.json` also

test/ClientRedoEnabledTest.js
puzpuzpuz: These two tests were borrowed from https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/client/ClientRedoEnabledTest.java

And that test class was added to fix the following bug: https://github.com/hazelcast/hazelcast/issues/15157

So, I decided to keep them around to avoid running into the same bug in Node.js client.

test/config/configurations/full.json
puzpuzpuz: Good point. Added the definition in a23763524422f15b6f73f37af1ca99399fe3c126

mdumandag (approved):

puzpuzpuz:
verify

puzpuzpuz:
verify

Issue #581, opened 2020-09-04
[CP Subsystem] Add IAtomicReference and ICountDownLatch support


puzpuzpuz:
We didn't have AP versions of there data structures, so we need to implement them from a scratch.

PR #593, opened 2020-09-14
[CP Subsystem] Add AtomicReference support

puzpuzpuz:
Depends on #591

Closes: #581

* Adds `IAtomicReference` as the last part of CP Subsystem support in v4.0
* Includes code samples and documentation
* Also includes minor tsdoc, documentation and code style improvements

Corresponding client protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/342

Corresponding PRD: https://hazelcast.atlassian.net/wiki/spaces/PM/pages/1457030796/Node.js+Client+CP+Sub+System

### DIFF
--- a/src/proxy/cpsubsystem/AtomicRefProxy.ts
+++ b/src/proxy/cpsubsystem/AtomicRefProxy.ts
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2008-2020, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/** @ignore *//** */
+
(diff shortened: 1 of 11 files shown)


src/proxy/cpsubsystem/AtomicRefProxy.ts
```diff
@@ -0,0 +1,109 @@
...
+export class AtomicRefProxy<E> extends BaseCPProxy implements IAtomicReference<E> {
+
+    constructor(client: HazelcastClient,
+                groupId: RaftGroupId,
+                proxyName: string,
+                objectName: string) {
+        super(client, CPProxyManager.ATOMIC_REF_SERVICE, groupId, proxyName, objectName);
+    }
+
+    compareAndSet(expect: E, update: E): Promise<boolean> {
```
mdumandag, line 38: Does TypeScript supports representing different types with the same type parameter(`E`)? What I am trying to say is that we use `number` and `string` in the code samples(`code_samples/atomic_reference.js`) as parameters. If they must be the same type in the `.ts` code-bases, maybe we should update the code sample

puzpuzpuz: In JS code nothing prevents users from doing that. In TS I'd rather mimic the Java interface. If you're a TS user and you want to store `number` and `string` values under the same `AtomicReference`, then it's possible via `getAtomicReference<number | string>`. This call will give you a `AtomicReference<number | string>` which will accept numbers and strings, but nothing else.

puzpuzpuz: I misunderstood the question. The answer is yes, TS supports union types.

mdumandag (approved):

PR #596, opened 2020-09-14
Add warm-up phase for benchmarks

puzpuzpuz:
Improves the following aspects of built-in benchmarks:
* Adds mandatory warm-up phase for each benchmark type. The number of operations is set to 0.1 of the total
* Merges common code into a single entry point. Any benchmark can be run now via `benchmark/Benchmark.js` script
* Adds support for overriding main parameters of the benchmark
* Improves benchmark output to include more relevant information
* Replaces `put` benchmark with `set` benchmark, which should produce "more load" on the writer side
* Replaces `remove` with `delete` in `random` benchmark to avoid the cost of unnecessary old value deserialization

#### Example

```bash
$ node benchmark/Benchmark.js get -t 3000000 -c 256
Benchmark type: get
Value size: 1024
Starting warm-up for 300000 operations
Warm-up finished
Took 18.345385751 seconds for 3000000 operations
Ops/s: 163528.85901221627
Benchmark finished
```


puzpuzpuz:
verify

mdumandag (approved):

PR #599, opened 2020-09-16
Avoid wrapping single response param

### LINKED ISSUE #598: Get rid of the unnecessary "CodecResponseParams" object when it contains single item
This is what we do for Java and Python clients.

When the response of the codec contains a single item (as it is the case for the most used APIs such as `map.get`, `map.put` etc.), there is no need to wrap it in a `ResponseParams` object. We could directly return it.

That should result in slightly better performance for such APIs.

puzpuzpuz:
Closes #598

Corresponding client protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/344


src/proxy/cpsubsystem/AtomicLongProxy.ts
```diff
@@ -45,8 +45,7 @@ export class AtomicLongProxy extends BaseCPProxy implements IAtomicLong {
         }
         return this.encodeInvokeOnRandomTarget(AtomicLongAddAndGetCodec, this.groupId, this.objectName, delta)
             .then((clientMessage) => {
```
mdumandag, line 47: Minor: Can be replaced with `.then(AtomicLongAddAndGetCodec.decodeResponse)`

src/proxy/cpsubsystem/CPProxyManager.ts
```diff
@@ -107,8 +107,7 @@ export class CPProxyManager {
         const clientMessage = CPGroupCreateCPGroupCodec.encodeRequest(proxyName);
         return this.client.getInvocationService().invokeOnRandomTarget(clientMessage)
             .then((clientMessage) => {
-                const response = CPGroupCreateCPGroupCodec.decodeResponse(clientMessage);
-                return response.groupId;
+                return CPGroupCreateCPGroupCodec.decodeResponse(clientMessage);
```
mdumandag, line 110: Minor: Can be replaced with `.then(CPGroupCreateCPGroupCodec.decodeResponse)`

mdumandag (approved):
Approving with minor comments

src/proxy/cpsubsystem/AtomicLongProxy.ts
puzpuzpuz: Good catch. Fixed in 1ce0e2d1eaedae13ece036ae6db2089cfa6dd745

src/proxy/cpsubsystem/CPProxyManager.ts
puzpuzpuz: Good catch. Fixed in 1ce0e2d1eaedae13ece036ae6db2089cfa6dd745

mdumandag (approved):

Issue #559, opened 2020-07-22
Measure 4.0 vs 3.12 client performance


puzpuzpuz:
As 4.0 release includes a lot of changes in codes, `ClientMessage`/`Buffer` manipulations, and updates dependencies, prior to the release we need to run some benchmarks in order to detect performance degradation (if any).

As a part of this task we should also consider increasing default value of `hazelcast.client.autopipelining.threshold.bytes` option.

puzpuzpuz:
### Analysis of `bluebird` to native promises migration

New version - https://github.com/hazelcast/hazelcast-nodejs-client/commit/3cb18c8adc11ed596eaa66379c751e8f506fff93
Old version - https://github.com/hazelcast/hazelcast-nodejs-client/commit/bdc0939fadeb179d982c103f091911a1e3983a98

Scenario: `IMap#set` operation, 1M operations, concurrency level 128, 1KB string values, 15 runs (with warm-ups and `sleep 1` after each run)

Results:
```
$ cat compare-bluebird.csv | Rscript benchmark/compare.R
                                             confidence improvement accuracy (*)   (**)  (***)
 benchmark/Benchmark.js set t=1M s=1KB c=128                 3.25 %       ±3.81% ±5.20% ±7.03%

Be aware that when doing many comparisons the risk of a false-positive
result increases. In this case there are 1 comparisons, you can thus
expect the following amount of false-positive results:
  0.05 false positives, when considering a   5% risk acceptance (*, **, ***),
  0.01 false positives, when considering a   1% risk acceptance (**, ***),
  0.00 false positives, when considering a 0.1% risk acceptance (***)
```

![compare-bluebird-plot](https://user-images.githubusercontent.com/37772591/93432188-d7c1ae00-f8cd-11ea-84e9-4737da956b6e.png)

puzpuzpuz:
### v4.0 RC vs. v3.12 performance comparison

Cluster versions - 4.0.2, 3.12.9
Client "new" version - https://github.com/hazelcast/hazelcast-nodejs-client/commit/523ecf416780f84b915b3dc83831ad4a6ef09fb8 (PR #599)
Client "old" version - https://github.com/hazelcast/hazelcast-nodejs-client/commit/1db493efccdcbb3ea6cb912687309c0b356c1acd (latest `3.12.x`)

Environment:
Same as above

Scenario:
[...]

PR #605, opened 2020-09-21
Reduce number of allocated Buffers for writes

puzpuzpuz:
Refs: #559

Gets rid of almost all unnecessary `Buffer` allocations done in the client message write process. The performance impact of this change is around 8-10% in `map.set(42, 42)` benchmark.


mdumandag (approved):

puzpuzpuz:

PR #575, opened 2020-08-21
Add test to verify that member list is not broken after hot restart

puzpuzpuz:
Closes #564

* Adds `ClientHotRestartEventTest` to verify correct behavior of the client after cluster restart when Hot Restart is in use
* Updates RC version and thrift generated code to `0.8-SNAPSHOT`
* Improves code style in tests and `ClientConnectionManager`
* Migrates `AutoPipeliningDisabledTest` to async/await

PR #610, opened 2020-09-24
Fix member list broken by hot restart

### LINKED ISSUE #472: Client member list state gets broken with hot restart

puzpuzpuz:
* Closes #472
* Uses test from #502, which does not involve hot restart as `ClientHotRestartEventTest` test introduced in #575 is tricky to port to a non-ownerless client
* Updates RC's IMDG version to `3.12.9` (I had problems with downloading `3.12.10-SNAPSHOT`, so it's better to use the latest stable release)


src/invocation/ClusterService.ts
```diff
@@ -286,9 +286,55 @@ export class ClusterService {
...
+
+        for (const member of this.members) {
+            const idx = deletedMembers.findIndex(member.equals, member);
+            if (idx === -1) {
+                addedMembers.push(member);
+            } else {
+                deletedMembers = deletedMembers.splice(idx, 1);
+            }
+        }
+
```
mdumandag, line 313: In the Java client, I see a comment (`// removal events should be added before added events`). I am not sure about the reasoning behind that but maybe we could do the same here

puzpuzpuz: Yeah, there must be a good reason behind this particular order. Changed it in 13b65e8db2890eed207d4d88b34de2ff78a2b8e7

mdumandag (approved):

PR #624, opened 2020-10-02
Fix #535: Change Long into number for correlationId

### LINKED ISSUE #535: Use JS numbers instead of Long for correlation ids
JS number type supports integers up to `2^53 - 1` without precision loss. Currently we use `Long` type from a 3rd-party library for correlation ids, yet we should consider switching to plain numbers. The switch should help us to get rid of some litter on hot path.

Even if we increment correlation id sequence 1M times per second, it will only run into precision issues after over 285 years.

The main job here is to update accessor methods in `ClientMessage#getCorrelationId/#setCorrelationId`, so that so that a `Long` object is not created when reading/writing from/to the underlying buffer. Thus, the change should be to introduce a pair of methods near `FixSizedTypesCodec#decodeLong/#encodeLong` (they could be called `decodeNumberFromLong`/`encodeNumberAsLong` or smth like that). The arithmetical part of the implementation (low/high int calculation) could be based on a simplified version of what long.js library [does](https://github.com/dcodeIO/long.js/blob/master/src/long.js#L161) in `Long.fromNumber()` function.

anthonydahanne:
Closes #535

Changed correlationId from `Long`to `number`

### DIFF
--- a/src/codec/builtin/FixSizedTypesCodec.ts
+++ b/src/codec/builtin/FixSizedTypesCodec.ts
@@ -19,34 +19,10 @@ import {BitsUtil} from '../../util/BitsUtil';
 import {UUID} from '../../core/UUID';
 
 // Taken from long.js, https://github.com/dcodeIO/long.js/blob/master/src/long.js
-
-/**
- * @type {number}
- * @const
- * @inner
- */
-var TWO_PWR_16_DBL = 1 << 16;
-
-/**
- * @type {number}
- * @const
- * @inner
- */
-var TWO_PWR_32_DBL = TWO_PWR_16_DBL * TWO_PWR_16_DBL;
-
-/**
- * @type {number}
- * @const
- * @inner
- */
-var TWO_PWR_64_DBL = TWO_PWR_32_DBL * TWO_PWR_32_DBL;
-
-/**
- * @type {number}
- * @const
- * @inner
- */
-var TWO_PWR_63_DBL = TWO_PWR_64_DBL / 2;
+const TWO_PWR_16_DBL = 1 << 16;
+const TWO_PWR_32_DBL = TWO_PWR_16_DBL * TWO_PWR_16_DBL;
+const TWO_PWR_64_DBL = TWO_PWR_32_DBL * TWO_PWR_32_DBL;
+const TWO_PWR_63_DBL = TWO_PWR_64_DBL / 2;
 
 /** @internal */
 export class FixSizedTypesCodec {
@@ -73,11 +49,9 @@ export class FixSizedTypesCodec {
         return new Long(low, high);
     }
 
-    static encodeNumberAsLong(buffer: Buffer, offset: number, value: any): void {
-       if (value <= -TWO_PWR_63_DBL) {
-            // MIN_VALUE
-            buffer.writeInt32LE(0, offset);
-            buffer.writeInt32LE(0x80000000|0, offset + BitsUtil.INT_SIZE_IN_BYTES);
+    static encodeNonNegativeNumberAsLong(buffer: Buffer, offset: number, value: number): void {
+        if (value < 0) {
+            throw new Error("Only positive numbers are allowed in this method, received: " + value);
         }
 
         if (value + 1 >= TWO_PWR_63_DBL) {
--- a/src/codec/builtin/FixSizedTypesCodec.ts
+++ b/src/codec/builtin/FixSizedTypesCodec.ts
@@ -18,6 +18,36 @@ import * as Long from 'long';
 import {BitsUtil} from '../../util/BitsUtil';
 import {UUID} from '../../core/UUID';
 
+// Taken from long.js, https://github.com/dcodeIO/long.js/blob/master/src/long.js
+
+/**
+ * @type {number}
+ * @const
+ * @inner
+ */
+var TWO_PWR_16_DBL = 1 << 16;
+
+/**
+ * @type {number}
+ * @const
+ * @inner
+ */
+var TWO_PWR_32_DBL = TWO_PWR_16_DBL * TWO_PWR_16_DBL;
+
+/**
+ * @type {number}
+ * @const
+ * @inner
+ */
+var TWO_PWR_64_DBL = TWO_PWR_32_DBL * TWO_PWR_32_DBL;
+
+/**
+ * @type {number}
+ * @const
+ * @inner
+ */
+var TWO_PWR_63_DBL = TWO_PWR_64_DBL / 2;
+
 /** @internal */
 export class FixSizedTypesCodec {
     static encodeInt(buffer: Buffer, offset: number, value: number): void {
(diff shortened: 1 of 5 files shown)


nfrankel:
Welcome @anthonydahanne!

anthonydahanne:
tests seem to be stuck somehow...

puzpuzpuz:
Hi @anthonydahanne. Thanks for your contribution. I'm afraid that the description in #535 wasn't precise enough and the issue assumes changes different from what this PR has. Let me describe the changes in greater details.

The main job here is to update accessor methods in `ClientMessage#getCorrelationId/#setCorrelationId`, so that so that a `Long` object is not created when reading/writing from/to the underlying buffer. Thus, the change should be to introduce a pair of methods near `FixSizedTypesCodec#decodeLong/#encodeLong` (they could be called `decodeNumberFromLong`/`encodeNumberAsLong` or smth like that). The arithmetical part of the implementation (low/high int calculation) could be based on a simplified version of what long.js library [does](https://github.com/dcodeIO/long.js/blob/master/src/long.js#L161) in `Long.fromNumber()` function.

Hopefully, this makes the idea clear. Please let me know if you have any further questions.

anthonydahanne:
hello @puzpuzpuz !
OK, that makes sense; thanks for the explanation.
I'll let you know when I have a new version ready.
Meanwhile, I don't know if that is expected, but the jenkins check remained in this status "Expected — Waiting for status to be reported " and is not accessible, at least for me.

puzpuzpuz:
I'll check what's going on with Jenkins status once you push the update. Usually it works fine, but there might be some kind of a problem.

mdumandag:
About the Jenkins question, it is configured so that it will only run automatically when the PR is sent from someone in the Hazelcast organization. If it is a community PR like this, we have to write `verify` as a comment to trigger the job

puzpuzpuz:
Yes, my bad. Forgot about the security measure. Thanks @mdumandag

anthonydahanne:
Hello @puzpuzpuz ; I've pushed an update; please let me know if I'm in the right direction.
I think it should replace the usage of `long.js` for `correlationId`; the fact that `correlationId` can only be positive simplified the replacement.
Thanks

src/codec/builtin/FixSizedTypesCodec.ts
```diff
@@ -18,6 +18,36 @@ import * as Long from 'long';
...
 import {UUID} from '../../core/UUID';
 
+// Taken from long.js, https://github.com/dcodeIO/long.js/blob/master/src/long.js
+
+/**
+ * @type {number}
+ * @const
+ * @inner
+ */
+var TWO_PWR_16_DBL = 1 << 16;
```
puzpuzpuz, line 28: Please use `const` instead of `var` here and in the following constants.

src/codec/builtin/FixSizedTypesCodec.ts
```diff
@@ -18,6 +18,36 @@ import * as Long from 'long';
 import {BitsUtil} from '../../util/BitsUtil';
 import {UUID} from '../../core/UUID';
 
+// Taken from long.js, https://github.com/dcodeIO/long.js/blob/master/src/long.js
+
+/**
+ * @type {number}
```
puzpuzpuz, line 24: This comment looks redundant as it provides no additional information. Information for types is already available since we use TypeScript. The same applies to the following constants.

src/codec/builtin/FixSizedTypesCodec.ts
```diff
@@ -43,6 +73,29 @@ export class FixSizedTypesCodec {
         return new Long(low, high);
     }
 
+    static encodeNumberAsLong(buffer: Buffer, offset: number, value: any): void {
```
puzpuzpuz, line 76: `value` should be of `number` type.

puzpuzpuz, line 76: As this method assumes non-negative numbers only, it should be renamed into smth like `encodeNonNegativeNumberAsLong` (or smth like that) to avoid confusion in future. It might be also a good idea to validate the input and throw if it's negative.

Alternatively, this method should be updated to support negative numbers.

src/codec/builtin/FixSizedTypesCodec.ts
```diff
@@ -43,6 +73,29 @@ export class FixSizedTypesCodec {
...
+       if (value <= -TWO_PWR_63_DBL) {
```
puzpuzpuz, line 77: Do we need this check, considering that the input has to be non-negative?

src/codec/builtin/FixSizedTypesCodec.ts
```diff
@@ -43,6 +73,29 @@ export class FixSizedTypesCodec {
...
+            // MIN_VALUE
+            buffer.writeInt32LE(0, offset);
+            buffer.writeInt32LE(0x80000000|0, offset + BitsUtil.INT_SIZE_IN_BYTES);
+        }
+
+        if (value + 1 >= TWO_PWR_63_DBL) {
+            // MAX_VALUE
+            buffer.writeInt32LE(0xFFFFFFFF|0, offset);
```
puzpuzpuz, line 85: nit: this and the following `buf.writeInt32LE()` calls could be replaced with a single `buf.copy()` call if we would introduce a const `Buffer` which would hold the `MAX_VALUE`.

puzpuzpuz:
@anthonydahanne the changes look like the right direction. 👍

I've left some comments to mark further improvements. Could you also add some unit tests to make sure that the implementation is correct? We don't have a unit test for this codec, but the right path for the test should be `test/unit/codec/builtin/FixSizedTypesCodecTest.js`.

src/codec/builtin/FixSizedTypesCodec.ts
anthonydahanne: ok

src/codec/builtin/FixSizedTypesCodec.ts
anthonydahanne: yes

src/codec/builtin/FixSizedTypesCodec.ts
anthonydahanne: yes

src/codec/builtin/FixSizedTypesCodec.ts
anthonydahanne: nope, removing

src/codec/builtin/FixSizedTypesCodec.ts
anthonydahanne: ok, I chose to use `encodePositiveNumber` to not have a double negative in the name : "non" "negative"
I chose to throw `Error` also , please let me know if there is a more suitable error

puzpuzpuz: Non-negative is a bit more precise, as the method works fine for `0`. As for the error class, it's fine to use `Error` here, as users should never see this exception.

anthonydahanne: ok, good point

puzpuzpuz:
@anthonydahanne it looks like you're almost there with this PR and only a few things left to do. Please let me know if there is something I can help you with.

anthonydahanne:
hey there @puzpuzpuz - yes, almost there :-) 
I should be good to go, I'll have a look this week and let you know

puzpuzpuz:
> hey there @puzpuzpuz - yes, almost there :-) 
> I should be good to go, I'll have a look this week and let you know

Thanks for the update!

anthonydahanne:
hello @puzpuzpuz 
Here is the test; you'll see that for big value, the test is failing :-/
Not sure how to allocate the buffer properly... Could you please help me ?
Thanks

puzpuzpuz:
> Here is the test; you'll see that for big value, the test is failing :-/
> Not sure how to allocate the buffer properly... Could you please help me ?

There was a missing `return` in `encodeNonNegativeNumberAsLong`. I've created https://github.com/anthonydahanne/hazelcast-nodejs-client/pull/1 to address this issue. The PR also includes some minor improvements

anthonydahanne:
thanks for your PR, ; I merged it.
The comments and additional test made sense, and clarified for me some of the edge cases.

puzpuzpuz:
verify

puzpuzpuz:
verify

puzpuzpuz:
@anthonydahanne there is one special case I forgot about. We reset correlation id to `-1` when copying client messages. See https://github.com/anthonydahanne/hazelcast-nodejs-client/pull/2 for a proposed fix.

puzpuzpuz:
verify

puzpuzpuz (approved):
Nice job. Thanks for your contribution!

puzpuzpuz:
verify

PR #626, opened 2020-10-05
Improve listener deregistration

### LINKED ISSUE #613: Sync listener deregistration logic with Java client
Error handling and return result logic of `ListenerService#deregisterListener` method are different in Node.js and Java clients. As Java client's logic seems to be more consistent, we need to fix inconsistencies on Node.js client side.

Node.js client implementation:
https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/listener/ListenerService.ts#L162

Java client implementation:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/listener/ClientListenerServiceImpl.java#L306

puzpuzpuz:
* Closes #613 
* Also implements changes from the following Java client PR: https://github.com/hazelcast/hazelcast/pull/17646. As the result, `ListenerService#deregisterListener` immediately returns (i.e. returns a resolved promise) and the remote de-registrations are implicitly re-tried
* Also fixes unhandled promise rejections in `ListenerServiceTest`

### DIFF
--- a/src/listener/ListenerService.ts
+++ b/src/listener/ListenerService.ts
@@ -16,7 +16,12 @@
 /** @ignore *//** */
 
 import {HazelcastClient} from '../HazelcastClient';
-import {HazelcastError} from '../core';
+import {
+    ClientNotActiveError,
+    HazelcastError,
+    IOError,
+    TargetDisconnectedError
+} from '../core';
 import {ClientConnection} from '../network/ClientConnection';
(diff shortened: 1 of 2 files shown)


puzpuzpuz:
Failed test is not related with changes in this PR. I'll submit a PR to increase the timeout necessary to fix the flakiness in a separate PR.

mdumandag (approved):

Issue #635, opened 2020-10-07
Configure eslint for tests


puzpuzpuz:
We currently don't lint source code of tests which is not nice. The configuration should be placed in the `./test` subdir and `lint` script should be updated in `package.json`.

katieraby:
Can I take this one on please?

puzpuzpuz:
@katieraby sure. Let me assign it to you, if you don't mind.

katieraby:
Pull request for this issue: https://github.com/hazelcast/hazelcast-nodejs-client/pull/637 😄

PR #637, opened 2020-10-07
Configure ESLint for Tests

katieraby:
Referencing issue #635 

**TASKS UNDERTAKEN**
- Installed eslint-plugin-mocha 
- Added eslint config file to test directory
- Updated lint script to reflect the plugin and new config file
- Currently added all the recommended rules

The recommended rules currently implemented are: 
Screenshot 2020-10-07 at 20 27 43

Please let me know if you would like to be more specific in terms of targeting specific rules.

I ran the lint script with the new configurations, and here is a snippet of the terminal with some of the errors and warnings being returned: 

Screenshot 2020-10-07 at 21 00 08

Leaving the codebase with 337 problems (292 errors, 45 warnings)
31 errors and 0 warnings potentially fixable with the `--fix` option.


puzpuzpuz:
As a part of the issue we need to fix reported linting errors. The simplest approach would be to fix all code style/formatting issues and put disable comments into tests with more complex problems, like nested tests.

I can also see some `@typescript-eslint` errors in the linter output, which doesn't good. We should exclude typescript-eslint rules for tests, as they're ment for the main code base.

katieraby:
OK, will work on fixing the errors and excluding the `@typescript-eslint` errors.

katieraby:
- Fixed styling/formatting errors, or errors where consts were inside describe blocks when they didn't need to be etc.
- Removed a duplicate test
- Disabled line where only one linting error occurred per file
- Disabled file with larger linting errors such as nested funcs, and usage of mocha hooks for only one test

Note: 8 TS linting errors remain
Currently working on trying to exclude the test suite from @eslint/typescript

puzpuzpuz:
@katieraby I'm back from a short vacation and it looks like you're almost there with this PR. Please let me know if there is anything I can help you with.

katieraby:
> @katieraby I'm back from a short vacation and it looks like you're almost there with this PR. Please let me know if there is anything I can help you with.

Thanks @puzpuzpuz, I am struggling with excluding the test directory from typescript/eslint errors, and have tried a number of things including adding overrides to the eslintrc, creating a separate tconfig.eslint.json. I've thought about two different approaches which are: excluding all .js extension files from typescript linting (which encompasses all but one of the test files), or just trying to remove the test directory from typescript linting but not excluding it from eslinting recommended + mocha.

This is proving challenging, perhaps because it is a mixed TS/JS codebase, or there doesn't seem to be a coherent way to exclude a directory from only some linting (overrides didn't seem to work).

Do you have any suggestions or any ideas to try?

Thanks 🙂

puzpuzpuz:
> I've thought about two different approaches which are: excluding all .js extension files from typescript linting (which encompasses all but one of the test files), or just trying to remove the test directory from typescript linting but not excluding it from eslinting recommended + mocha.

I think it could be as simple as running TS linting against `./src` dir and JS linting against `./test` dir (we could include `./benchmark` and `./code_samples` dirs, but that's not a must have). This way we could simply use basic eslint CLI `eslint [options] file.js [file.js] [dir]` syntax in `package.json`'s lint script.

katieraby:
> > I've thought about two different approaches which are: excluding all .js extension files from typescript linting (which encompasses all but one of the test files), or just trying to remove the test directory from typescript linting but not excluding it from eslinting recommended + mocha.
> 
> I think it could be as simple as running TS linting against `./src` dir and JS linting against `./test` dir (we could include `./benchmark` and `./code_samples` dirs, but that's not a must have). This way we could simply use basic eslint CLI `eslint [options] file.js [file.js] [dir]` syntax in `package.json`'s lint script.

Sorry about the delay :) I am going to give this another go - the problem I was having was that the overarching ESlint config on the outside was taking precedent over the one in tests (and hence always using TS linting), but I'm now thinking that moving the outer .eslintrc into the src/ directory would solve this issue. I'm still not sure how to configure the eslint script to reflect both of these and the mocha plugin?

puzpuzpuz:
> I'm still not sure how to configure the eslint script to reflect both of these and the mocha plugin?

Not sure if I understand the problem. Could you describe it in more details?

puzpuzpuz:
Closing this one in favor of #745

PR #641, opened 2020-10-19
Release session on unexpected FencedLock error

puzpuzpuz:
Even if the calls to the lock, `lock` or `tryLock` fail for some reason other than `SessionExpiredError` or `WaitKeyCancelledError`, we were not releasing the sessions we acquired in the beginning, before throwing the exception. This fix aims to solve this problem by catching all kinds of exceptions (other than the two mentioned above), releasing the session and then rethrowing the catched exception.

Java client counterpart PR: https://github.com/hazelcast/hazelcast/pull/17697


mdumandag (approved):

puzpuzpuz:
verify

PR #654, opened 2020-10-28
Remove source maps from npm artifact

puzpuzpuz:
There is no point in having source maps in the npm artifact. Moreover, we already removed them in 4.0.


mdumandag (approved):

puzpuzpuz:

PR #662, opened 2020-11-09
Reject/retry pending invocations on connection close

### LINKED ISSUE #663: Pending invocations are not rejected or retried on connection close
When the connection is closed for a long running server-side operation, 3.12 client closes the connection, but does not properly reject/retry pending invocations for that connection.

This issue was the source of `LockProxyTest`'s flakiness. In this test [`lock.lock()`](https://github.com/hazelcast/hazelcast-nodejs-client/blob/3.12.x/test/lock/LockProxyTest.js#L175)'s promise was never resolved under the described conditions and the test was failing with a timeout.

puzpuzpuz:
Closes #663
Refs: #562

* Adds clean task for invocations with closed connections
* Adds support for `redoOperation` option
* Also fixes `generateKeyOwnedBy` function in `LockProxyTest` to deal with string keys. Previously it was generating number keys which was wrong as they were later converted to strings

### DIFF
--- a/src/invocation/InvocationService.ts
+++ b/src/invocation/InvocationService.ts
@@ -269,8 +269,9 @@ export class InvocationService {
 
         if (clientMessage.hasFlags(BitsUtil.LISTENER_FLAG)) {
             setImmediate(() => {
-                if (this.eventHandlers.has(correlationId)) {
-                    this.eventHandlers.get(correlationId).handler(clientMessage);
+                const invocation = this.eventHandlers.get(correlationId);
+                if (invocation !== undefined) {
+                    invocation.handler(clientMessage);
                 }
             });
             return;
--- a/src/invocation/InvocationService.ts
+++ b/src/invocation/InvocationService.ts
@@ -25,9 +25,15 @@ import {
     InvocationTimeoutError,
     IOError,
     RetryableHazelcastError,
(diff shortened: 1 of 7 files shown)


src/invocation/InvocationService.ts
```diff
@@ -222,24 +269,45 @@ export class InvocationService {
 
         if (clientMessage.hasFlags(BitsUtil.LISTENER_FLAG)) {
             setImmediate(() => {
-                if (this.eventHandlers[correlationId] !== undefined) {
-                    this.eventHandlers[correlationId].handler(clientMessage);
+                if (this.eventHandlers.has(correlationId)) {
```
mdumandag, line 272: Minor: We can try to get the handler first, and if it is not undefined, call it

mdumandag (approved):
Looks good to me. Approving with a minor comment. We should also update this line https://github.com/hazelcast/hazelcast-nodejs-client/blob/3.12.x/src/config/ClientNetworkConfig.ts#L49

src/invocation/InvocationService.ts
puzpuzpuz: Addressed in f255caaf6d8fc835c8f04eab96d7b8b59b58e4c3

puzpuzpuz:

PR #683, opened 2020-11-19
Fix floating promises

puzpuzpuz:
We had multiple places in the client where the result of a promise was completely ignored. This PR adds proper handlers to all of such places.

Also adds eslint rule to detect floating promises. Kudos go to @aigoncharov who pointed me at this awesome rule.

### DIFF
--- a/src/proxy/NearCachedMapProxy.ts
+++ b/src/proxy/NearCachedMapProxy.ts
@@ -40,9 +40,13 @@ export class NearCachedMapProxy<K, V> extends MapProxy<K, V> {
 
         this.nearCache = this.client.getNearCacheManager().getOrCreateNearCache(name);
         if (this.nearCache.isInvalidatedOnChange()) {
+            const logger = this.client.getLoggingService().getLogger();
             this.addNearCacheInvalidationListener().then((id) => {
                 this.invalidationListenerId = id;
                 this.nearCache.setReady();
+            }).catch((e) => {
+                logger.warn('NearCachedMapProxy', 'Failed to register Near Cache invalidation listener for '
+                    + name + ' map.', e);
             });
         } else {
             this.nearCache.setReady();
--- a/src/proxy/NearCachedMapProxy.ts
+++ b/src/proxy/NearCachedMapProxy.ts
@@ -47,6 +47,7 @@ export class NearCachedMapProxy<K, V> extends MapProxy<K, V> {
             }).catch((e) => {
                 logger.warn('NearCachedMapProxy', 'Failed to register Near Cache invalidation listener for '
                     + name + ' map.', e);
+                this.nearCache.setReady(e);
             });
         } else {
             this.nearCache.setReady();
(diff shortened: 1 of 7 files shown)


src/proxy/NearCachedMapProxy.ts
```diff
@@ -40,9 +40,13 @@ export class NearCachedMapProxy<K, V> extends MapProxy<K, V> {
 
         this.nearCache = this.client.getNearCacheManager().getOrCreateNearCache(name);
         if (this.nearCache.isInvalidatedOnChange()) {
+            const logger = this.client.getLoggingService().getLogger();
             this.addNearCacheInvalidationListener().then((id) => {
                 this.invalidationListenerId = id;
                 this.nearCache.setReady();
+            }).catch((e) => {
+                logger.warn('NearCachedMapProxy', 'Failed to register Near Cache invalidation listener for '
+                    + name + ' map.', e);
```
mdumandag, line 49: If we have an error here, should we reject the `ready` promise? If we don't do that, the next call to `get` will not be resolved forever since it waits for the `ready` promise. 

By the way, should the other methods in this proxy also wait for that promise?

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -64,15 +64,17 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
-            this.batchSize, this.serializationService, this.client.getLoggingService().getLogger(), this);
+            this.batchSize, this.serializationService, logger, this);
 
         this.runners[listenerId] = runner;
 
         this.ringbuffer.tailSequence().then((sequence: Long) => {
             runner.sequenceNumber = sequence.toNumber() + 1;
             runner.next();
+        }).catch((e) => {
+            logger.warn('ReliableTopicProxy', 'Failed to fetch sequence for runner.', e);
```
mdumandag, line 77: I am fine with that change, but it felt weird when I see that this method returns a string, instead of a promise of string.
We probably cannot change this due to backward compatibility, though.

src/proxy/NearCachedMapProxy.ts
puzpuzpuz: > If we have an error here, should we reject the `ready` promise? If we don't do that, the next call to `get` will not be resolved forever since it waits for the `ready` promise.

Good idea. Addressed in 1f24513adcb8c2ce67f355ae9a7478a6cc968923

> By the way, should the other methods in this proxy also wait for that promise?

As far as I understand, this deferred promise is required to postpone reads until the listener is ready, so that the chances that near cache is properly warmed up are higher. Writes in near cache are local as it's a write-behind cache and that's probably why the promise is ignored there.

In general, I'd prefer to wait for the listener in `client.getMap()` rather than in each individual operation. Such behavior should be less error-prone. Created #684 to track this enhancement.

mdumandag: Yes, makes sense

src/proxy/topic/ReliableTopicProxy.ts
puzpuzpuz: Yes, this implicit behavior doesn't look nice. Maybe we could introduce a safer method and deprecate this one. Created #685 to track this enhancement.

mdumandag (approved):

puzpuzpuz:
@mdumandag many thanks for your review!

Issue #616, opened 2020-09-30
Flaky ClientBackupAcksTest test [API-1201]


puzpuzpuz:
Got this failure in the PR builder on `4.0.x` branch: http://jenkins.hazelcast.com/job/NodeJS-pr-builder/613/testReport/junit/(root)/should%20receive%20backup%20acks%20in%20smart%20mode/ClientBackupAcksTest_should_receive_backup_acks_in_smart_mode/

```
should receive backup acks in smart mode.ClientBackupAcksTest should receive backup acks in smart mode (from ClientBackupAcksTest)
Error Message
Invocation 7 failed because of missed backup acks
Stacktrace
Error: Invocation 7 failed because of missed backup acks
    at new IndeterminateOperationStateError (lib/core/HazelcastError.js:18:12810)
    at Invocation.detectAndHandleBackupTimeout (lib/invocation/InvocationService.js:32:2819)
    at Timeout.Util_1.scheduleWithRepetition [as _onTimeout] (lib/invocation/InvocationService.js:35:2947)
```

This failure may be caused by load put on the CI server.

puzpuzpuz:
Raising priority for this one as it frequently shows up in CI runs

puzpuzpuz:
I think I found the source of the problem behind the test flakiness. There are two members in the test.

On 5702 member the backup operation fails on the following check:
https://github.com/hazelcast/hazelcast/blob/987c3574f7f8e15a7fcb8ef8042788f7c7f5e69a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java#L132

`validationFailure` contains the following:
```
[...]

PR #689, opened 2020-11-20
Fix client resource leak on failed start

### LINKED ISSUE #687: Client does not clean up resources on failed start
If one of client start steps fails, there may be active background tasks left, like `HeartbeatManager.task` or `Statistics.task`. The client should stop all scheduled tasks on a failed start.

puzpuzpuz:
Closes #687

Client was not stopping internal periodical tasks on failed start. As a result, those tasks kept working and the whole client instance was leaking. At least three client instances were leaking in our test suite.

### DIFF
--- a/src/HazelcastClient.ts
+++ b/src/HazelcastClient.ts
@@ -424,7 +424,7 @@ export class HazelcastClient {
     /** @internal */
     onClusterRestart(): void {
         this.getLoggingService().getLogger()
-            .info('HazelcastClient', 'Clearing local state of the client, because of a cluster restart');
+            .info('HazelcastClient', 'Clearing local state of the client, because of a cluster restart.');
         this.nearCacheManager.clearAllNearCaches();
         this.clusterService.clearMemberListVersion();
     }
@@ -443,7 +443,7 @@ export class HazelcastClient {
             this.clusterService.start(configuredMembershipListeners);
             this.clusterViewListenerService.start();
         } catch (e) {
-            logger.error('HazelcastClient', 'Client failed to start', e);
+            logger.error('HazelcastClient', 'Client failed to start.', e);
             throw e;
         }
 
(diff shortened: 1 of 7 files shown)


src/HazelcastClient.ts
```diff
@@ -469,7 +469,11 @@ export class HazelcastClient {
                 return this;
             })
             .catch((e) => {
-                logger.error('HazelcastClient', 'Client failed to start', e);
+                this.shutdown()
```
mdumandag, line 472: Should we wait for this shutdown call before throwing immediately?

test/ClientProxyTest.js
```diff
@@ -52,9 +52,9 @@ describe('ClientProxyTest', function () {
         assert.equal(mapProxy.getConnectedServerVersion(), -1);
     });
 
-    it('Client with a 3.7 server connection should return the version', function () {
+    it('client with a 4.7 server connection should return the version', function () {
```
mdumandag, line 55: Minor: `4.7 server` -> `4.1 server`

puzpuzpuz: Fixed. Thanks for spotting this.

src/HazelcastClient.ts
puzpuzpuz: There are two reasons why I implemented the shutdown-on-start-error in "async" way. First, `shutdown` call itself might take some time to finish. Second, `shutdown` may throw (reject to be more precise) and shadow the original error. It's possible to mitigate the at least the second problem, but I'm not sure if it's worth it. Do you see any benefit for the end user in "sync" approach?

puzpuzpuz:
verify

src/HazelcastClient.ts
mdumandag: Well, at first I thought the shutdown call should be a quick one. I thought the only thing that could take some time is the shutdown call of the connection manager. If it is possible for this to take some time, then it makes sense to keep things as it is.

puzpuzpuz: Yes, `CPSubsystem` may take some time to release sessions.

mdumandag: But, there should not be any sessions created during the startup

puzpuzpuz: Yes, I forgot to mention that the session clean up scenario is not currently possible, but that may change in future versions, as `shutdown` is async since 4.0.

mdumandag (approved):

puzpuzpuz:

puzpuzpuz:
Failed test is a known flaky test: #616

mdumandag (approved):

puzpuzpuz:
verify

puzpuzpuz:
verify

PR #698, opened 2020-12-01
Create pub-sub.js

recaifurkan:
simple pub-sub example

### DIFF
--- a/code_samples/pub-sub.js
+++ b/code_samples/pub-sub.js
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2008-2020, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


code_samples/pub-sub.js
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const { Client } = require('hazelcast-client');
+
+
+(async () => {
```
puzpuzpuz, line 21: It's better to use `code_samples/org-website/ReliableTopicSample.js` as the foundation for this code sample (with comments removed, just like in other `code_samples/*.js` files). Also, it would be great if you could add `await new Promise((resolve) => setTimeout(resolve, 1000));` into `ReliableTopicSample.js` within this PR as it seems to be not waiting for the message ATM.

puzpuzpuz: Forgot to mention that it's also good to rename this script to `reliable_topic.js`, so that the name is consistent with other code samples.

puzpuzpuz:
To speed up things, I'm closing this one in favor of #706

PR #704, opened 2020-12-08
Fix connection not immediately closed on socket error

### LINKED ISSUE #705: Connection is not immediately closed on socket error
Client connection is not determined correctly in the [`socket.on('error')` handler](https://github.com/hazelcast/hazelcast-nodejs-client/blob/65eb99317ac62fd3d8974e7a59854c7602bf6a8b/src/network/ClientConnectionManager.ts#L546) as the lookup was based on address. As connection's remote address gets updated on successful auth with the data received from the member, it can be different from the initial address used for connecting.

Another problem was that the handler does not close the connection, but only emits related events.

See https://github.com/hazelcast/hazelcast-nodejs-client/issues/666#issuecomment-740229446 and further discussion for more details

puzpuzpuz:
Fixes #705

As a result of the fix, the underlying connection in `socket.on('error')` handler gets closed reliable when an expected error occurs.

### DIFF
--- a/src/network/ClientConnectionManager.ts
+++ b/src/network/ClientConnectionManager.ts
@@ -264,6 +264,12 @@ export class ClientConnectionManager extends EventEmitter {
             })
             .then((socket) => {
                 clientConnection = new ClientConnection(this.client, translatedAddress, socket, this.connectionIdCounter++);
+                // close the connection proactively on errors
+                socket.on('error', (err: any) => {
+                    if (err.code === 'EPIPE' || err.code === 'ECONNRESET') {
+                        clientConnection.close('Connection closed by other side', err);
+                    }
+                });
                 return this.initiateCommunication(socket);
             })
             .then(() => clientConnection.registerResponseCallback(processResponseCallback))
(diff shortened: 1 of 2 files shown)


src/network/ClientConnectionManager.ts
```diff
@@ -264,6 +264,12 @@ export class ClientConnectionManager extends EventEmitter {
             })
             .then((socket) => {
                 clientConnection = new ClientConnection(this.client, translatedAddress, socket, this.connectionIdCounter++);
+                // close the connection proactively on errors
+                socket.on('error', (err: any) => {
+                    if (err.code === 'EPIPE' || err.code === 'ECONNRESET') {
```
mdumandag, line 269: I know it was like this before (I also didn't change this part), but should we close the connection here without checking the error code? The documentation of the `error` event says that `Emitted when an error occurs. The 'close' event will be called directly following this event.` So the question is after we receive an error event here(which is not `EPIPE` or 
 `ECONNRESET`, ), is it possible to continue operating?

puzpuzpuz: Yes, I had a similar question about that check, but decided to keep things as were. `git blame` tells me that it was like this since the initial error handling implementation and I failed to find a good reason for this check. So, I'm going to remove it. Thanks for bringing this question to the table.

puzpuzpuz: Addressed in 39c99306f6c2475ee53a95f4de9a06c1bcac2ce6

puzpuzpuz:
verify

mdumandag (approved):

PR #701, opened 2020-12-01
Update QuerySample.js

recaifurkan:
sql query is misspelled and valuesWithPredicate should be used instead of values ​​method.

PR #706, opened 2020-12-09
Fix incorrect code samples for query and topic

puzpuzpuz:
* Combines fixes from #698 and #701
* Also adds a new code sample for reliable topic


mdumandag (approved):

puzpuzpuz:
cc @burakcelebi as we need to update corresponding code samples on the org website.

Issue #666, opened 2020-11-12
Flaky ListenersOnReconnectTest


puzpuzpuz:
http://jenkins.hazelcast.com/job/NodeJS-10-windows-4.0.x/16/testReport/junit/kill%20two%20members%20[0,2],%20listener%20still%20receives%20map/put%20event%20[smart=false]/ListenersOnReconnectTest_kill_two_members__0_2___listener_still_receives_map_put_event__smart_false_/

```
Error Message
Heartbeat timed out
Stacktrace
Error: Heartbeat timed out
    at new TargetDisconnectedError (lib\core\HazelcastError.js:18:10274)
    at Timeout.Util_1.scheduleWithRepetition [as _onTimeout] (lib\invocation\InvocationService.js:35:2688)
```

aslatter:
This looks similar to an issue I just ran into - if I bounce Hazelcast, my Hazelcase-nodejs-client which is already connected waits ~1 minute to reconnect when we next try to use the connection, instead of reconnecting much more quickly.

In my debugging, the issue is that the socket-close handler isn't able to pro-actively clean up the connection.

ConnectionManager has this attempt at cleanup - when we set up the initial TLS connection we add an error-callback to the Node.js socket to proactively clean-up broken sockets:

```ts
        socket.on('error', (e: any) => {
            this.logger.warn('ConnectionManager', 'Could not connect to address ' + address.toString(), e);
            connectionResolver.reject(e);
            if (e.code === 'EPIPE' || e.code === 'ECONNRESET') {
                const connection = this.getConnectionFromAddress(address);
                if (connection != null) {
                    this.onConnectionClose(connection);
                }
            }
        });
```

However *getConnectionFromAddress* fails, because the 'address' we are using for lookup doesn't match the 'remoteAddress' stored in the 'ClientConnection' object.

I don't quite know why this is, but I suspect it's because we update the ClientConnection 'remoteAddress' with the response from authentication:

authenticateOnCluster -> handleSuccessfulAuth -> `connection.setRemoteAddress(response.address);`

In my case, the 'address' we are using for lookup is "localhost:port", but the 'remoteAddress' we fail to match to is "<my IP address>:port".

Let me know if you'd rather have this in a separate issue.

aslatter:
Ah - my example fails on Linux and Windows, if that's helful. I just noticed this issue has "Windows" in its name.

puzpuzpuz:
[...]

PR #716, opened 2020-12-14
Fix cluster reconnection race in ClientConnectionManager

puzpuzpuz:
Refs #666
Refs #717
(a 4.0.z backport will follow)

When an active connection is closed, `triggerClusterReconnection` execution may lead to races with `reconnectToMembersTask`. When a member restarts, as a possible outcome multiple connections to the same member might be open leading to duplicate event registrations and other anomalities. This fix adds necessary checks to prevent this from happening.


puzpuzpuz:
verify

mdumandag (approved):

puzpuzpuz:
@mdumandag many thanks for the quick review!

PR #719, opened 2020-12-15
Remove ad-hoc timeouts in tests

### LINKED ISSUE #452: Test Timeout
In tests timeouts seem random variables. It would be better to have some global timeout for tests.

puzpuzpuz:
Closes #452

* Removes ad-hoc timeouts in tests in favor of the global one
* Increases global test timeout to 2 mins
* Includes minor improvements for some tests

As a result, CI builds running on Windows should become more stable.

### DIFF
--- a/test/list/ListProxyTest.js
+++ b/test/list/ListProxyTest.js
@@ -19,170 +19,169 @@ const expect = require('chai').expect;
 const RC = require('./../RC');
 const { Client } = require('../../');
 const { ItemEventType } = require('../../lib/proxy/ItemListener');
+const { deferredPromise } = require('../../lib/util/Util');
 
 describe('ListProxyTest', function () {
 
     let cluster;
     let client;
-    let listInstance;
+    let list;
 
     before(async function () {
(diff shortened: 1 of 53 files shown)


test/list/ListProxyTest.js
```diff
@@ -269,16 +265,22 @@ describe('ListProxyTest', function () {
...
+                    expect(itemEvent.member).to.not.be.equal(null);
+                    itemRemovedTriggered.resolve();
+                } catch (err) {
+                    itemRemovedTriggered.reject(err);
+                }
             }
         });
         const removed = await listInstance.removeItemListener(registrationId);
         expect(removed).to.be.true;
+        await itemRemovedTriggered.promise;
```
mdumandag, line 284: We just register/deregister listener here, there should not be any waiting

puzpuzpuz: I've just noticed the problem and fixed the test. Could you take another look?

mdumandag (approved):

puzpuzpuz:

PR #721, opened 2020-12-15
Improve logging around socket errors

puzpuzpuz:
#709 introduced duplicate logs in the case of socket errors, so this PR is aimed to get rid of duplicates and improve logging

### DIFF
--- a/src/network/ClientConnection.ts
+++ b/src/network/ClientConnection.ts
@@ -486,10 +486,17 @@ export class ClientConnection {
         let message = this.toString() + ' closed. Reason: ';
         if (this.closedReason != null) {
             message += this.closedReason;
-        } else if (this.closedCause != null) {
-            message += this.closedCause.name + '[' + this.closedCause.message + ']';
         } else {
             message += 'Socket explicitly closed';
+        }
+
+        if (this.closedCause != null) {
+            message += ' - ';
+            const cause = this.closedCause as NodeJS.ErrnoException;
+            if (cause.code != null) {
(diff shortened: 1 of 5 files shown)


src/network/ClientConnection.ts
```diff
@@ -486,12 +486,19 @@ export class ClientConnection {
         let message = this.toString() + ' closed. Reason: ';
         if (this.closedReason != null) {
             message += this.closedReason;
-        } else if (this.closedCause != null) {
-            message += this.closedCause.name + '[' + this.closedCause.message + ']';
         } else {
             message += 'Socket explicitly closed';
         }
 
+        if (this.closedCause != null) {
```
mdumandag, line 493: Minor: If both `closedReason` and `closedCause` are not null, this is producing messages like `Reason: Failed to authenticate connection - Authentication failed. The configured cluster name..`. Maybe we could change those `connection.close` calls and make the produced log message better

mdumandag (approved):

src/network/ClientConnection.ts
puzpuzpuz: Good suggestion. Tried to improve log messages in c8d38f8e73898482b02f080482939adb1e759255

Please take another look.

mdumandag (approved):

src/network/ClientConnection.ts
mdumandag: Looks good, thanks

puzpuzpuz:

PR #735, opened 2020-12-24
Speed up client connecting to cluster

puzpuzpuz:
If a client was not configured explicit ports for the members, it is likely to try 5702 and 5703 as well. Leading to increased latency because connecting to a non existing server port takes time.

Also includes the following:
* Improve Cloud provider implementation which was not caching addresses on `loadAddresses` call
* Move Cloud discovery related tests into unit test dir
* Fix `markEnterprise` usage in `ClientFailoverTest`
* Introduce `AddressUtil`

Refs: https://github.com/hazelcast/hazelcast/pull/14249

PRD: https://hazelcast.atlassian.net/wiki/spaces/PM/pages/1192558858/Node.js+Client+Avoid+Random+Connection+Delay+When+Port+not+Specified


mdumandag (approved):

puzpuzpuz:
@mdumandag many thanks for the review!

Issue #737, opened 2020-12-25
Potential connection/socket leak in case of connection timeout


puzpuzpuz:
Connection manager doesn't not close the underlying connection (and socket) in case of connection timeout. This could potentially lead to socket leaks. The problem may occur in case of initial connection timeout (`triggerConnect` method) and further timeout (e.g. in `initiateCommunication` method).

puzpuzpuz:
Closed with #738

PR #738, opened 2020-12-25
Make sure to close socket on connection timeout

puzpuzpuz:
Closes: #737

Connection manager was not closing the underlying connection and socket in case of connection timeout. This could potentially lead to socket leaks.

Also includes the following:
* Fix request timeout handling in `HazelcastCloudDiscovery`. The [`timeout` event](https://nodejs.org/api/http.html#http_event_timeout) wasn't handled correctly
* Some minor improvements in tests.

### DIFF
--- a/src/network/ClientConnectionManager.ts
+++ b/src/network/ClientConnectionManager.ts
@@ -277,8 +277,15 @@ export class ClientConnectionManager extends EventEmitter {
         return timedPromise(
             connectionResolver.promise,
             this.connectionTimeoutMillis,
-            new HazelcastError(`Connection timed out to address ${address}.`)
-        ).finally(() => this.pendingConnections.delete(addressKey));
+            new HazelcastError('Connection timed out to address ' + address.toString())
+        ).catch((err) => {
+            // make sure to close connection on errors
+            if (clientConnection != null) {
+                clientConnection.close(null, err);
+            }
+            throw err;
+        })
+        .finally(() => this.pendingConnections.delete(addressKey));
     }
 
     getRandomConnection(): ClientConnection {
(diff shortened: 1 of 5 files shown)


src/network/ClientConnectionManager.ts
```diff
@@ -277,8 +277,15 @@ export class ClientConnectionManager extends EventEmitter {
         return timedPromise(
             connectionResolver.promise,
             this.connectionTimeoutMillis,
-            new HazelcastError(`Connection timed out to address ${address}.`)
-        ).finally(() => this.pendingConnections.delete(addressKey));
+            new HazelcastError('Connection timed out to address ' + address.toString())
```
mdumandag, line 280: Since the connection timeout logic is moved inside the `connectXXXSocket` functions, why do we still need a `timedPromise` here?

puzpuzpuz: That's because we still have other method invocations here where we do network comms without preserving the timeout. Examples of such methods are `translateAddress` (which may lead to a network call to the Hazelcast Cloud server) and `authenticateOnCluster`. That's why I decided to keep the per-`getOrConnect` timeout in place.

puzpuzpuz: WDYT?

mdumandag: I see, but wouldn't this be a problem for the authentication part. In there, we have 120 seconds of authentication timeout (same as the Java client), and because of this check, it might timeout earlier than that

puzpuzpuz: Yes, keeping things consistent with Java client certainly makes sense. Removed the `timedPromise` part in 15279db69c82dc23ce97e5a4f0aab289e2727555

puzpuzpuz:
@mdumandag I've addressed your comment and fixed an issue with `HazelcastCloudDiscovery`. It wasn't handling request as it's required by `https` module (the [`timeout` event](https://nodejs.org/api/http.html#http_event_timeout) wasn't handled correctly).

Could you check the last 3 commits?

mdumandag (approved):
Thanks for the changes, looks good

puzpuzpuz:
@mdumandag many thanks for the review!

PR #745, opened 2020-12-31
Add linting for tests and code samples

### LINKED ISSUE #635: Configure eslint for tests

puzpuzpuz:
Closes #635

Each source code directory (src, test, code_samples) now has its own eslint configuration.
    
Also includes the following:
* Fixes issues in tests and code samples reported by eslint
* Updates readme for code samples

### DIFF
--- a/code_samples/.eslintrc
+++ b/code_samples/.eslintrc
@@ -1,6 +1,5 @@
 {
     "env": {
-        "mocha": true,
         "es6": true,
         "node": true
     },
--- a/code_samples/.eslintrc
+++ b/code_samples/.eslintrc
@@ -1,0 +1,13 @@
+{
+    "env": {
+        "mocha": true,
+        "es6": true,
+        "node": true
+    },
+    "parserOptions": {
+        "ecmaVersion": 8
+    },
+    "extends": [
+        "eslint:recommended"
+    ]
+}
--- a/code_samples/org-website/GlobalSerializerSample.js
+++ b/code_samples/org-website/GlobalSerializerSample.js
@@ -22,12 +22,18 @@ class GlobalSerializer {
         this.id = 20;
     }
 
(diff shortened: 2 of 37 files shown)


puzpuzpuz:
@mdumandag I've rebased this one, so it's ready for a review.

code_samples/.eslintrc
```diff
@@ -0,0 +1,13 @@
+{
+    "env": {
+        "mocha": true,
```
mdumandag, line 3: Why do we need `mocha` in code samples?

code_samples/org-website/GlobalSerializerSample.js
```diff
@@ -22,12 +22,12 @@ class GlobalSerializer {
         this.id = 20;
     }
 
-    read(input) {
+    read(input) { // eslint-disable-line no-unused-vars
```
mdumandag, line 25: Assuming this code sample be visible in the `.org` site, should we display these ignore comments? I know that it makes not much sense but maybe we can use JSON.parse/stringify here to demonstrate the concept?

code_samples/.eslintrc
puzpuzpuz: Good catch. Addressed in 13cc25c0903a7d4d671c49bd1751eac20f1b161e

code_samples/org-website/GlobalSerializerSample.js
puzpuzpuz: Good suggestion. Change the sample in 13cc25c0903a7d4d671c49bd1751eac20f1b161e

mdumandag (approved):

puzpuzpuz:
@mdumandag thank you so much for the code review!

PR #746, opened 2021-01-11
Fix RepetitionTaskTest flakiness

puzpuzpuz:
Increases timeouts in order to fix the following `RepetitionTaskTest` flakiness on Windows CI machine:
http://jenkins.hazelcast.com/job/NodeJS-12-windows-master/71/testReport/junit/(root)/should%20be%20cancelled%20after%20interval/RepetitionTaskTest_should_be_cancelled_after_interval/
```
should be cancelled after interval.RepetitionTaskTest should be cancelled after interval (from RepetitionTaskTest)

Error Message
expected 1 to equal 2
Stacktrace
AssertionError: expected 1 to equal 2
    at Context.<anonymous> (test\unit\RepetitionTaskTest.js:58:31)
    at runNextTicks (internal/process/task_queues.js:62:5)
    at processTimers (internal/timers.js:489:9)
```

### DIFF
--- a/test/unit/RepetitionTaskTest.js
+++ b/test/unit/RepetitionTaskTest.js
@@ -52,7 +52,7 @@ describe('RepetitionTaskTest', function () {
         await TestUtil.promiseWaitMilliseconds(200);
         Util.cancelRepetitionTask(task);
         expect(counter).to.be.equal(2);
-        await TestUtil.promiseWaitMilliseconds(50);
+        await TestUtil.promiseWaitMilliseconds(75);
         expect(counter).to.be.equal(2);
     });
 
--- a/test/unit/RepetitionTaskTest.js
+++ b/test/unit/RepetitionTaskTest.js
@@ -23,45 +23,41 @@ describe('RepetitionTaskTest', function () {
 
     it('should be cancelled before timeout', async function () {
         let counter = 0;
-        const task = Util.scheduleWithRepetition(function () {
-            counter++;
-        }, 50, 75);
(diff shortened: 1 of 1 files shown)


test/unit/RepetitionTaskTest.js
```diff
@@ -23,45 +23,41 @@ describe('RepetitionTaskTest', function () {
...
-            counter++;
-        }, 50, 75);
+        const task = Util.scheduleWithRepetition(() => counter++, 50, 100);
 
-        await TestUtil.promiseWaitMilliseconds(130);
+        await TestUtil.promiseWaitMilliseconds(200);
         Util.cancelRepetitionTask(task);
         expect(counter).to.be.equal(2);
-        await TestUtil.promiseWaitMilliseconds(75);
+        await TestUtil.promiseWaitMilliseconds(50);
```
mdumandag, line 55: Minor: Should this be something larger than 50 ms(maybe 75ms) to make sure that the task is not run the third time? Because the total wait time is 250ms and the task is supposed to run three times after 250ms of its creation, so I am not sure whether or not this test would deterministically fail if there were some problems in the implementation

mdumandag (approved):

test/unit/RepetitionTaskTest.js
puzpuzpuz: Good suggestion. Changed the wait time to 75ms. See 6fdda4049efb1311921d549ed83840adfdba8eb9

PR #751, opened 2021-01-18
Support new metrics binary format for client stats

### LINKED ISSUE #556: Integrate with 4.0 client statistics
As the binary format of client statistics (metrics to be more precise) has changed in IMDG 4.0 (see [this class](https://github.com/hazelcast/hazelcast/blob/bc6923733a4cc2f485e093ba0fa3bccf16bd5bac/hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/MetricsCompressor.java) to learn the new format), we need either to implement support for it in Node.js client or disable this feature until the next minor release.

We should also consider introducing a new, simplified version of the format, so that the migration would be easier for all non-Java clients.

puzpuzpuz:
Closes #556

For backward-compatibility purposes, metrics are written into both client attributes and new metrics binary.

Binary generator core PR: https://github.com/hazelcast/hazelcast/pull/18066

Related MC issues:
- https://github.com/hazelcast/management-center/issues/4165
- https://github.com/hazelcast/management-center/issues/4166


src/statistics/Statistics.ts
```diff
@@ -167,31 +190,85 @@ export class Statistics {
...
+    }
 
-        for (const gaugeName in this.allGauges) {
-            const gaugeValueFunc = this.allGauges[gaugeName];
+    private addSimpleMetric(compressor: MetricsCompressor,
+                            metric: string,
+                            value: number,
+                            type: ValueType,
+                            unit?: ProbeUnit): void {
+        const dotIdx = metric.lastIndexOf('.');
```
mdumandag, line 231: Maybe not a big deal since the gauge names are pretty short, but maybe we can separate them in the registration part and get rid of this index checks are substring operations. `this.registerGauge('runtime.maxMemory', ...` -> `this.registerGauge('runtime', 'maxMemory', ...`

src/statistics/Statistics.ts
```diff
@@ -218,25 +295,79 @@ export class Statistics {
...
+                ValueType.LONG, ProbeUnit.COUNT);
+            this.addNearCacheMetric(stats, compressor,
+                'ownedEntryCount', name, nameWithPrefix, nearCacheStats.entryCount,
+                ValueType.LONG, ProbeUnit.COUNT);
+            this.addNearCacheMetric(stats, compressor,
+                'expirations', name, nameWithPrefix, nearCacheStats.expiredCount,
+                ValueType.LONG, ProbeUnit.COUNT);
+
+            // kept for backwards-compatibility purposes with Management Center 4.0
+            this.addEmptyAttribute(stats, 'lastPersistenceDuration', nameWithPrefix);
```
mdumandag, line 326: What happens when you don't send them? I see that in Python client, we don't send empty values for those below

src/statistics/MetricsCompressor.ts
```diff
@@ -0,0 +1,382 @@
...
+export enum ProbeUnit {
+
+    BYTES = 0,
+    MS = 1,
+    NS = 2,
+    PERCENT = 3,
+    COUNT = 4,
+    BOOLEAN = 5,
+    ENUM = 6,
+    US = 7
```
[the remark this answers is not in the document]
puzpuzpuz: I don't like the idea of porting dead code, so I left a comment. See dd2a2e094520d0c039e064720bc61ee5de8ec71c

src/statistics/MetricsCompressor.ts
```diff
@@ -0,0 +1,382 @@
...
+            this.metricsBuffer.writeInt(this.getDictionaryId(descriptor.metric));
+        }
+        if ((mask & MASK_DISCRIMINATOR) === 0) {
+            this.metricsBuffer.writeInt(this.getDictionaryId(descriptor.discriminator));
+        }
+        if ((mask & MASK_DISCRIMINATOR_VALUE) === 0) {
+            this.metricsBuffer.writeInt(this.getDictionaryId(descriptor.discriminatorValue));
+        }
+        if ((mask & MASK_UNIT) == 0) {
+            this.metricsBuffer.writeByte(descriptor.unit);
```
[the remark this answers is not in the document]
puzpuzpuz: Good finding. Initially I assumed the `unit` is always present, but found that it may be `null` in Java client's stats gauges. Forgot to update this part of the code accordingly. Fixed in dd2a2e094520d0c039e064720bc61ee5de8ec71c

src/statistics/MetricsCompressor.ts
```diff
@@ -0,0 +1,382 @@
...
+        }
+
+        // include excludedTargets and tags bytes for compatibility purposes
+        if ((mask & MASK_EXCLUDED_TARGETS) == 0) {
```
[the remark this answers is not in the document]
puzpuzpuz: Yes, we need them. That's for the very first iteration when `lastDescriptor` is `undefined` and then the `mask` is `0` (see `calculateDescriptorMask()`).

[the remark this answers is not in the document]
mdumandag: Oh sorry, I was wrong, we write them if their bits are not set

src/statistics/MetricsCompressor.ts
```diff
@@ -0,0 +1,382 @@
...
+    private writeDictionary(): void {
+        const words = this.dictionary.words();
+        this.dictionaryBuffer.writeInt(words.length);
+
+        let lastWordText = '';
+        for (const word of words) {
+            const wordText = word.word;
+            if (wordText.length > UNSIGNED_BYTE_MAX_VALUE) {
+                // this should have been checked earlier, this is a safety check
+                throw new Error('Dictionary element too long: ' + wordText);
```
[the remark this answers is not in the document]
puzpuzpuz: > Can this throw cause problems in our code? I guess nobody is catching it, will it terminate the process?

That depends on Node.js version and it's possible to override the default behavior. See https://nodejs.org/api/process.html#process_event_uncaughtexception

This error shouldn't be thrown and if it gets thrown, it's a bug. But I agree that we could make things slightly more safe here. Addressed in 61eb68eb0afd67d896880137fd39841c34446b99

src/statistics/Statistics.ts
puzpuzpuz: I'd prefer to keep them as is as current approach makes registration and gauge handling more readable. These `String.lastIndexOf()` should be pretty cheap especially when compared with the subsequent zlib compression.

src/statistics/Statistics.ts
puzpuzpuz: If we don't send them, MC won't receive them. But I didn't check how MC 4.0 behaves when they're not present. Going to do a quick check now.

src/statistics/Statistics.ts
mdumandag: Ok, sounds good to me

src/statistics/Statistics.ts
puzpuzpuz: I tried to remove all null gauges and empty near cache stats and so no issues in MC 4.0 apart from https://github.com/hazelcast/management-center/issues/4165 which is a known issue.

So, I removed all of these gauges and got rid of `addEmptyAttribute()` method. See bf6d04783b65fef8d92a2fa8bb723a9408415ae9

puzpuzpuz:
@mdumandag could you take another look at the PR? All comments seem to be addressed.

mdumandag (approved):
Thanks for the changes, it looks good to me

puzpuzpuz:

puzpuzpuz:
@mdumandag I've pushed the workaround fix discussed in https://github.com/hazelcast/management-center/issues/4165 to include it into this PR. Please check e1786f39bd7c94eb98088578a75be136b65cef18

mdumandag:
Looks good to me

PR #802, opened 2021-02-13
Replace utf with string in public API 

### LINKED ISSUE #801: [TRACKING ISSUE] Do not use deprecated methods of ObjectDataInput/Output, PortableRead…

See https://github.com/hazelcast/hazelcast/pull/18178 for details.

---

…er/Writer, and ClassDefinitionBuilder

We have deprecated methods containing `UTF` in their names and
introduced new methods that contain `String` instead of it in https://github.com/hazelcast/hazelcast/pull/18100.

This is a cleanup PR to use newly introduced methods instead of the
deprecated ones. The change is done automatically via IDE and
verified manually by going over the changes.

EE Counterpart: https://github.com/hazelcast/hazelcast-enterprise/pull/3939

srknzl:
Methods containing "utf" is replaced with "string", and old ones are deprecated.

Closes #801 and #792


srknzl:
verify

src/serialization/Data.ts
```diff
@@ -124,10 +124,20 @@ export interface DataOutput {
 
     writeShortArray(shorts: number[]): void;
 
+    /** @deprecated since version 4.1.
```
mdumandag, line 127: Since version should be 4.2, we already released 4.1. Also, it would be good if we can also link the suggested method so that users can easily navigate to it while reading the documentation.

Can you also update the other places?

src/serialization/ObjectData.ts
```diff
@@ -191,7 +191,14 @@ export class ObjectDataOutput implements DataOutput {
         this.writeArray(this.writeShort, shorts);
     }
 
+    /** @deprecated since version 4.1.
+     * This method will be deprecated in next major version. Please use writeString instead.
+     */
```
mdumandag, line 196: Since we mention the deprecation in the interface, I don't think this documentation is necessary. Same applies to other places

srknzl: Will it be removed? Maybe I mistyped it. `This method will be removed in next major version`?

mdumandag: In the next major version(5.0), we can remove it

srknzl: I deleted the `This method will be deprecated in next major version.` part. Is it fine without it too? I mean should we write to inform before removing?

mdumandag: No, I guess the warning should stay until we remove it

srknzl: Ok. hopefully, I handled your points now

src/serialization/Portable.ts
```diff
@@ -84,8 +91,13 @@ export interface PortableWriter {
 
     writeShortArray(fieldName: string, shorts: number[]): void;
 
+    /** @deprecated since version 4.1.
```
mdumandag, line 94: There are some leftover `4.1`s in this file

srknzl: Good catch 👍🏻 I went over all files but looks like missed these.

src/serialization/ObjectData.ts
```diff
@@ -191,7 +191,14 @@ export class ObjectDataOutput implements DataOutput {
         this.writeArray(this.writeShort, shorts);
     }
 
+    /** @deprecated since version 4.2.
```
mdumandag, line 194: Do we need these jsdocs? I mean, they are already there in the interface, I think duplicating it here in the implementation is unnecessary. The same applies to other places

srknzl: Ok I forget this sorry, I will remove them

srknzl: Well, it turns out that my IDE does not show deprecation by marking method as strike-through when I don't put deprecation to the class that implements the interface. However, this should be ok since the classes are internal and only we use them.

mdumandag: Same for me, when I explicitly declare the type with the interface, it shows it. But as you said, this is not a problem, even we don't use the concrete type apart from initializing the object.

src/serialization/portable/DefaultPortableWriter.ts
```diff
@@ -19,11 +19,11 @@ import {PortableSerializer} from './PortableSerializer';
 import {PositionalDataOutput} from '../Data';
 import {ClassDefinition, FieldDefinition} from './ClassDefinition';
 import {BitsUtil} from '../../util/BitsUtil';
-import {Portable, FieldType} from '../Portable';
+import {Portable, FieldType, PortableWriter} from '../Portable';
 import * as Long from 'long';
 
 /** @internal */
-export class DefaultPortableWriter {
+export class DefaultPortableWriter implements PortableWriter{
```
srknzl, line 26: Apart from deleting deprecation comments for classes, I noticed this one does not implement PortableWriter

mdumandag (approved):

Issue #803, opened 2021-02-18
Kill RC process after integration test execution


puzpuzpuz:
Currently Remote Controller process is kept alive after integration test suite passes. We need to make sure to kill the process once tests get executed or aborted.

srknzl:
Maybe writing the pid to a file helps

puzpuzpuz:
> Maybe writing the pid to a file helps

It could work. Or we could update the runner script which would run RC in a child process and make sure to kill it once tests pass or when a signal like SIGTERM is received.

srknzl:
> > Maybe writing the pid to a file helps
> 
> It could work. Or we could update the runner script which would run RC in a child process and make sure to kill it once tests pass or when a signal like SIGTERM is received.

Done! check #804

Issue #752, opened 2021-01-19
Rewrite scripts in Node.js


puzpuzpuz:
We have some scripts for downloading and running RC in `./scripts`. They're written as a bash script for *nix and batch file for Windows. To reduce the maintenance burden we could rewrite them in Node.js.

Refs: https://github.com/hazelcast/hazelcast-python-client/pull/299/

srknzl:
Handled both, could you check if I miss something @puzpuzpuz

PR #808, opened 2021-02-28
Eslint rule changes & linting

srknzl:
## Motivation
* To reduce reviewing overhead due to small code styling nits

## Changes:

* Now, a top level .eslintrc is used for all codes excepts for src folder. If any folder needs more config, a config file is created in that folder extending root config. 
* add rules to root config since eslint: recommended does not include all we want.
* linting according to new rules


.eslintignore
```diff
@@ -0,0 +1,3 @@
+src/.eslintrc.js
+scripts/.eslintrc.js
+src/codec/**/*.ts
```
srknzl, line 3: codecs are autogenerated as I know, so ignore them. (They contains double quotes)

scripts/.eslintrc.js
```diff
@@ -0,0 +1,95 @@
...
+        'eol-last': 'error',
+        'eqeqeq': 'warn',
+        'func-call-spacing': 'warn',
+        'indent': 'warn',
+        'keyword-spacing': [
+            'error',
+            {
+                'after': true,
+                'before': true
+            }
```
srknzl, line 62: These rules are based on `test-runner.js` (eslint's auto detect feature), then I removed the ones that are too specific. Basically, I tried to keep the ones @puzpuzpuz asked me to refactor in prs #752 and #803

scripts/download-rc.js
```diff
@@ -43,7 +43,8 @@ const downloadRC = () => {
     if (fs.existsSync(`hazelcast-remote-controller-${HAZELCAST_RC_VERSION}.jar`)) {
         console.log('remote controller already exists, not downloading from maven.');
     } else {
-        console.log(`Downloading: remote-controller jar com.hazelcast:hazelcast-remote-controller:${HAZELCAST_RC_VERSION}`);
+        console.log('Downloading: remote-controller jar com.hazelcast:hazelcast-remote-controller:'
```
srknzl, line 46: the line was too long

scripts/test-runner.js
```diff
@@ -3,6 +3,7 @@ const fs = require('fs');
 const os = require('os');
 const net = require('net');
 const {spawnSync, spawn} = require('child_process');
+
```
srknzl, line 6: I generally separate relative imports and standard module imports by a newline

scripts/test-runner.js
```diff
@@ -48,7 +49,8 @@ const isAddressReachable = (host, port, timeoutMs) => {
             resolve(true);
         });
     });
-}
+};
+
```
srknzl, line 53: I think the code is more readable with a newline between function declarations

scripts/test-runner.js
```diff
@@ -57,67 +59,61 @@ const startRC = async (background) => {
                 'start /min "hazelcast-remote-controller" cmd /c '
               + `java -Dhazelcast.enterprise.license.key=${HAZELCAST_ENTERPRISE_KEY} -cp ${CLASSPATH} `
               + 'com.hazelcast.remotecontroller.Main --use-simple-server> rc_stdout.txt 2>rc_stderr.txt', [], {
-                stdio: 'inherit',
-                shell: true
-            });
+                    stdio: 'inherit',
```
srknzl, line 62: due to eslint's indentation warning (4 spaces for every level deeper)

scripts/test-runner.js
```diff
@@ -57,67 +59,61 @@ const startRC = async (background) => {
...
+                    shell: true
+                }
+            );
         } else {
-            try {
```
srknzl, line 64: this is due to eslint warning about unnecessary try catch statement

scripts/download-rc.js
```diff
@@ -43,7 +43,8 @@ const downloadRC = () => {
...
+        + HAZELCAST_RC_VERSION);
```
puzpuzpuz, line 47:
```suggestion
            + HAZELCAST_RC_VERSION);
```

scripts/.eslintrc.js
```diff
@@ -0,0 +1,95 @@
...
+    'env': {
+        'commonjs': true,
+        'node': true,
+        'es6': true
+    },
+    'extends': 'eslint:recommended',
+    'parserOptions': {
+        'ecmaVersion': 12
+    },
+    'rules': {
```
puzpuzpuz, line 27: Do we need all of these customizations for eslint rules? For tests and code samples we simply use `eslint:recommended`:
https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/test/.eslintrc
https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/code_samples/.eslintrc

srknzl: You can check this page to see which rules are in eslint:recommended. https://eslint.org/docs/rules/
In my first look what I see is, the most important rules such as quotes, semicolons, max-len and no-multiple-empty-lines are not included. No need to mention others.

scripts/download-rc.js
srknzl: I guess we should put a rule even for this if any. And I bet there is one. The goal of this pr is to detect these nits by linter.

scripts/.eslintrc.js
puzpuzpuz: Thanks for the clarification. Then it would be nice to include these changes into configs for tests and code samples. Even better would be if we could somehow share the config to avoid duplication (the only caveat is tests which require mocha plugin).

srknzl: Maybe we could do it with "extends", I will check it out

srknzl: I did this extension mechanism. You can check it out

.eslintrc
```diff
@@ -0,0 +1,78 @@
+{
+    "env": {
+        "commonjs": true,
+        "node": true,
+        "es2021": true
```
srknzl, line 5: As stated in [here](https://eslint.org/docs/user-guide/configuring/language-options#specifying-environments), putting es2021: true here enables ecmaVersion:12, likewise es6 enables ecmaVersion: 6. So, I think no need to write ecmaVersion explicitly. Btw, should we use es2021 and why? I thought these ecma versions are backward compatible and used the latest one.

.eslintrc
```diff
@@ -0,0 +1,78 @@
...
+    },
+    "extends": "eslint:recommended",
+    "rules": {
+        "arrow-spacing": [
```
srknzl, line 9: I deleted some of the rules and added no-useless-concat. Could you check each rule if they are ok? We can delete ones that are unnecessary.

scripts/download-rc.js
srknzl: Could not find a rule for this. Also a rule that only forbids spaces in 
```
import { a} from b;
```

scripts/test-runner.js
srknzl: Deleted `indent` rule

src/.eslintrc.js
```diff
@@ -40,6 +40,7 @@ module.exports = {
         '@typescript-eslint/no-unused-vars': [ 'error', { args: 'none' } ],
         'no-prototype-builtins': 'off',
         'prefer-rest-params': 'off',
-        'max-len': [ 'error', 130, 4 ]
+        'max-len': [ 'error', 130, 4 ],
+        'quotes': ['error', 'single'],
```
srknzl, line 44: note that linter config in src/ does not extend root conf,g file. If we did that, we would have to change a lot in src/ folder.  We could add some of them like we added max-len and quotes above. Let me know if we will add any other rules here as well.

test/.eslintrc
```diff
@@ -1,21 +1,18 @@
 {
     "env": {
-        "mocha": true,
-        "es6": true,
```
srknzl, line 4: these are extended

test/.eslintrc
```diff
@@ -1,21 +1,18 @@
...
     "extends": [
-        "eslint:recommended",
+        "../.eslintrc",
         "plugin:mocha/recommended"
     ],
     "rules": {
         "mocha/no-hooks-for-single-case": "off",
-        "mocha/no-setup-in-describe": "off"
+        "mocha/no-setup-in-describe": "off",
+        "camelcase": "off",
```
srknzl, line 15: two of the test files has non camelcase variables, so I disabled camelcase in test folder

test/.eslintrc
```diff
@@ -1,21 +1,18 @@
...
+        "max-len": ["warn", 130]
```
srknzl, line 16: due to test names lines are long in test folder. made the line limit 130

test/TestUtil.js
```diff
@@ -129,7 +129,7 @@ exports.markServerVersionAtLeast = function (_this, client, expectedVersion) {
     let actNumber;
     if (process.env['SERVER_VERSION']) {
         actNumber = BuildInfo.calculateServerVersionFromString(process.env['SERVER_VERSION']);
-    } else if (client != null) {
+    } else if (client !== null) {
```
srknzl, line 132: writing this comment for all changes in test/**/*.js

* all changes except == ----> ===, != ---> !== and line splitting due to long line are done automatically by eslint.

test/unit/serialization/PortableObjects.js
```diff
@@ -132,9 +132,9 @@ class PortableObjectV2{
         this.a_short = a_short;
         this.an_integer = an_integer;
```
srknzl, line 133: one of the non camelcase variables

test/integration/ClientShutdownTest.js
```diff
@@ -36,9 +36,9 @@ describe('ClientShutdownTest', function () {
 
     afterEach(async function () {
         sandbox.restore();
-        if (cluster !== null) {
+        if (cluster !== undefined) {
```
srknzl, line 39: I changed these values from null to undefined to fix test fails due to equality operator change([eqeqeq](https://eslint.org/docs/rules/eqeqeq) rule)

.eslintrc
puzpuzpuz: I'd prefer to keep using ES6 syntax, as Node.js 10 and 12 do not support all ES2021 features: https://node.green/#ES2020. The same consideration applies to ES2017+

src/.eslintrc.js
puzpuzpuz: Yes, that's fine. Our rules for TS code are slightly different from JS ones.

test/integration/map/MapProxyTest.js
```diff
@@ -559,26 +559,28 @@ describe('MapProxyTest', function () {
...
-                            expect(entryEvent.name).to.equal('test');
-                            expect(entryEvent.key).to.equal('key10');
-                            expect(entryEvent.value).to.equal('val10');
-                            expect(entryEvent.mergingValue).to.be.null;
-                            expect(entryEvent.member).to.not.be.null;
-                            done();
-                        } catch (err) {
-                            done(err);
+            it('addEntryListener on key, entryAdded fires because predicate returns true for that entry, inlVal=yes',
+            function (done) {
```
puzpuzpuz, line 563: nit: I'd prefer to add padding here:
```suggestion
                function (done) {
```

test/integration/nearcache/MigratedDataTest.js
```diff
@@ -89,7 +89,8 @@ describe('MigratedDataTest', function () {
         await RC.terminateCluster(cluster.id);
     });
 
-    it('killing a server migrates data to the other node, migrated data has new uuid, near cache discards data with old uuid', async function () {
+    it('killing a server migrates data to the other node, migrated data has new uuid,'
```
puzpuzpuz, line 92:
```suggestion
    it('killing a server migrates data to the other node, migrated data has new uuid, '
```

test/unit/serialization/BinaryCompatibilityTest.js
```diff
@@ -145,12 +145,14 @@ describe('BinaryCompatibilityTest', function () {
...
 
                     if (!ReferenceObjects.skipOnSerialize[varName]) {
-                        it(varName + '-' + convertEndiannesToByteOrder(isBigEndian) + '-' + version + ' serialize deserialize', function () {
-                            const service = createSerializationService(isBigEndian, stripArticle(varName).toLowerCase());
-                            const data = service.toData(object);
-                            const deserialized = service.toObject(data);
-                            expectAlmostEqual(deserialized, object);
-                        });
+                        it(varName + '-' + convertEndiannesToByteOrder(isBigEndian) + '-' + version + ' serialize deserialize',
+                        function () {
```
puzpuzpuz, line 149:
```suggestion
                            function () {
```

test/unit/serialization/portable/ClassDefinitionTest.js
```diff
@@ -61,9 +60,11 @@ describe('ClassDefinitionTest', function () {
         expect(a.getFieldById.bind(a, 0)).to.not.throw(Error);
     });
 
-    it('getFieldById():When index is more than or equal to total number of fields it should throw an range exception.', function () {
-        expect(a.getFieldById.bind(a, 3)).to.throw(Error);
-    });
+    it('getFieldById():When index is more than or equal to total number of fields it should throw an range exception.',
+    function () {
```
puzpuzpuz, line 64:
```suggestion
        function () {
```

puzpuzpuz (commented):
Good job! The only concern I have is about ES2021 syntax in eslint. Other comments are nits.

srknzl:
Can you write rules for changes you suggested

puzpuzpuz:
> Can you write rules for changes you suggested if applicable. We can add it to eslint

The indent ones should be something like `/*eslint indent: ["error", 2, { "FunctionDeclaration": {"body": 1, "parameters": 1} }]*/`
https://eslint.org/docs/rules/indent

But I didn't test whether it works.

srknzl:
nope, the related indent config here is 

/*eslint indent: ["error", 2, { "CallExpression": {"arguments": 1} }]

since it() is a function call.

.eslintrc
srknzl: Actually, by setting ecmaVersion to 8 we were using es2017 instead of es6 before this pr. Setting es6: true makes ecmaVersion:7, then we make it 8 by reassigning it. For example async is added in es2017 which we used in tests, code samples, scripts. 

This auto setting is explained at https://eslint.org/docs/user-guide/configuring/language-options#specifying-environments

Thus, setting es2017: true which sets ecmaVersion to 8 automatically.

.eslintrc
```diff
@@ -73,6 +73,24 @@
                 "max": 1
             }
         ],
-        "no-useless-concat": "error"
+        "no-useless-concat": "error",
+        "indent": ["warn", 4, {
+            "CallExpression": {
+                "arguments": 1
+            },
+            "SwitchCase": 1,
```
srknzl, line 81: SwitchCase cannot be set to "off" to disable linting about switch cases statements. I did not understand why. Someone proposed being able to set it to off [here]
(https://github.com/eslint/eslint/issues/7270) and developers said `it is good idea we won't be able to do it`. 

Similarly, in FunctionDecleration and FunctionExpression, "body" param is by default 1 and they cannot be set to "off" as well. If I set them to off, linter gives parsing error. I guess all of the three options are fine with the value "1". Body options are omitted since their default is 1. SwitchCase is not omitted and explicitly set to 1 since SwitchCase's default value is 0 and it is different from our current switch case indentation level which is 1. 

With this eslint configuration and last changes included, no linting errors left in all folders.

 Addressed your points in last review using "indent" rule. Plus, [there is one more](https://github.com/hazelcast/hazelcast-nodejs-client/pull/808/commits/053c3c4a7e7534a48ecbe534ea257d69fddf703c#diff-c13103958dde39fbec06f98103bd3387a6c127f5e3bc287d7a5fe42b4aaa3b53L77) which is due to 
```
{
    "FunctionExpression": {
         "body" : 1
     }
}
```
rule.

.eslintrc
```diff
@@ -0,0 +1,97 @@
...
+            "error",
+            "last"
+        ],
+        "space-before-blocks": "warn",
+        "spaced-comment": [
+            "warn",
+            "always"
+        ],
+        "no-trailing-spaces": "warn",
+        "no-multi-spaces": "warn",
```
srknzl, line 70: thought this would be good to add. https://eslint.org/docs/rules/no-multi-spaces

In the commit https://github.com/hazelcast/hazelcast-nodejs-client/pull/808/commits/bbe9a22ff5f7f3024ab493ce0400355aacb9269b there are some fixes regarding this as well.

scripts/download-rc.js
srknzl: Ok, changing this manually. Could not find a rule. I guess it does not exist. (I searched for aligning multiline string concatenations)

srknzl:
At first, due to eqeqeq rule I added, I converted == to === and != to !== in this pr. But at some point this change started making tests fail due to undefined and null comparison etc. I decided to remove eqeqeq rule from eslint and reverted all my changes regarding eqeqeq

.eslintrc
srknzl: I asked them in an issue see

https://github.com/eslint/eslint/issues/14167.

Good thing is we can actually configure that by ignoreNodes. I'll handle that

srknzl: After adding ignoreNode, I can now revert the change https://github.com/hazelcast/hazelcast-nodejs-client/pull/808/commits/053c3c4a7e7534a48ecbe534ea257d69fddf703c#diff-c13103958dde39fbec06f98103bd3387a6c127f5e3bc287d7a5fe42b4aaa3b53L77 which was due to implicit
```
{
    "FunctionExpression": {
         "body" : 1
     }
}
``` 
option of indent rule that I thought I could not disable.

.eslintrc
puzpuzpuz: Makes sense. Let's use `es2017: true` then.

.eslintrc
puzpuzpuz: Yes, that's a good addition.

puzpuzpuz:
> Does our build process check for linting issues? In other words does it run `npm run lint` and check for errors?

Yes, it does. Let's keep the rule removed for now. As for the `if (foo){}` check, I prefer to check for the exact type or `undefined/null`.

`foo` will be evaluated to `false` for many non-null values like `1` or `""` which is not we want to achieve with such checks.

test/unit/serialization/PortableObjects.js
puzpuzpuz: Let's leave them as is for now.

srknzl:
All tests passed on my local + no linting errors. Are we all set?

puzpuzpuz:
verify

puzpuzpuz:
> All tests passed on my local + no linting errors and warnings. Are we all set?

Yes, as soon as CI passes, I'm merging this one.

puzpuzpuz (approved):

PR #811, opened 2021-02-28
Using root level hooks to destroy rc connection

srknzl:
* This change is my suggestion to deal with
  * mocha will hang if we do not use --exit option. (right now it is used mocha config file.) As in the link below, using --exit is not necessarily recommended. 
  * construction of rc controller every time `RC.js` required. 
  * possible resource leak
* The controller is contructed once in beforeAll hook and its connection is destroyed on afterAll hook.
* Background on [--exit](https://mochajs.org/#-exit) 
* Background on [global root hooks I used](https://mochajs.org/#root-hook-plugins) 

Update:

* **small problem:** If I run small portion of tests with a pattern, or no test runs, mocha exits without hanging. However, when I call all the integration tests, it hangs at the end.
* They are probably due to open sockets opened somewhere and not closed. Wtfnode output when all tests passed and hanging started:  

```
[WTF Node?] open handles:
- File descriptors: (note: stdio always exists)
  - fd 2 (tty) (stdio)
  - fd 1 (tty) (stdio)
- Sockets:
  - 127.0.0.1:46138 -> 127.0.0.1:9999
  - ::ffff:127.0.0.1:9999 -> ::ffff:127.0.0.1:46138
  - 127.0.0.1:46184 -> 127.0.0.1:9999
  - ::ffff:127.0.0.1:9999 -> ::ffff:127.0.0.1:46184
  - 127.0.0.1:46190 -> 127.0.0.1:9999
  - ::ffff:127.0.0.1:9999 -> ::ffff:127.0.0.1:46190

```

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,6 +1,5 @@
 {
     "timeout": 120000,
-    "exit": true,
     "exclude": [
         "test/soak_test/**",
         "test/user_code/**"
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -3,5 +3,6 @@
     "exclude": [
         "test/soak_test/**",
         "test/user_code/**"
-    ]
+    ],
+    "require": "test/hooks.js"
 }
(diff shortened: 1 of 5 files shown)


srknzl:
I ran integration test in master and there is only one connection during test to localhost:9701 contrary to what I guessed. Also, --exit seems to clean everything including socket servers used in ClientConnectionManagerTest integration test. (port 9999) and cluster related connections(5701 and 9701). I guess we do not need this pr.

PR #816, opened 2021-03-12
Bump typescript version to 4

srknzl:
This pr aims to update typescript version to be 4.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
     "main": "lib/index.js",
     "types": "lib/index.d.ts",
     "dependencies": {
-        "long": "4.0.0",
-        "@types/long": "4.0.1"
+        "@types/long": "4.0.1",
+        "long": "4.0.0"
     },
     "devDependencies": {
         "@types/node": "^10.17.28",
@@ -26,7 +26,7 @@
         "sinon-chai": "^3.5.0",
         "thrift": "^0.12.0",
         "typedoc": "^0.18.0",
-        "typescript": "^3.9.7",
+        "typescript": "^4.2.3",
         "winston": "^3.3.3",
         "yargs": "^16.0.3"
     },
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     },
     "devDependencies": {
         "@types/node": "^10.17.28",
-        "@typescript-eslint/eslint-plugin": "^3.10.0",
-        "@typescript-eslint/parser": "^3.10.0",
+        "@typescript-eslint/eslint-plugin": "^4.17.0",
+        "@typescript-eslint/parser": "^4.17.0",
         "chai": "^4.2.0",
         "chai-as-promised": "^7.1.1",
         "eslint": "^7.7.0",
--- a/src/aggregation/Aggregator.ts
+++ b/src/aggregation/Aggregator.ts
@@ -22,6 +22,7 @@ import * as AggregatorFactory from './AggregatorFactory';
 /**
  * Base interface for all aggregators.
  */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export interface Aggregator<R> extends IdentifiedDataSerializable {
 }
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -5,8 +5,8 @@
     "main": "lib/index.js",
     "types": "lib/index.d.ts",
     "dependencies": {
-        "long": "4.0.0",
-        "@types/long": "4.0.1"
+        "@types/long": "4.0.1",
```
srknzl, line 8: These dependencies' order changes every time I run npm install. So I've kept it like this.

package.json
```diff
@@ -5,13 +5,13 @@
...
+        "long": "4.0.0"
     },
     "devDependencies": {
         "@types/node": "^10.17.28",
-        "@typescript-eslint/eslint-plugin": "^3.10.0",
-        "@typescript-eslint/parser": "^3.10.0",
+        "@typescript-eslint/eslint-plugin": "^4.17.0",
```
srknzl, line 13: Since I've updated typescript, the eslint gave a warning about the new typescript version not supported. Thus, I updated eslint plugin too.

src/aggregation/Aggregator.ts
```diff
@@ -22,6 +22,7 @@ import * as AggregatorFactory from './AggregatorFactory';
 /**
  * Base interface for all aggregators.
  */
+// eslint-disable-next-line
```
srknzl, line 25: After I updated the eslint plugin, it gave linting errors about unused variables. I could not understand why we have a <R> generic type here which is not used in this interface. So I had to add a eslint disable line at the top. Feel free suggest another solution

src/proxy/MapListener.ts
```diff
@@ -26,7 +26,7 @@ export interface MapListener<K, V> extends EntryListener<K, V> {
 /**
  * Map event hander function.
  */
-export type MapEventListener<K, V> = (mapEvent: MapEvent) => void;
+export type MapEventListener = (mapEvent: MapEvent) => void;
```
srknzl, line 29: This has <K,V>, which is also not used, and was causing linting errors.

package.json
mdumandag: Can you also verify that the updated dependencies support our minimum supported node version (v10). IIRC, there were some problems before with eslint so we had to use an older version

src/aggregation/Aggregator.ts
mdumandag: It is probably ported from the Java side, in which we use the `R`, but it is required for the server-side. https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/aggregation/Aggregator.java#L84

We probably don't need this in the client side. I am not sure whether or not removing the generic type parameter can be counted as backward-compatible change, so lets leave it like that

mdumandag:
verify

src/aggregation/Aggregator.ts
srknzl: Is this backward compatibility related to client or server? Plus, by backward compatible which versions do you mean? last major version, that is 4.x, maybe? So any 4.x client should be compatible with features in older versions of 4.x?

mdumandag: I mean the client. 

What we cannot be sure of is that some users might be using this public interface (or the MapEventListener) (I highly doubt there are any), with the type parameters. If so, their code won't compile when they upgrade the client version

package.json
srknzl: In https://github.com/typescript-eslint/typescript-eslint/blob/v4.17.0/package.json we have 

```
"engines": {
    "node": "^10.12.0 || >=12.0.0"
  }
```

I guess this means node should be at least 10.12 which is a problem for us? Let me investigate other versions of this

mdumandag: Then this should be fine. We only use eslint as a devDependency and our PR builder uses Node.js v10.21, so that combination should work just fine

src/aggregation/Aggregator.ts
srknzl: Reverted interface changes.

mdumandag (approved):

srknzl:
Thanks @mdumandag for review

PR #818, opened 2021-03-15
Refactor jsdocs

srknzl:
Aims to fix grammatical errors and improve understandability of jsdoc comments

### DIFF
--- a/src/core/Member.ts
+++ b/src/core/Member.ts
@@ -27,7 +27,7 @@ export interface Member {
     address: Address;
 
     /**
-     * Unique id of a member in a cluster.
+     * Unique id of the member in a cluster.
      */
     uuid: UUID;
 
--- a/src/core/Member.ts
+++ b/src/core/Member.ts
@@ -27,7 +27,7 @@ export interface Member {
     address: Address;
 
     /**
-     * Unique id of member in cluster.
+     * Unique id of a member in a cluster.
      */
     uuid: UUID;
 
--- a/src/proxy/IMap.ts
+++ b/src/proxy/IMap.ts
@@ -341,8 +341,12 @@ export interface IMap<K, V> extends DistributedObject {
      * Locks the given key for this map. Promise is resolved when lock is
      * successfully acquired. This means it may never be resolved if some
      * other process holds the lock and does not unlock it. A lock may be
-     * acquired on non-existent keys. Other processes wait on the locked non-existent key.
-     * When this client puts the non-existent key, it is allowed to do that.
+     * acquired on non-existent keys. In that case, other clients would
+     * block until the non-existent key is unlocked. If the lock holder
+     * introduces the key to the map, the put operation is not blocked.
+     * If a client not holding a lock on the non-existent key tries to
+     * introduce the key while a lock exists on the non-existent key,
+     * the put operation blocks until it is unlocked.
      *
      * Locking is reentrant, meaning that the lock owner client can obtain the
      * lock multiple times. If the lock was acquired multiple times, then `unlock()`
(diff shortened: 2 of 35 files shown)


src/proxy/IMap.ts
```diff
@@ -341,7 +341,7 @@ export interface IMap<K, V> extends DistributedObject {
      * Locks the given key for this map. Promise is resolved when lock is
      * successfully acquired. This means it may never be resolved if some
      * other process holds the lock and does not unlock it. A lock may be
-     * acquired on non-existent keys. Other processes wait on non-existent key.
+     * acquired on non-existent keys. Other processes wait on the locked non-existent key.
```
srknzl, line 344: Do not have strong opinion here. I did not understand this non-existent key thing. Could anyone explain?

src/proxy/PNCounter.ts
```diff
@@ -29,21 +29,21 @@ import {DistributedObject} from '../core/DistributedObject';
...
  * In this case, the caller will be notified with a `TargetDisconnectedError`
  *
  * The read and write methods provide monotonic read and RYW (read-your-write)
  * guarantees. These guarantees are session guarantees which means that if
  * no replica with the previously observed state is reachable, the session
- * guarantees are lost and the method invocation will throw a
+ * guarantees are lost, and the method invocation will throw a
  * `ConsistencyLostError`. This does not mean
- * that an update is lost. All of the updates are part of some replica and
+ * that an update is lost. All the updates are part of some replica and
```
srknzl, line 40: Reference for use of "all of" and "all": https://jakubmarian.com/is-it-all-the-or-all-of-the-in-english/

src/core/Member.ts
```diff
@@ -27,7 +27,7 @@ export interface Member {
     address: Address;
 
     /**
-     * Unique id of member in cluster.
+     * Unique id of a member in a cluster.
```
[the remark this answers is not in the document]
srknzl: I guess this ok --> Unique id of the member in a cluster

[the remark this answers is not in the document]
srknzl: or the for cluster as well?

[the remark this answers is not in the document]
srknzl: In general I think cluster can be many, client can be connected to any cluster, so used a for them; however the `client` is same and only, so `the` is for the client.

src/proxy/IMap.ts
srknzl: What about writing the same thing here as well? Certainly this comment is less understandable due to its client centric perspective

mdumandag: It makes sense, the writer of this jsdoc probably didn't copy directly since the Java side documentation is mentioning threads. What do you think about replacing this part with something like
```
A lock may be acquired on non-existent keys. In that case, other clients would block until the non-existent key is unlocked. If the lock holder introduces the key to the map, the put operation is not blocked. If a client not holding a lock on the non-existent key tries to introduce the key while a lock exists on the non-existent key, the put operation blocks until it is unlocked.
```

srknzl: Seems ok to me a question:
>  If the lock holder introduces the key to the map, the put operation is not blocked
Did not get if this "block" is for lock holder, or other clients?

mdumandag: The lock holder will not block but others will

srknzl:
should we add the change for 'DISCONNECTED'  lifecycle event for  as well to this pr? However, the java side did not mention cluster instead of member either. @mdumandag

mdumandag:
I think we can, you can also send a fix to the Java side

mdumandag (approved):

PR #819, opened 2021-03-15
HazelcastClient implements ClientForConnectionManager

srknzl:
ConnectionManager get the client instance with an interface called `ClientForConnectionManager`. This pr makes the `HazelcastClient` implement the interface. 

This is not a must change, but it helps to suppress some IDE's unused method behaviour. (e.g in jetbrains IDE)

### DIFF
--- a/src/network/ConnectionManager.ts
+++ b/src/network/ConnectionManager.ts
@@ -272,7 +272,7 @@ export class ConnectionRegistryImpl implements ConnectionRegistry {
     }
 }
 
-interface ClientForConnectionManager {
+export interface ClientForConnectionManager {
     onClusterChange(): void;
 
     sendStateToCluster(): Promise<void>;
--- a/src/network/ConnectionManager.ts
+++ b/src/network/ConnectionManager.ts
@@ -272,6 +272,7 @@ export class ConnectionRegistryImpl implements ConnectionRegistry {
     }
 }
 
+/** @internal */
 export interface ClientForConnectionManager {
     onClusterChange(): void;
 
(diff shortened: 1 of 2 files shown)


src/network/ConnectionManager.ts
```diff
@@ -272,7 +272,7 @@ export class ConnectionRegistryImpl implements ConnectionRegistry {
     }
 }
 
-interface ClientForConnectionManager {
+export interface ClientForConnectionManager {
```
puzpuzpuz, line 275: This one should be `@internal`

mdumandag (approved):

mdumandag:
verify

puzpuzpuz (approved):

srknzl:
Hazelcast client is not internal but it implements an internal interface that is not possible I guess

puzpuzpuz:
> Hazelcast client is not internal but it implements an internal interface that is not possible I guess

As long as the interface is not included into generated typings and API docs, it's fine. Interfaces don't exist in the runtime (JS code).

srknzl:
> As long as the interface is not included into generated typings and API docs, it's fine. Interfaces don't exist in the runtime (JS code).

Even if `ClientForConnectionManager` is marked as internal it is generated in d.ts files. Specifically, `npm run validate-user-code` after `npm run compile ` fails due to import of non-existing `ClientForConnectionManager`.

srknzl:
Closing as I don't think there is a way to strip internal interfaces from a public api's "implements" part. 

from this line specifically(here, the interface is marked as internal and I tried to move the interface into HazelcastClient.ts as well):
```ts
export declare class HazelcastClient implements ClientForConnectionManager
```

PR #822, opened 2021-03-18
Change method of determining which files to publish  

### LINKED ISSUE #815: Specifying published files easier
We are using .npmignore file to ignore files from published. Some files always included as said in this link:
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files

What we can do instead is, add "files": "lib/" to package.json and delete .npmignore file. Keeping track of what to ignore is harder.

E.g You can see currently tsconfig.eslint.json is included in package when installed via `npm i hazelcast-client`

These are what is in the published files right now:

```
 lib
.npmignore
package-lock.json(this should be excluded normally unless npm-shrinkwrap.json is used)
tsconfig.eslint.json
LICENSE
package.json
README.md

```

srknzl:
Deletes .npmignore and adds "lib/" to package.json's "files" field. This makes it easier to track what we publish. This pr also removes some unnecessary files in the published files like `tslint.eslint.config`. Running `npx npm-packlist` will give the list of files that will be published.

for more detail you can go to the issue

fixes #815


package.json
```diff
@@ -4,6 +4,9 @@
     "description": "Hazelcast - open source In-Memory Data Grid - client for Node.js",
     "main": "lib/index.js",
     "types": "lib/index.d.ts",
+    "files": [
+        "lib/"
```
mdumandag, line 8: what about README and LICENSE?

srknzl: They are included by default, check this link in the linked issue https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files

mdumandag (approved):

PR #823, opened 2021-03-18
Infinite reconnection attempts by default behaviour

### LINKED ISSUE #756: [TRACKING ISSUE] Set cluster connect timeout to infinite by default

**Set cluster connect timeout to infinite by default**

See https://github.com/hazelcast/hazelcast/pull/18094 for details.

---

After an internal discussion, we have decided to set the cluster
connect timeout to infinite by default. Also, the multiplier
for the sleep time between consecutive connect attempts is changed
to `1.05` from `1.0`. The new sleep times between connection
attempts can be seen in the following graph. https://www.desmos.com/calculator/hl7rnbk4gf
(x-axis: attempt number, y-axis: sleep time in seconds)

Also, updated some logging statements.

Note that, this is a breaking change for the users who rely on the client to stop
trying to connect cluster after some time, but we decided

srknzl:
Infinite reconnection attempts by changing default `clusterConnectTimeoutMillis`. Also makes default multiplier `1.05` 

close #756

### DIFF
--- a/src/config/ConfigBuilder.ts
+++ b/src/config/ConfigBuilder.ts
@@ -120,8 +120,8 @@ export class ConfigBuilder {
                 this.effectiveConfig.connectionStrategy.connectionRetry.multiplier = tryGetNumber(value);
             } else if (key === 'clusterConnectTimeoutMillis') {
                 const clusterConnectTimeoutMillis = tryGetNumber(value);
-                if (clusterConnectTimeoutMillis < 0) {
-                    throw new RangeError('clusterConnectTimeoutMillis can only be non-negative');
+                if (clusterConnectTimeoutMillis != -1 && clusterConnectTimeoutMillis < 0) {
+                    throw new RangeError('clusterConnectTimeoutMillis can be only positive or -1');
                 }
                 this.effectiveConfig.connectionStrategy.connectionRetry.clusterConnectTimeoutMillis = clusterConnectTimeoutMillis;
             } else if (key === 'jitter') {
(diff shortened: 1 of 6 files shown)


src/network/WaitStrategy.ts
```diff
@@ -40,7 +40,8 @@ export class WaitStrategy {
         this.initialBackoffMillis = initialBackoffMillis;
         this.maxBackoffMillis = maxBackoffMillis;
         this.multiplier = multiplier;
-        this.clusterConnectTimeoutMillis = clusterConnectTimeoutMillis;
+        this.clusterConnectTimeoutMillis = clusterConnectTimeoutMillis === -1 ?
+            Number.POSITIVE_INFINITY : clusterConnectTimeoutMillis;
```
mdumandag, line 44: Looking at the other places we use infinite values, I suggest you use `Numer.MAX_SAFE_INTEGER` here. IMHO, it is also better for logging purposes. (`cluster connect timeout: Infinity ms` vs `cluster connect timeout: 9007199254740991 ms`)

src/config/ConfigBuilder.ts
```diff
@@ -119,8 +119,11 @@ export class ConfigBuilder {
             } else if (key === 'multiplier') {
                 this.effectiveConfig.connectionStrategy.connectionRetry.multiplier = tryGetNumber(value);
             } else if (key === 'clusterConnectTimeoutMillis') {
-                this.effectiveConfig.connectionStrategy.connectionRetry
-                    .clusterConnectTimeoutMillis = tryGetNumber(value);
+                const clusterConnectTimeoutMillis = tryGetNumber(value);
+                if (clusterConnectTimeoutMillis < 0) {
+                    throw new RangeError('clusterConnectTimeoutMillis can only be non-negative');
```
mdumandag, line 124: No, -1 is a better default value. I meant the error message `clusterConnectTimeoutMillis can be only positive or -1` is wrong, as it can be zero. It should have been `clusterConnectTimeoutMillis can be only non-negative or -1`

mdumandag: 0 has a special meaning that, do not try to connect to cluster if the attempt fails first time

srknzl: Got your point now

mdumandag (approved):

srknzl:
Thanks @mdumandag

PR #825, opened 2021-03-19
Update dependencies

mdumandag:
There were some dependencies updated since the last time. Also,
`npm audit` was showing some vulnerabilities for some of them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,26 +9,26 @@
         "long": "4.0.0"
     },
     "devDependencies": {
-        "@types/node": "^10.17.28",
-        "@typescript-eslint/eslint-plugin": "^4.17.0",
-        "@typescript-eslint/parser": "^4.17.0",
-        "chai": "^4.2.0",
+        "@types/node": "^10.17.55",
+        "@typescript-eslint/eslint-plugin": "^4.18.0",
+        "@typescript-eslint/parser": "^4.18.0",
+        "chai": "^4.3.4",
         "chai-as-promised": "^7.1.1",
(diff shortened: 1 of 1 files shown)


srknzl:
This looks good. I am trying to understand why `npm update` does not update versions in package.json (npm v7). And I saw this bug https://github.com/npm/cli/issues/2704. Obviously the documentation of npm update is wrong and the new behaviour is worse (changed in v6 as stated in the issue). 

One more point, do you know why we don't include package-lock.json into git. I saw https://github.com/hazelcast/hazelcast-nodejs-client/pull/526, but come on, is the only purpose of it to be published? Can you check https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json

srknzl:
Plus, according to https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files, package-lock.json file is always excluded from publishing unless shrinkwrapped. Contrary to what you said in #526. (I mean, purpose of package-lock.json being to be published)

mdumandag:
See the discussion at https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

I don't think package lock is much helpful for library packages.

srknzl:
I mean it is useful for getting same versions among developers during development. It does not have to affect end users.

srknzl:
I wish we had an option to add a lock file for just devDependencies, but I think we can't do this. Forget about the lock file then

srknzl (approved):

PR #826, opened 2021-03-19
Properly log tried addresses when the client cannot connect to cluster

mdumandag:
Sets do not properly display their elements when they are being tried
to represented as text values. We were trying to convert the tried
addresses set to a string while logging that the client is unable
to connect a cluster which was not working very well.

Now, we convert the set to an array and join it with `, ` to produce
a better output.

Before:
```
... The following addresses were tried: [object Set]
```

After:
```
... The following addresses were tried: 127.0.0.1:5701, 127.0.0.1:5703, 127.0.0.1:5702
```


srknzl:
Thanks this was in my "TODO"  notes

srknzl (approved):

PR #827, opened 2021-03-22
Make dev dependency versions similar

srknzl:
This uses tilde (~) instead of caret (^) in `package.json` for dev dependencies. This way, the development dependency versions developers and CI use will be as similar as possible, leading to less "working on my machine" problem.


mdumandag (approved):
OK for the changes, but I have a question. Since we are being more restrictive about the dependencies, what do you think about setting up something like dependabot for the repository to update them more frequently?

mdumandag:
We do have it for the hazelcast repository(hence the Java client) but not for others

Issue #788, opened 2021-01-26
[TRACKING ISSUE] SQL: New row serialization protocol


ghost:
The tracking issue for the Java side PR.

**SQL: New row serialization protocol**

See https://github.com/hazelcast/hazelcast/pull/18089 for details.

---

This PR introduces the new serialization mechanics for SQL row data. 

Before the PR, we serialized every value as `Data`. It had two problems:
1. It was inefficient.
2. It serialized some types (e.g. decimal) in a way, that the non-Java client cannot read.

With the new PR, we introduce the new custom codec `SqlPageCodec` that serializes values using a custom built-in list serializer depending on the type of the column. The main points:
1. We define the `SqlPage` class that may store values either in the row format or in the columnar format. We use the row format on the member size, and then convert it to the columnar format during decoding. This allows us to avoid copying between row format and columnar format on the member-side, making encoding almost garbage-free.
2. For every column, we pick the proper encoder based on the column type. Since every column in SQL could be nullable, we introduce several new list encoders for fixed-size types that encode nulls in the form of bitmap. See `ListCNFixedSizeCodec` and classes that use it.
3. We support custom encoders only for `VARCHAR` , `BOOLEAN`, `TINYINT`, `INTEGER`, and `BIGINT` types. Other types are still serialized using the `Data` object. We will add more custom encoders gradually. In the end, only the `OBJECT` type will be serialized as `Data`. The idea is to merge the infrastructure ASAP, and then focus on specific types (which could be time-consuming, because we need to align the encoding format with the new serialization format). This should be done before the 4.2 release.
4. The test `SqlPageCodecTest` tests all combinations of null and non-null values for every type.

Protocol part: https://github.com/hazelcast/hazelcast-client-protocol/pull/370
Supercedes https://github.com/hazelcast/hazelcast/pull/18018

PR #847, opened 2021-04-05
[API-270] SQL support

srknzl:
Adds sql support

merge https://github.com/hazelcast/hazelcast-nodejs-client/pull/859 first

fixes https://github.com/hazelcast/hazelcast-nodejs-client/issues/788


srknzl (commented):
ignore this 😄

srknzl:
verify

srknzl:
verify

srknzl:
verify

test/integration/backward_compatible/sql/DecodeAndSerializationTest.js
```diff
@@ -0,0 +1,733 @@
...
+        await basicSetup();
+        const script = `
+                    var map = instance_0.getMap("${mapName}");
+                    for (var key = 1; key < 12; key++) {
+                        map.set(new java.lang.Integer(key), java.time.LocalTime.of(key+3,key+2,key+1,key + 1e8));
+                    }
+                `;
+        await RC.executeOnController(cluster.id, script, Lang.JAVASCRIPT);
+        const result = client.getSqlService().execute(
+            `SELECT * FROM ${mapName} WHERE this > CAST (? AS TIME) AND this < CAST (? AS TIME) ORDER BY __key ASC`,
```
sancar, line 442: Can you remind me why do we need these casts?

srknzl: Because for date time related classes, strings are sent behind the scenes, which are needed to be casted.

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+                throw new RangeError('Timeout millis cannot be less than -1');
+            }
+        }
+
+        if (sqlStatementOptions.hasOwnProperty('cursorBufferSize') && tryGetNumber(sqlStatementOptions.cursorBufferSize) <= 0) {
+            throw new RangeError('Cursor buffer size cannot be negative');
+        }
+
+        if (sqlStatementOptions.hasOwnProperty('expectedResultType'))
+            tryGetEnum(SqlExpectedResultType, sqlStatementOptions.expectedResultType);
```
mdumandag, line 153: I think it would be good to use `{}` even if the if statement has a body containing single line to make it consistent across the code base

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+        if (sqlStatementOptions.hasOwnProperty('returnRawResult'))
+            tryGetBoolean(sqlStatementOptions.returnRawResult);
+    }
+
+    /**
+     * Converts datetime related wrapper classes to string to be able to serialize them. (No default serializers yet)
+     * @internal
+     * @param value
+     */
+    convertToStringIfDatetimeValue(value: any) {
```
mdumandag, line 164: I couldn't remember exactly, but didn't we say that we don't perform any implicit conversions for date values? We suppose to write a documentation about them since those implicit casts might not work perfectly

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+    execute(sql: SqlStatement): SqlResult;
+    execute(sql: string, params?: any[], options?: SqlStatementOptions): SqlResult;
+    execute(sql: string | SqlStatement, params?: any[], options?: SqlStatementOptions): SqlResult {
+        let sqlStatement: SqlStatement;
+
+        if (typeof sql === 'string') {
+            sqlStatement = {
+                sql: sql
+            };
+            if (params !== undefined) { // todo: document this
```
mdumandag, line 183: Will you deal with this todo?

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+                this.connectionManager.getClientUuid(),
+                SqlErrorCode.CONNECTION_PROBLEM,
+                'Client is not currently connected to the cluster.'
+            );
+        }
+
+        const queryId = SqlQueryId.fromMemberId(connection.getRemoteUuid());
+
+        const expectedResultType: SqlExpectedResultType =
+            sqlStatement.options?.expectedResultType ? SqlExpectedResultType[sqlStatement.options.expectedResultType]
```
mdumandag, line 213: I am not really familiar with typescript but what will happen if the `expectedResultType` has a 0 value? Which branch we will take?

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+
+        try {
+            const serializedParams = [];
+            if (Array.isArray(sqlStatement.params)) { // params can be undefined
+                for (let param of sqlStatement.params) {
+                    param = this.convertToStringIfDatetimeValue(param);
+                    serializedParams.push(this.serializationService.toData(param));
+                }
+            }
+            const cursorBufferSize = sqlStatement.options?.cursorBufferSize ?
```
mdumandag, line 224: same comment about the 0 value of `cursorBufferSize`

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+                sqlStatement.options.cursorBufferSize : SqlServiceImpl.DEFAULT_CURSOR_BUFFER_SIZE;
+            const requestMessage = SqlExecuteCodec.encodeRequest(
+                sqlStatement.sql,
+                serializedParams,
+                sqlStatement.options?.timeoutMillis ? sqlStatement.options.timeoutMillis : SqlServiceImpl.DEFAULT_TIMEOUT,
```
mdumandag, line 229: same comment about the 0 value of `timeoutMillis`

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+                cursorBufferSize,
+                sqlStatement.options?.schema ? sqlStatement.options.schema : SqlServiceImpl.DEFAULT_SCHEMA,
```
mdumandag, line 231: what if the `schema` is an empty string (or any falsy string value)

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+     *
+     * @param sqlStatement
+     * @throws RangeError if validation is not successful
+     * @internal
+     */
+    static validateSqlStatement(sqlStatement: SqlStatement | null): void {
+        if (sqlStatement === null) {
+            throw new RangeError('Sql statement cannot be null');
+        }
+        tryGetString(sqlStatement.sql);
```
mdumandag, line 122: just being string is not enough, it should also be non-empty

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+     * Validates SqlStatementOptions
+     *
+     * @param sqlStatementOptions
+     * @throws RangeError if validation is not successful
+     * @internal
+     */
+
+    static validateSqlStatementOptions(sqlStatementOptions: SqlStatementOptions): void {
+        if (sqlStatementOptions.hasOwnProperty('schema'))
+            tryGetString(sqlStatementOptions.schema);
```
mdumandag, line 139: schema can be null on the java side. does this check allow this?

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+                cursorBufferSize,
+                sqlStatement.options?.hasOwnProperty('returnRawResult') ?
+                    sqlStatement.options.returnRawResult : SqlServiceImpl.DEFAULT_FOR_RETURN_RAW_RESULT,
+                this.connectionManager.getClientUuid()
+            );
+
+            this.invocationService.invokeOnConnection(connection, requestMessage).then(clientMessage => {
+                this.handleExecuteResponse(clientMessage, res);
+            }).catch(err => {
+                console.log(err);
```
mdumandag, line 250: probably leftover

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+                res.onExecuteError(
+                    new HazelcastSqlException(
+                        connection.getRemoteUuid(), SqlErrorCode.CONNECTION_PROBLEM, err.message, err
```
mdumandag, line 253: Would it be to check whether or not the connection is alive here? If it is not alive, we can say that cluster topology is changed (like the Java side), and if it is, we can throw a generic error

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+                SqlErrorCode.GENERIC,
+                `An error occurred during SQL execution: ${error.message}`,
+                error
+            );
+        }
+    }
+
+    close(connection: Connection, queryId: SqlQueryId): Promise<void> {
+        const requestMessage = SqlCloseCodec.encodeRequest(queryId);
+        return this.invocationService.invokeOnConnection(connection, requestMessage).then(() => {
```
mdumandag, line 271: Since this is just used by SqlResult internally, we can make this return directly the `return this.invocationService.invokeOnConnection(connection, requestMessage)`, instead of adding a no-op continuation

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+        return this.invocationService.invokeOnConnection(connection, requestMessage).then(clientMessage => {
+            const response: SqlFetchResponseParams = SqlFetchCodec.decodeResponse(clientMessage);
+            if (response.error !== null) {
+                throw new HazelcastSqlException(
+                    response.error.originatingMemberId,
+                    response.error.code,
+                    response.error.message
+                );
+            }
+            assertNotNull(response.rowPage);
```
mdumandag, line 286: I think we can remove this assert

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import * as Long from 'long';
+
+export enum SqlExpectedResultType {
+    ANY,
+    ROWS,
+    UPDATE_COUNT
+}
+
+export type SqlExpectedResultTypeStrings = keyof typeof SqlExpectedResultType;
```
mdumandag, line 25: Why do we need this type?

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,90 @@
...
+export interface SqlStatementOptions {
+    /**
+     * The schema name. The engine will try to resolve the non-qualified object identifiers from the statement in the
+     * given schema. If not found, the default search path will be used, which looks for objects in the predefined
+     * schemas `partitioned` and `public`.
+     *
+     * * The schema name is case sensitive. For example, `foo` and `Foo` are different schemas.
+     * * The default value is `null` meaning only the default search path is used.
+     */
+    schema?: string;
```
mdumandag, line 39: should this have the type of `string | null`?

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * The execution timeout in milliseconds. If the timeout is reached for a running statement, it will be cancelled forcefully.
+     * Zero value means no timeout. -1 means that the timeout in server config will be used. Other negative values are prohibited.
```
mdumandag, line 42: Can we put all numbers into backticks?

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,90 @@
...
+     * Defaults to `-1`.
+     */
+    timeoutMillis?: Long;
```
mdumandag, line 45: Should we allow numbers here too, to make it easier to use?

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,90 @@
...
+     * boost for queries with large result sets at the cost of increased memory consumption.
+     *
+     * Defaults to `4096`.
+     */
+    cursorBufferSize?: number;
+    /**
+     * Expected result type of SQL query. By default, set to `ANY`. Available values
+     * are `ANY`, `ROWS`, and `UPDATE_COUNT`.
+     */
+    expectedResultType?: SqlExpectedResultTypeStrings;
```
mdumandag, line 66: What happens when I try to set it to something like `SqlExpectedResultType.ANY`?

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as Long from 'long';
+
+export enum SqlExpectedResultType {
```
mdumandag, line 19: Missing documentation

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Expected result type of SQL query. By default, set to `ANY`. Available values
+     * are `ANY`, `ROWS`, and `UPDATE_COUNT`.
+     */
+    expectedResultType?: SqlExpectedResultTypeStrings;
+    /**
+     * If true, SQL result will iterate over {@link SqlRow}s. If false SQL result, will iterate over regular objects.
+     * Defaults to `false`.
+     */
+    returnRawResult?: boolean;
```
mdumandag, line 71: Why do we have this extra parameter(`returnRawResult`). Can you write two equivalent codes one that sets it to false and one that sets it to true to show the difference? I am trying to justify the reasoning behind this

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/**
+ * Represents an SQL statement. This can be used to prepare SQL statement before {@link SqlService.execute}.
```
mdumandag, line 75: There is a bit more documentation in the Java side

src/sql/index.ts
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/** @ignore *//** */
+
+/**
+ * Public API re-exports.
+ */
+
+export * from './DatetimeClasses';
```
mdumandag, line 22: Do the other public exports automatically reexported?

src/util/RandomLB.ts
```diff
@@ -24,12 +24,24 @@ import {randomInt} from '../util/Util';
  * @internal
  */
 export class RandomLB extends AbstractLoadBalancer {
-    next(): Member {
-        const members = this.getMembers();
+    next(): Member | null {
+        return this._next(false);
+    }
+
+    _next(dataMember: boolean): Member | null {
```
mdumandag, line 31: should we also make this method private?

src/util/RoundRobinLB.ts
```diff
@@ -34,8 +34,12 @@ export class RoundRobinLB extends AbstractLoadBalancer {
         this.index = randomInt(INITIAL_SEED_CAP);
     }
 
-    next(): Member {
-        const members = this.getMembers();
+    next(): Member | null {
+        return this._next(false);
+    }
+
+    _next(dataMember: boolean): Member | null {
```
mdumandag, line 41: should we also make this method private?

mdumandag (commented):
Reviewed the implementation only (apart from the date related parts). I will review the tests after resolving all my comments to make it easier

src/codec/builtin/BigDecimalCodec.ts
```diff
@@ -0,0 +1,46 @@
...
+        return CodecUtil.nextFrameIsNullFrame(clientMessage) ? null : BigDecimalCodec.decode(clientMessage);
+    }
+
+    static toScale(bigIntString: string, scale: number): string {
+        if (scale === 0) {
+            return bigIntString;
+        } else if (scale > 0) {
+            // '1111'
+            if (scale < bigIntString.length) {
+                return bigIntString.substring(0, bigIntString.length - scale) + '.'
```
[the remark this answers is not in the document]
srknzl: for negative numbers I fixed the codec.

src/codec/builtin/ListCNFixedSizeCodec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    static decode<T>(
+        frame: Frame,
+        itemSizeInBytes: number,
+        decoder: (buffer: Buffer, position: number) => T
+    ): T[] {
+        const type = FixSizedTypesCodec.decodeByte(frame.content, 0);
+        const count = FixSizedTypesCodec.decodeInt(frame.content, 1);
+
+        const res = [];
```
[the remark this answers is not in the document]
srknzl: I think so

src/codec/builtin/BigDecimalCodec.ts
```diff
@@ -3,44 +3,52 @@ import {CodecUtil} from './CodecUtil';
...
 /** @internal */
 export class BigDecimalCodec {
 
     static decode(clientMessage: ClientMessage): string {
         const buffer = clientMessage.nextFrame().content;
         const contentSize = FixSizedTypesCodec.decodeInt(buffer, 0);
-        const body = new Uint8Array(buffer.slice(BitsUtil.INT_SIZE_IN_BYTES, BitsUtil.INT_SIZE_IN_BYTES + contentSize));
-        const scale = FixSizedTypesCodec.decodeInt(buffer, BitsUtil.INT_SIZE_IN_BYTES + contentSize);
+        let body = buffer.slice(BitsUtil.INT_SIZE_IN_BYTES, BitsUtil.INT_SIZE_IN_BYTES + contentSize);
+        const byteArray = new Uint8Array(body);
```
mdumandag, line 13: Why are we still wrapping it into Uint8Array?

src/codec/builtin/BigDecimalCodec.ts
```diff
@@ -3,44 +3,52 @@ import {CodecUtil} from './CodecUtil';
...
+        const signBit = body.slice(0, 1).readUInt8(0);
```
mdumandag, line 14: Can we do it without slicing the body?

src/codec/builtin/BigDecimalCodec.ts
```diff
@@ -3,44 +3,52 @@ import {CodecUtil} from './CodecUtil';
...
+        const isNegative = signBit > 127;
+        if (isNegative) { // negative, convert two's complement to positive
+            const newByteArray = [];
```
mdumandag, line 17: can we allocate the buffer beforehand since we know its size?

src/codec/builtin/BigDecimalCodec.ts
```diff
@@ -3,44 +3,52 @@ import {CodecUtil} from './CodecUtil';
...
         } else if (scale > 0) {
-            // '1111'
             if (scale < bigIntString.length) {
-                return bigIntString.substring(0, bigIntString.length - scale) + '.'
+                return (isNegative ? '-' : '') + bigIntString.substring(0, bigIntString.length - scale) + '.'
                     + bigIntString.substring(bigIntString.length - scale);
             } else {
                 const numberOfZerosAfterDecimal = scale - bigIntString.length;
-                return '0.' + '0'.repeat(numberOfZerosAfterDecimal) + bigIntString
+                return (isNegative ? '-' : '') + '0.' + '0'.repeat(numberOfZerosAfterDecimal) + bigIntString
```
mdumandag, line 48: can we make it like `(isNegative ? '-0.' : '0.') + ...` to make it concatenate less?

src/codec/builtin/ListCNFixedSizeCodec.ts
```diff
@@ -18,37 +17,35 @@ export class ListCNFixedSizeCodec {
     ): T[] {
         const type = FixSizedTypesCodec.decodeByte(frame.content, 0);
         const count = FixSizedTypesCodec.decodeInt(frame.content, 1);
+        let position = ListCNFixedSizeCodec.HEADER_SIZE;
```
mdumandag, line 20: Can we define those a little bit below on the `default` branch, since they are used there?

src/codec/builtin/ListCNFixedSizeCodec.ts
```diff
@@ -18,37 +17,35 @@ export class ListCNFixedSizeCodec {
...
-                assert.strictEqual(type, ListCNFixedSizeCodec.TYPE_MIXED, 'Invalid type');
-                let position = ListCNFixedSizeCodec.HEADER_SIZE;
-                let readCount = 0;
                 while (readCount < count) {
                     const bitmask = FixSizedTypesCodec.decodeByte(frame.content, position);
                     for (let i = 0; i < ListCNFixedSizeCodec.ITEMS_PER_BITMASK && readCount < count; i++) {
                         const mask = 1 << i;
                         if ((bitmask & mask) === mask) {
-                            res.push(decoder(frame.content, position));
+                            res[i] = decoder(frame.content, position);
```
mdumandag, line 41: I don't think this works, this will cycle between the first 8 elements of the result and overwrite elements. We should be using the readCount as the index

src/sql/SqlResult.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    /**
+     * Returns row metadata of the result.
+     * @returns SqlRowMetadata if rows exists in the result, otherwise, e.g update count result is received, to null.
+     * If SQL execution was not successful, this promise is rejected with an error.
+     */
+    getRowMetadata(): Promise<SqlRowMetadata | null>;
+
+    /**
+     * Return whether this result has rows to iterate. False if update count is returned or a response is not received yet.
```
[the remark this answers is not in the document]
srknzl: yes good catch

src/sql/SqlResult.ts
```diff
@@ -0,0 +1,360 @@
...
+            return this.closeDeferred.promise;
+        }
+
+        this.closeDeferred = deferredPromise<void>();
+
+        const error = new HazelcastSqlException(this.clientUUID, SqlErrorCode.CANCELLED_BY_USER, 'Cancelled by user');
+        // Reject execute with user cancellation error.
+        this.onExecuteError(error);
+        // Reject ongoing fetch request if there is one.
+        if (this.fetchDeferred?.promise)
```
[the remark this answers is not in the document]
srknzl: It will not reject the promise. fetch() has a check whether or not the result is closed. If so it will be rejected.

src/codec/builtin/BigDecimalCodec.ts
srknzl: yes. I was trying to get first bit but only bytes can be sliced

src/codec/builtin/BigDecimalCodec.ts
srknzl: In order to be able to negate bytes we need a ArrayBuffer view. UInt8Array works for this case.

src/codec/builtin/BigDecimalCodec.ts
srknzl: yes

src/codec/builtin/ListCNFixedSizeCodec.ts
srknzl: that given an eslint warning that's why I moved them

src/codec/builtin/BigDecimalCodec.ts
mdumandag: I get what you are trying to do, but how this is different than what you are doing?
```js
const buf = Buffer.from([1,2,3,4,5]);
const negated = Buffer.allocUnsafe(buf.length);

for (let i = 0; i < buf.length; i++) {
    negated[i] = ~buf[i];
}
```

src/codec/builtin/ListCNFixedSizeCodec.ts
mdumandag: I think we can suppress the warning, it doesn't make sense to define something that won't be used

src/codec/builtin/BigDecimalCodec.ts
srknzl: thanks I did not know I can do that

srknzl: Now it is much simpler

scripts/download-rc.js
```diff
@@ -92,13 +92,13 @@ const downloadRC = () => {
             console.log('hazelcast-enterprise.jar already exists, not downloading from maven.');
         } else {
             console.log('Downloading: hazelcast enterprise jar '
-                + `com.hazelcast:hazelcast-enterprise:${HAZELCAST_ENTERPRISE_VERSION}`);
+                + `com.hazelcast:hazelcast-enterprise-all:${HAZELCAST_ENTERPRISE_VERSION}`);
```
asimarslan, line 95: Why do we need all jar?

srknzl: when enterprise jar is used `Error: Cannot execute SQL query because "hazelcast-sql" module is not in the classpath.
` is thrown

src/sql/SqlResult.ts
```diff
@@ -0,0 +1,360 @@
...
+                    name: this.rowMetadata.getColumns()[i].name,
+                    value: this.serializationService.toObject(this.currentPage.getValue(this.currentPosition, i))
+                });
+            }
+            return new SqlRowImpl(values, this.rowMetadata);
+        } else { // Return objects
+            const result: SqlRowAsObject = {};
+            for (let i = 0; i < this.currentPage.getColumnCount(); i++) {
+                const columnMetadata = this.rowMetadata.getColumnByIndex(i);
+                if (columnMetadata != null) {
```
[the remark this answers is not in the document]
srknzl: it can be undefined, updated usage and docstring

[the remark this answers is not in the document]
srknzl: I think there is no need to check for undefined here though, going to delete it.

src/core/LoadBalancer.ts
```diff
@@ -59,7 +59,7 @@ export interface LoadBalancer {
     /**
      * Returns whether this instance supports getting data members through a call to {@link nextDataMember()}.
      *
-     * @return Returns the next member or `null` if no member is available
+     * @return Returns if this load balancer can get a data member.
```
mdumandag, line 62: `@return Returns if this load balancer can get a data member.` -> `@return `True` if this load balancer can get a data member.`

src/sql/SqlResult.ts
```diff
@@ -66,30 +66,45 @@ export type SqlRowType = SqlRow | SqlRowAsObject;
  * }
  * ```
  *
+ * #### Usage for update count
+ * ```js
+ * const updated = result.getUpdateCount();
```
mdumandag, line 71: you forget to await it

src/sql/SqlResult.ts
```diff
@@ -0,0 +1,360 @@
...
+     * Fetches the next page. Called internally by iteration logic. Pages are fetched whenever the current page is fully
+     * iterated.
+     */
+    fetch(): Promise<SqlPage> {
+        // If there is an ongoing fetch, return that promise
+        if (this.fetchDeferred?.promise)
+            return this.fetchDeferred.promise;
+
+        // Do not start a fetch if close is called
+        if (this.closeDeferred?.promise) {
```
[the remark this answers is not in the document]
srknzl: Are you sure? Java side sets an error to the fetch result. 

```java
try {
    // Do nothing if the result is already closed.
    if (closed) {
        return;
    }

    // If the cancellation is initiated before the first response is received, then throw cancellation errors on
    // the dependent methods (update count, row metadata, iterator).
    if (state == null) {
        onExecuteError(QueryException.cancelledByUser());
    }

    // Make sure that all subsequent fetches will fail.
    if (fetch == null) {
        fetch = new SqlFetchResult();
    }

    onFetchFinished(null, QueryException.cancelledByUser());

    // Send the close request.
    service.close(connection, queryId);
} finally {
    // Set the closed flag to avoid multiple close requests.
    closed = true;
}
```

with the line 

```java
onFetchFinished(null, QueryException.cancelledByUser());
```

subsequent calls to fetch causes the same error to be thrown again.

src/sql/SqlResult.ts
```diff
@@ -66,30 +66,45 @@ export type SqlRowType = SqlRow | SqlRowAsObject;
...
      * Returns row metadata of the result.
-     * @returns SqlRowMetadata if rows exists in the result, otherwise, e.g update count result is received, to null.
+     * @returns SQL row metadata if rows exist in the result; otherwise null.
      * If SQL execution was not successful, this promise is rejected with an error.
      */
     getRowMetadata(): Promise<SqlRowMetadata | null>;
 
     /**
-     * Return whether this result has rows to iterate. False if update count is returned or a response is not received yet.
+     * Return whether this result has rows to iterate. False if update count is returned, true is rows are returned.
```
mdumandag, line 107: `true is rows` -> `true if rows`

src/sql/SqlRow.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    getObject<T>(columnNameOrIndex: string | number): T {
+        if (typeof columnNameOrIndex === 'number') {
+            if (columnNameOrIndex < 0 || columnNameOrIndex >= this.namedValues.length) {
+                throw new IndexOutOfBoundsError(`Index ${columnNameOrIndex} does not exists.`);
+            }
+            return this.namedValues[columnNameOrIndex].value;
+        } else if (typeof columnNameOrIndex === 'string') {
+            for (const value of this.namedValues) {
```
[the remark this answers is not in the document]
srknzl: I stored the name for finding it, you are right I will use the rowmetadata.

src/sql/SqlResult.ts
```diff
@@ -0,0 +1,360 @@
...
+        // Return the close promise if a close request is already started
+        if (this.closeDeferred?.promise) {
+            return this.closeDeferred.promise;
+        }
+
+        this.closeDeferred = deferredPromise<void>();
+
+        const error = new HazelcastSqlException(this.clientUUID, SqlErrorCode.CANCELLED_BY_USER, 'Cancelled by user');
+        // Reject execute with user cancellation error.
+        this.onExecuteError(error);
```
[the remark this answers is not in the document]
mdumandag: OK about resolving/rejecting the promise twice (which I don't like, but it is in the spec so I can live with it :smile: ) but there are other side effects of this call, like setting `this.updateCount = Long.fromInt(-1);`. Can this have bad consequences?

src/sql/SqlResult.ts
```diff
@@ -319,6 +335,9 @@ export class SqlResultImpl implements SqlResult {
     /** Returns if there are rows to be iterated. Used by {@link next}. */
     hasNext(): Promise<boolean> {
         return this.executeDeferred.promise.then(() => {
+            return this.isRowSet();
```
mdumandag, line 338: We shouldn't need this extra promise chain (`return this.isRowSet();) We could check it already in the `then` block below by looking at the updateCount

src/codec/builtin/BigDecimalCodec.ts
```diff
@@ -11,7 +11,7 @@ export class BigDecimalCodec {
         const contentSize = FixSizedTypesCodec.decodeInt(buffer, 0);
         let body = buffer.slice(BitsUtil.INT_SIZE_IN_BYTES, BitsUtil.INT_SIZE_IN_BYTES + contentSize);
         const byteArray = new Uint8Array(body);
-        const signBit = body.slice(0, 1).readUInt8(0);
+        const signBit = body.readUInt8(0);
```
mdumandag, line 14: this is not the sign bit. it is the byte that contains the sign bit. To fix the naming, maybe we can remove this and just do `const isNegative = body.readUInt8(0) > 127;` or `const isNegative = body[0] & 0x80 === 1`

src/sql/SqlResult.ts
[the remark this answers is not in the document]
mdumandag: Ok, you are right, I guess I mis-remembered it

src/sql/SqlResult.ts
mdumandag: Also, I think the name of the variable is not good

src/sql/SqlService.ts
```diff
@@ -0,0 +1,290 @@
...
+     * @throws {@link IllegalArgumentError} If arguments are not valid
+     * @throws {@link HazelcastSqlException} If there is an error running SQL
+     */
+    execute(sql: SqlStatement): SqlResult;
+}
+
+/** @internal */
+export class SqlServiceImpl implements SqlService {
+    /** Value for the timeout that is not set. */
+    static readonly TIMEOUT_NOT_SET = Long.fromInt(-1);
```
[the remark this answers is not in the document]
srknzl: I chose to use simple objects for sql statement. I think they can belong to SqlService.

src/sql/SqlService.ts
srknzl: > We suppose to write a documentation about them since those implicit casts might not work perfectly
What do you mean? We are sending them as strings internally. This method is used to convert the wrapper classes to string.

srknzl: Implicit cast issue was writing cast to sql string instead of user writing it

src/sql/SqlService.ts
srknzl: I was going to write a comment, I wrote now

src/sql/SqlService.ts
mdumandag: Why do we send them as strings? I remember we said we won't support sending date types as parameters in 4.2. I remember saying that, if users want to send it, they can convert it to string, add cast to their queries. We should just document this behavior, and fix it 5.0

src/sql/SqlService.ts
srknzl: It takes strings of the enum, not values like 0 and 1.

src/sql/SqlService.ts
srknzl: it cannot have 0, it is validated. same for expectedResultType, validated.

src/sql/SqlService.ts
srknzl: You are right about this one. 0 should be handled differently

srknzl: no actually this is a long so an object. 0 value is truthy

src/sql/SqlService.ts
srknzl: `(or any falsy string value)` which is empty string. Handled empty string thanks. FYI, `sqlStatement.options?.schema` returns undefined if options is null or undefined.

src/sql/SqlService.ts
srknzl: I think instead of user giving null, he or she should not give it at all. This way we can simplify validation.

src/sql/SqlStatement.ts
srknzl: This is similar to other configs, where an enum is expected. I think it is like this because this way it is easier to use without an enum import. Entering integers directly like 0 and 1 on the other hand, is less readable.

I don't quite remember the reason though.

src/sql/SqlStatement.ts
srknzl: ok it would be better I think

src/sql/SqlStatement.ts
srknzl: Validation error. See my other related comment

src/sql/SqlService.ts
```diff
@@ -42,6 +42,92 @@ import {HzLocalTime, HzLocalDate, HzLocalDateTime, HzOffsetDateTime} from './Dat
 
 /**
  * SQL Service of the client. You can use this service to execute SQL queries.
+ *
+ * The service is in beta state. Behavior and API might change in future releases. Binary compatibility is not
```
mdumandag, line 45: Binary compatibility is a concept in Java. I don't think there is an equivalent for Node.js. I think we should delete that sentence from here.

src/sql/SqlService.ts
```diff
@@ -42,6 +42,92 @@ import {HzLocalTime, HzLocalDate, HzLocalDateTime, HzOffsetDateTime} from './Dat
...
+ * guaranteed between minor or patch releases.
+ *
+ * If this cluster is a Hazelcast Jet cluster, a statement can be executed by either the default SQL backend or by Hazelcast
```
mdumandag, line 48: I think we don't support Jet clusters in non-Java clients (at least in the 4.x series), so mentioning this here seems unnecessary to me.

src/sql/SqlStatement.ts
srknzl: Are you talking about `This object is mutable. Properties are read once before the execution is started. Changes to properties do not affect the behavior of already running statements.`. Removed the first sentence it does not make sense for js version.

src/sql/SqlService.ts
```diff
@@ -42,6 +42,92 @@ import {HzLocalTime, HzLocalDate, HzLocalDateTime, HzOffsetDateTime} from './Dat
...
+ *
+ * The text below summarizes features supported by the default SQL engine. For a summary of Hazelcast Jet SQL features
+ * see `com.hazelcast.jet.sql` package javadoc in Hazelcast Jet (once released).
+ *
+ *
+ * ### Overview
+ * Hazelcast is able to execute distributed SQL queries over the following entities:
+ * * IMap
+ *
+ * When an SQL statement is submitted to a member, it is parsed and optimized by the `hazelcast-sql` module, that is based
```
mdumandag, line 60: I don't think we should mention Calcite etc. in the clients. It is an implementation detail. Also, I have modified the documentation in the Python client PR so that we mention these details at the end, instead of scaring the user at the beginning with Java terms like classpath, etc. Maybe we can do the same here

src/sql/index.ts
srknzl: no in code samples repo they exist, here it is forgetten

src/sql/SqlService.ts
```diff
@@ -42,6 +42,92 @@ import {HzLocalTime, HzLocalDate, HzLocalDateTime, HzOffsetDateTime} from './Dat
...
+ * Every IMap instance is exposed as a table with the same name in the `partitioned` schema. The `partitioned`
+ * schema is included into a default search path, therefore an IMap could be referenced in an SQL statement with or without the
+ * schema name.
+ *
+ * ###### Column Resolution
+ * Every table backed by an IMap has a set of columns that are resolved automatically. Column resolution uses IMap entries
+ * located on the member that initiates the query. The engine extracts columns from a key and a value and then merges them
+ * into a single column set. In case the key and the value have columns with the same name, the key takes precedence.
+ *
+ * Columns are extracted from objects as follows:
```
mdumandag, line 78: We can note that resolution happens on the server-side. 

I used something like this in the Python doc.

```
Columns are extracted from objects as follows (which happens on the
    server-side)
```

src/sql/SqlService.ts
```diff
@@ -42,6 +42,92 @@ import {HzLocalTime, HzLocalDate, HzLocalDateTime, HzOffsetDateTime} from './Dat
...
+ * key (value) object has fields, then the whole key (value) field is exposed as a normal field. Otherwise the field is hidden.
+ * Hidden fields can be accessed directly, but are not returned by `SELECT * FROM ...` queries.
+ *
+ * If the member that initiates a query doesn't have local entries for the given IMap, the query fails.
+ *
+ * Consider the following key/value model:
+ *
+ * ```ts
+ * class PersonKey {
+ *         personId: number;
```
mdumandag, line 94: I think we should make this portable and show writePortable method to show the column names and types. That is the only way of setting the schema from the client side. For the python client, I used such samples

```
    Consider the following key/value model: ::

        class PersonKey(Portable):
            def __init__(self, person_id=None, department_id=None):
                self.person_id = person_id
                self.department_id = department_id

            def write_portable(self, writer):
                writer.write_long("person_id", self.person_id)
                writer.write_long("department_id", self.department_id)

            ...

        class Person(Portable):
            def __init__(self, name=None):
                self.name = name

            def write_portable(self, writer):
                writer.write_string("name", self.name)

            ...
```

src/sql/SqlService.ts
```diff
@@ -42,6 +42,92 @@ import {HzLocalTime, HzLocalDate, HzLocalDateTime, HzOffsetDateTime} from './Dat
...
+ * * __key OBJECT (hidden)
+ * * this OBJECT (hidden)
+ *
+ * ##### Consistency
+ *
+ * Results returned from IMap query are weakly consistent:
+ * * If an entry was not updated during iteration, it is guaranteed to be returned exactly once.
+ * * If an entry was modified during iteration, it might be returned zero, one or several times.
+ *
+ *
```
mdumandag, line 115: Unnecessary newline

src/core/LoadBalancer.ts
srknzl: I think old version is still ok, but I will change. I used `if` in the same meaning with `whether`

src/sql/SqlResult.ts
srknzl: yeah good point

src/sql/SqlService.ts
```diff
@@ -42,6 +42,92 @@ import {HzLocalTime, HzLocalDate, HzLocalDateTime, HzOffsetDateTime} from './Dat
...
+ * #### Usage
+ *
+ * When a query is executed, an {@link SqlResult} is returned. The returned result is an async iterable. It can also be
+ * iterated using {@link SqlResult.next} method. The result must be closed at the end.
+ * The code snippet below demonstrates a typical usage pattern:
+ *
+ * ```
+ * const client = await Client.newHazelcastClient();
+ * let result = client.getSqlService().execute('SELECT __key, this FROM myMap WHERE this > ?', [1]);
```
mdumandag, line 125: In this code sample, we can use the class we used above to make it better.

See the Java code sample, for example
```
 *     HazelcastInstance instance = ...;
 *
 *     try (SqlResult result = instance.sql().execute("SELECT * FROM person")) {
 *         for (SqlRow row : result) {
 *             long personId = row.getObject("personId");
 *             String name = row.getObject("name");
 *             ...
 *         }
 *     }
```

src/codec/builtin/BigDecimalCodec.ts
srknzl: rather should be `(body[0] & 0x80) > 0`

src/sql/SqlService.ts
srknzl: I vote for removing it

src/sql/SqlService.ts
srknzl: btw is your example missing factory id and class id

srknzl: Should I delete them too?

src/network/ConnectionManager.ts
```diff
@@ -194,26 +198,43 @@ export class ConnectionRegistryImpl implements ConnectionRegistry {
     /**
      * Returns a random connection from active connections. If smart routing enabled, connection is returned using
      * load balancer. Otherwise, it is the first connection in connection registry.
+     * @param dataMember if only data members should be considered
```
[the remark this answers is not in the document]
mdumandag: This is marked as resolved but I still see the duplicate documentation. We write the same thing in `interface ConnectionRegistry` and `class ConnectionRegistryImpl`.

src/sql/SqlService.ts
mdumandag: I think they are unrelated and can be omitted. We should just mention that this is a portable class

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,103 @@
...
+     * * The default value is `null` meaning only the default search path is used.
+     */
+    schema?: string | null;
+    /**
+     * The execution timeout in milliseconds. If the timeout is reached for a running statement, it will be cancelled forcefully.
+     * Zero value means no timeout. `-1` means that the timeout in server config will be used.
+     * Other negative values are prohibited.
+     * Defaults to `-1`.
+     */
+    timeoutMillis?: Long | number;
```
mdumandag, line 52: BTW, thinking about this again and looking at the other config options we used throughout the client (like `maxBackoffMillis` or `clusterConnectTimeoutMillis` of `ConnectionRetryConfig`, which are `long`s in Java), I have a feeling that just allowing numbers here would be better, constructing a long is harder and the range of the numbers are huge enough. What do you think about it?

src/sql/SqlService.ts
```diff
@@ -0,0 +1,420 @@
...
+                }
+            }
+            const cursorBufferSize = sqlStatement.options?.cursorBufferSize ?
+                sqlStatement.options.cursorBufferSize : SqlServiceImpl.DEFAULT_CURSOR_BUFFER_SIZE;
+            const requestMessage = SqlExecuteCodec.encodeRequest(
+                sqlStatement.sql,
+                serializedParams,
+                timeoutMillis,
+                cursorBufferSize,
+                sqlStatement.options?.schema || sqlStatement.options?.schema === '' || sqlStatement.options?.schema === null ?
```
mdumandag, line 331: Since the `DEFAULT_SCHEMA` is `null`, can't we simplify this with `sqlStatement.options?.schema`  instead of this ternary if logic?

src/sql/SqlResult.ts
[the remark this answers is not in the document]
srknzl: I don't think so. Just checked all the methods. calling close will cause all of them to reject, or no-op which ok I think

src/network/ConnectionManager.ts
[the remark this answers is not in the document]
srknzl: Done, I think I forgot it.

src/sql/SqlService.ts
srknzl: `sqlStatement.options?.schema` is undefined if options is undefined, not null. So directly using `sqlStatement.options?.schema` does not work. We can remove `sqlStatement.options?.schema === null` check, but I wanted to be explicit. Going to remove it.

src/sql/SqlStatement.ts
srknzl: I agree that can be addressed later, a backlog issue would be good.

src/sql/SqlService.ts
srknzl: I don't think we need to do that. Users will send wrapper classes as parameters in 4.2, they will be converted to string internally and sent. After default serializers is added, we will send them as byte array encoded, not strings. This way the user API will be the same for both cases, and there will be no inconvenience.

srknzl: We will only make in easier by removing the need for explicit casting when the default serializers for date and time classes are added. I think sometimes doing things implicitly can help, and this is one of them

src/sql/SqlStatement.ts
srknzl: Created an issue https://github.com/hazelcast/hazelcast-nodejs-client/issues/912

src/sql/SqlResult.ts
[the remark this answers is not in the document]
mdumandag: One thing that comes to my mind. You get an update count for your query. You initiated close, then the result object will return -1 when you try to get update count. I am not sure if we are okay with this kind of behavior. What do you think?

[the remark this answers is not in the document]
srknzl: If user got an update count, the result should be already closed isn't it?  since:

```js
if (page.isLast()) {
            this.last = true;
            this.closed = true;
}
```

calling close again is no-op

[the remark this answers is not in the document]
srknzl: Ah sorry this is not called but this:
```js
if (rowMetadata !== null) { // Result that including rows
            this.rowMetadata = rowMetadata;
            this.onNextPage(rowPage);
            this.updateCount = Long.fromInt(-1);
        } else { // Result that including update count
            this.updateCount = updateCount;
            this.closed = true;
        }
```

else part is executed

src/sql/SqlService.ts
mdumandag: I think we can discuss this with the team. I haven't looked at the documentation PR yet, but If we decide to send strings for these types, we should document this behavior

src/sql/SqlResult.ts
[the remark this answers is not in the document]
srknzl: Hmm actually calling close() again is not no-op I think we can add that check to the close().

```js
if(this.closed){
    return
}
```

src/sql/SqlService.ts
mdumandag: Sounds okay

srknzl: Added null support anyway :)

src/sql/SqlService.ts
```diff
@@ -0,0 +1,408 @@
...
+                cursorBufferSize,
+                sqlStatement.options?.hasOwnProperty('returnRawResult') ?
+                    sqlStatement.options.returnRawResult : SqlServiceImpl.DEFAULT_FOR_RETURN_RAW_RESULT,
+                this.connectionManager.getClientUuid()
+            );
+
+            this.invocationService.invokeOnConnection(connection, requestMessage).then(clientMessage => {
+                this.handleExecuteResponse(clientMessage, res);
+            }).catch(err => {
+                if(!connection.isAlive()){
```
mdumandag, line 339: Can we extract this to a separate function, and use it in other places? I see that Java does such rethrow logic for here, for handleExecute/FetchResponse and for close

src/sql/SqlStatement.ts
mdumandag: BTW, let me refresh my TypeScript knowledge :smile: `schema?: string` means it can either be undefined or string right? I think we can be fine with that

srknzl: you are right. just forgot to change it, it should be schema?: string | null;

srknzl: no I have already changed it :D

mdumandag: Sorry for not understanding this earlier. I have rusty TypeScript knowledge. 

Maybe it is more work for you but I think we should aim for the best possible/good-looking API for the users. I think if you also feel that `schema?: string` is better/more idiomatic than `schema?: string | null`, I vote for reverting these changes but I leave the decision to you

srknzl: The behavior change is allowing null or not. You proposed allowing null. ? means this option is optional. If it is not provided we will not validate it and use default schema. Do you propose disallowing null?

src/sql/SqlService.ts
srknzl: I guess you are not proposing removing date/time support altogether. Users should be able to use them still. Instead of them giving strings and making it harder for them, wrapper classes are written. Wrapper classes can be uses instead of them giving strings. We can document this internal behaviour and when serializer is added they will be already familiar with the wrapper class, which is better than introducing the wrapper class later on.

src/sql/SqlResult.ts
```diff
@@ -0,0 +1,380 @@
...
+            this.fetchDeferred.reject(this.sqlService.toHazelcastSqlException(err, this.connection));
+        });
+
+        return this.fetchDeferred.promise;
+    }
+
+    /** Returns if there are rows to be iterated. Used by {@link next}. */
+    hasNext(): Promise<boolean> {
+        return this.executeDeferred.promise.then(() => {
+            if (this.updateCount !== Long.fromInt(-1)) {
```
mdumandag, line 339: Since `hasNext` is called a lot, I think we can either
- make the `Long.fromInt(-1)` a constant and check against it, instead of recreating it each time
- use `rowMetadata` with null check instead of `updateCount`

I think second option should be faster

src/sql/SqlResult.ts
```diff
@@ -0,0 +1,380 @@
...
+                return Promise.reject(new IllegalStateError('This result contains only update count'));
+            }
+            // While loop similar logic is written via recursion when written in async manner
+            const checkHasNext = (): Promise<boolean> => {
```
mdumandag, line 343: I don't know much about this topic in Node.js but, this closure does not capture anything from above. So, it can be moved into a separate function.

Will this improve performance? Because, from what I see, we are creating a closure each time we call `hasNext`. This cost can be avoided with the approach I mentioned

src/sql/SqlService.ts
```diff
@@ -0,0 +1,422 @@
...
+/**
+ * SQL Service of the client. You can use this service to execute SQL queries.
+ *
+ * The service is in beta state. Behavior and API might change in future releases.
+ *
+ * ### Overview
+ * Hazelcast is able to execute distributed SQL queries over the following entities:
+ * * IMap
+ *
+ * During optimization, a statement is converted into a directed acyclic graph (DAG) that is sent to cluster members
```
mdumandag, line 51: We can remove the this paragraph too, since it is server-side implementation detail

src/sql/SqlStatement.ts
mdumandag: I meant removing the Long from the type and making it `timeoutMillis?: number;`. This cannot be changed later. 

I think this will simplify the API and implementation.

src/sql/SqlService.ts
mdumandag: Since the CodecUtil encodes `undefined` and `null` as a `NULL_FRAME`, I think just the thing I wrote will be enough.

src/sql/SqlStatement.ts
mdumandag: Yes. I didn't understand this was optional initially. So, instead of `null`, they will not provide it at all as you suggested. This is more common in JS right rather than setting something to `null`

src/sql/SqlService.ts
mdumandag: Yes, I think there still should be wrapper classes. I thought we decided to let users do things like `execute("SELECT * FROM map WHERE date < CAST(?  AS DATE)", myHzLocalDate.toString())` but not `execute("SELECT * FROM map WHERE date < CAST(?  AS DATE)", myHzLocalDate)` to make it more explicit

src/sql/SqlResult.ts
srknzl: not sure but will do, reasonable

src/sql/SqlService.ts
srknzl: what about 

Every table backed by an IMap has a set of columns that are resolved automatically. Column resolution uses IMap entries
 located on the member that initiates the query. The engine extracts columns from a key and a value and then merges them into a single column set. In case the key and the value have columns with the same name, the key takes precedence.

src/sql/SqlService.ts
srknzl: I think we should use default schema, I have a test that makes sure defaults are used if an option is not provided. So, I think

`sqlStatement.options?.hasOwnProperty('schema') ? sqlStatement.options.schema : SqlServiceImpl.DEFAULT_SCHEMA`

is ok, no need to over simplify.

you can search test with `should call encodeRequest with correct params`

src/sql/SqlService.ts
srknzl: We can do that if you want. No worries. Going to delete implicit string conversion

mdumandag: Don't to immediately :smile: I thought we can discuss this with the team in the standup

srknzl: Ok easy to revert that commit, no worries. Should we ask it in the slack channel?

mdumandag: This might be the first question to ask in tomorrow's standup

srknzl:
verify

srknzl:
verify

srknzl:
verify

src/sql/SqlResult.ts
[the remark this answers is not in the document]
srknzl: I added returning a Promise.resolve() code, this way I think there will be no problem. similar check also exists on java side in close().

src/sql/SqlService.ts
srknzl: I kept it for now

src/sql/SqlPage.ts
```diff
@@ -0,0 +1,68 @@
...
+     * Returns the value in certain row and column
+     */
+    getValue(rowIndex: number, columnIndex: number): any {
+        return this.columns[columnIndex][rowIndex];
+    }
+
+    /**
+     * This is needed for better mocking. Constructor mocking is not trivial with sinon.
+     */
+    static newPage(columnTypes: SqlColumnType[], columns: any[][], last: boolean): SqlPage {
```
alexb5dh, line 65: I think in Java method with similar signature is called `fromColumns`.

src/sql/SqlResult.ts
```diff
@@ -0,0 +1,389 @@
...
+ * const updateCount = await result.getUpdateCount();
+ * ```
+ *
+ * You don't need to call {@link close} in this case.
+ *
+ */
+export interface SqlResult extends AsyncIterable<SqlRowType> {
+    /**
+     *  Returns next {@link SqlRowType} iteration result. You should not call this method when result does not contains
+     *  rows.
```
alexb5dh, line 80:
```suggestion
     *  Returns next {@link SqlRowType} iteration result. You should not call this method when result does not contain
     *  rows.
```

.eslintrc
```diff
@@ -97,5 +97,8 @@
             ]
         }],
         "space-in-parens": "error"
+    },
```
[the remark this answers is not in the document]
srknzl: they are needed since we started to use BigInt which is supported in es2018. Also node version needs to be updated, bigint is supported 10.4 and above.

I think gitignore change is coming from another pr since I merged master into this pr a while ago

scripts/test-runner.js
```diff
@@ -51,70 +51,48 @@ const isAddressReachable = (host, port, timeoutMs) => {
     });
 };
 
-const startRC = async (background) => {
```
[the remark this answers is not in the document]
srknzl: running it in the background was hard to control afterwards, killing it etc. So npm run startrc now starts it in the foreground which is easier to use

src/core/HazelcastError.ts
```diff
@@ -468,3 +470,28 @@ export class UndefinedErrorCodeError extends HazelcastError {
         Object.setPrototypeOf(this, UndefinedErrorCodeError.prototype);
     }
 }
+
+/**
```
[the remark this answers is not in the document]
srknzl: it was like this before, I think it is ok though; since all classes are related. What's you proposition?

src/invocation/ClusterService.ts
```diff
@@ -75,12 +74,12 @@ export class ClusterService implements Cluster {
     /**
      * Gets the member with the given UUID.
      *
-     * @param uuid The UUID of the member.
-     * @return The member that was found, or undefined if not found.
+     * @param {string} uuid The UUID of the member as string.
+     * @return {MemberImpl | undefined} The member that was found, or undefined if not found.
      */
-    getMember(uuid: UUID): MemberImpl {
+    getMember(uuid: string): MemberImpl {
```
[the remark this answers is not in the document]
srknzl: This was related to change in `getRandomConnection`. I added `this.clusterService.getMember` to the method which returns a string. Converting from string to UUID is harder so, I changed this method

src/sql/SqlStatement.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * Properties are read once before the execution is started.
+ * Changes to properties do not affect the behavior of already running statements.
+ */
+export interface SqlStatement {
+    /**
+     * SQL string. The SQL string cannot be empty. SQL string placeholder character is question mark(`?`). A RangeError
+     * is thrown during execute if `sql` is not valid.
+     */
+    sql: string;
```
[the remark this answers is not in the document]
srknzl: RangeError is a common error for invalid arguments. What do you prefer?

src/sql/SqlPage.ts
srknzl: changed, makes sense, thanks

src/core/HazelcastError.ts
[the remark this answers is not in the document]
alexb5dh: As far as I know Java and .NET have each class in a separate file. But it looks like JS don't have this convention.

src/sql/SqlStatement.ts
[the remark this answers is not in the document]
alexb5dh: From the name I just thought that it's about index being out of bounds of the array, similar to what [ArgumentOutOfRangeException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is used for in C#.

So, for example in our client I would probably use something like [FormatException](https://docs.microsoft.com/en-us/dotnet/api/system.formatexception), just regular [ArgumentException](https://docs.microsoft.com/en-us/dotnet/api/system.argumentexception) or will create a custom one. But I'm not familiar with error/exception conventions in Node.

test/TestUtil.js
```diff
@@ -30,6 +29,30 @@ exports.promiseLater = function (time, func) {
     }));
 };
 
+/**
+ * Returns rejection reason if rejected, otherwise returns a dummy error.
+ */
+exports.getRejectionReasonOrDummy = async function (object, asyncMethodName, ...params) {
```
mdumandag, line 35: I think we could call it just `getRejectionReasonOrThrow` to make it easier to understand

test/TestUtil.js
```diff
@@ -30,6 +29,30 @@ exports.promiseLater = function (time, func) {
...
+    try {
+        await object[asyncMethodName](...params);
+        return new Error('dummy error');
```
mdumandag, line 38: we should throw here to make the test fail in case the method above does not throw
```suggestion
        throw new Error('Expected the call the throw, but it didn't.');
```

test/TestUtil.js
```diff
@@ -30,6 +29,30 @@ exports.promiseLater = function (time, func) {
     }));
 };
 
+/**
+ * Returns rejection reason if rejected, otherwise returns a dummy error.
+ */
+exports.getRejectionReasonOrDummy = async function (object, asyncMethodName, ...params) {
+    try {
+        await object[asyncMethodName](...params);
```
mdumandag, line 37: Instead of this, would it be more readable (both here and when we use it in tests) if we accept an async function?

So, it will be used as `await TestUtil.getRejectionReasonOrThrow(async () => await result.next())`

test/TestUtil.js
mdumandag: Similar comments apply to `getThrownErrorOrDummy` below

test/integration/backward_compatible/sql/ResultTest.js
```diff
@@ -0,0 +1,91 @@
...
+    });
+
+    afterEach(async function () {
+        await someMap.clear();
+    });
+
+    it('should reject iteration after close()', async function () {
+        await result.close();
+
+        try {
```
mdumandag, line 67: If we change the `getRejection` method above as I suggested, we could use it here as 
```
const error = await TestUtil.getRejectionReasonOrThrow(async () => {
    for await (const row of result) {};
});
```

and I feel this reads better

.eslintrc
```diff
@@ -98,5 +98,8 @@
         }],
         "space-in-parens": "error",
         "curly": "error"
+    },
+    "parserOptions": {
+        "ecmaVersion": 2018
```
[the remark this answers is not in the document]
srknzl: It is needed for for await loops in test folder. I will move this to eslintrc file in test folder.

test/.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
     "env": {
-        "mocha": true
+        "mocha": true,
+        "es2020": true
```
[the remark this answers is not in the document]
srknzl: needed for bigint constructor

test/integration/backward_compatible/sql/DecodeAndSerializationTest.js
```diff
@@ -0,0 +1,745 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const { Lang } = require('../../remote_controller/remote-controller_types');
+const { SqlColumnType } = require('../../../../lib/sql/SqlColumnMetadata');
```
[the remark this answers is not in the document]
srknzl: used getter like methods for imports

test/integration/backward_compatible/sql/DecodeAndSerializationTest.js
```diff
@@ -0,0 +1,745 @@
...
+    };
+
+    before(async function () {
+        TestUtil.markClientVersionAtLeast(this, '4.2');
+        cluster = await RC.createCluster(null, null);
+        await RC.startMember(cluster.id);
+    });
+
+    const basicSetup = async () => {
+        client = await Client.newHazelcastClient({
```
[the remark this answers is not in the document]
srknzl: addIndex method usage is also different in portable test, so I think this is fine like this

test/integration/backward_compatible/sql/DecodeAndSerializationTest.js
```diff
@@ -0,0 +1,745 @@
...
+            `SELECT * FROM ${mapName} WHERE this > ? AND this < ? ORDER BY __key ASC LIMIT ${expectedRowCount}`,
+            [long.fromNumber(10), long.fromNumber(16)]
+        );
+
+        const rows2 = [];
+
+        for await (const row of result2) {
+            rows2.push(row);
+        }
+        rows2.length.should.be.eq(expectedRowCount);
```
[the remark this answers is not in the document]
srknzl: done

test/integration/backward_compatible/sql/DecodeAndSerializationTest.js
```diff
@@ -0,0 +1,745 @@
...
+
+        for (let i = 0; i < rows.length; i++) {
+            (rows[i]['age'].eq(expectedValues[i].age)).should.be.true;
+            (rows[i]['height'] - expectedValues[i].height).should.be.lessThan(1e-5);
+            rows[i]['__key'].should.be.eq(expectedKeys[i]);
+        }
+    });
+});
+
+describe('Decode/Serialize portable with server config', function () {
```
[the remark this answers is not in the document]
srknzl: it is not but Sancar told me to test both cases, I think we can remove `with config` test

test/integration/backward_compatible/sql/ExecuteTest.js
```diff
@@ -0,0 +1,564 @@
...
+            await RC.terminateCluster(cluster.id);
+            await client.shutdown();
+        });
+
+        afterEach(async function () {
+            await someMap.clear();
+        });
+
+        // Sorts sql result rows by __key, first the smallest __key
+        const sortByKey = (array) => {
```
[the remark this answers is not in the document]
srknzl: removed

test/TestUtil.js
srknzl: it is the same thing since the error is catched and returned

test/TestUtil.js
```diff
@@ -30,6 +29,30 @@ exports.promiseLater = function (time, func) {
...
+        throw new Error('Expected the call the throw, but it didn\'t.');
+    } catch (e) {
+        return e;
+    }
+};
+
+/**
+ * Returns thrown error if thrown, otherwise returns a dummy error.
+ */
+exports.getThrownErrorOrDummy = function (fn) {
```
mdumandag, line 47: We should probably rename this to `getThrownErrorOrThrow`. Also, it would be good to update documentation of these two functions

mdumandag (approved):

test/integration/backward_compatible/note.md
```diff
@@ -0,0 +1,61 @@
+
```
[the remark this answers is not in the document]
mdumandag: this file is still there

test/TestUtil.js
srknzl: I think `returns a dummy error` is accurate. What do you prefer?

test/integration/backward_compatible/note.md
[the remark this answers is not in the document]
srknzl: is it? are you sure

test/TestUtil.js
mdumandag: Oh, I missed it. I think it would be much better if we change it to throw. This way, if `fn()` does not throw, the test will automatically fail with the error message below. Otherwise, the test may again fail but with a different error message such as expected a different error type, etc. which is not really intuitive.

test/integration/backward_compatible/sql/DecodeAndSerializationTest.js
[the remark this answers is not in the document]
mdumandag: Yes, I think we can remove it. From the client's perspective, they are the same.

test/TestUtil.js
srknzl: makes sense, same for getRejectionReason right?

mdumandag: yes

srknzl:
@mdumandag Thanks for the review

PR #862, opened 2021-04-19
Fix client version handling in the statistics tests

mdumandag:
A fix for the data we send during the statistics collection is
also backported to 4.0.2, so we need to account for that and
fix the client version handling.

This is a required change for the backward compatibility tests.


srknzl (approved):
LGTM

PR #863, opened 2021-04-20
Use assertTrueEventually instead of setTimeout for getDistributedObjects test

srknzl:
makes getDistributedObjects tests a little bit stronger, they do not rely on 300 of fixed timeout now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
         "pregenerate-docs": "rimraf docs",
         "generate-docs": "typedoc --options typedoc.json",
         "lint": "eslint --ext .ts src && eslint --plugin mocha test && eslint code_samples && eslint scripts",
+        "lint:fix": "eslint --fix --ext .ts src && eslint --fix --plugin mocha test && eslint --fix code_samples && eslint --fix scripts",
         "startrc": "node scripts/test-runner.js startrc"
     },
     "repository": {
--- a/test/integration/backward_compatible/HazelcastClientTest.js
+++ b/test/integration/backward_compatible/HazelcastClientTest.js
@@ -119,14 +119,13 @@ configParams.forEach(function (cfg) {
         it('getDistributedObjects returns all dist objects', function () {
             managed.getObject(client.getMap.bind(client, 'map'));
             managed.getObject(client.getSet.bind(client, 'set'));
-            TestUtil.assertTrueEventually(function () {
-                return client.getDistributedObjects().then(function (distObjects) {
-                    const names = distObjects.map((o) => {
-                        return o.getName();
-                    });
-                    expect(names).to.have.members(['map', 'set']);
+            return TestUtil.assertTrueEventually(async function () {
+                const distObjects = await client.getDistributedObjects();
+                const names = distObjects.map((o) => {
+                    return o.getName();
                 });
-            });
+                expect(names).to.have.members(['map', 'set']);
+            }, 100, 5000);
         });
 
         it('getDistributedObjects does not return removed object', function () {
@@ -134,16 +133,14 @@ configParams.forEach(function (cfg) {
             managed.getObject(client.getMap.bind(client, 'map2'));
             managed.getObject(client.getMap.bind(client, 'map3'));
 
-            TestUtil.assertTrueEventually(function () {
-                return managed.destroy('map1').then(function () {
-                    client.getDistributedObjects().then(function (distObjects) {
-                        const names = distObjects.map(function (o) {
-                            return o.getName();
-                        });
-                        expect(names).to.have.members(['map2', 'map3']);
-                    });
+            return TestUtil.assertTrueEventually( async function () {
+                await managed.destroy('map1');
+                const distObjects = await client.getDistributedObjects();
+                const names = distObjects.map(function (o) {
+                    return o.getName();
                 });
-            });
+                expect(names).to.have.members(['map2', 'map3']);
+            }, 100, 5000);
         });
     });
 });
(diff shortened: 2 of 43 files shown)


test/integration/backward_compatible/HazelcastClientTest.js
```diff
@@ -116,44 +116,34 @@ configParams.forEach(function (cfg) {
...
-                    try {
-                        const names = distObjects.map((o) => {
-                            return o.getName();
-                        });
-                        expect(names).to.have.members(['map', 'set']);
-                        done();
-                    } catch (e) {
-                        done(e);
-                    }
+            TestUtil.assertTrueEventually(function () {
```
mdumandag, line 122: I think you need to await the promise returned by the ` TestUtil.assertTrueEventually`. 

Also, you can make the callback an async function and await the `client.getDistributedObjects` call which would make it much more readable.

The same applies to other places

mdumandag (changes requested):

test/integration/backward_compatible/HazelcastClientTest.js
srknzl: Thanks, good catch. In addition to fixing them, I adjusted the timeout to be 5 seconds. It should be enough.

test/integration/backward_compatible/HazelcastClientTest.js
```diff
@@ -116,44 +116,31 @@ configParams.forEach(function (cfg) {
...
-                    try {
-                        const names = distObjects.map((o) => {
-                            return o.getName();
-                        });
-                        expect(names).to.have.members(['map', 'set']);
-                        done();
-                    } catch (e) {
-                        done(e);
-                    }
+            return TestUtil.assertTrueEventually(async function () {
```
mdumandag, line 122: Just to be consistent with the rest of the tests, can we use arrow functions there?

test/integration/backward_compatible/HazelcastClientTest.js
```diff
@@ -116,44 +116,31 @@ configParams.forEach(function (cfg) {
...
-                            const names = distObjects.map(function (o) {
-                                return o.getName();
-                            });
-                            expect(names).to.have.members(['map2', 'map3']);
-                            done();
-                        } catch (e) {
-                            done(e);
-                        }
-                    });
+            return TestUtil.assertTrueEventually( async function () {
```
mdumandag, line 136: Same comment as above. Also, there is an unnecessary space in `TestUtil.assertTrueEventually( async`.

test/integration/backward_compatible/HazelcastClientTest.js
```diff
@@ -116,44 +116,31 @@ configParams.forEach(function (cfg) {
...
+                await managed.destroy('map1');
+                const distObjects = await client.getDistributedObjects();
+                const names = distObjects.map(function (o) {
```
mdumandag, line 139: Same comment as above for arrow functions

mdumandag (approved):
Approving with minor styling comments

test/integration/backward_compatible/HazelcastClientTest.js
srknzl: would we want to enable https://eslint.org/docs/rules/prefer-arrow-callback flag

srknzl: we can do that just for tests

srknzl: seems there is a mocha friendly one too https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/prefer-arrow-callback.md

srknzl: we can add 

```
"mocha/prefer-arrow-callback": ["error", {
            "allowNamedFunctions": true
}]
```

to the .eslintrc in the test folder. This will allow named functions, mocha hooks. By default it won't change the function usage if it's behaviour would be different after the change.

mdumandag: Sounds good to me. It would be good to be consistent across all the tests

package.json
```diff
@@ -47,6 +47,7 @@
         "pregenerate-docs": "rimraf docs",
         "generate-docs": "typedoc --options typedoc.json",
         "lint": "eslint --ext .ts src && eslint --plugin mocha test && eslint code_samples && eslint scripts",
+        "lint:fix": "eslint --fix --ext .ts src && eslint --fix --plugin mocha test && eslint --fix code_samples && eslint --fix scripts",
```
srknzl, line 50: `npm run lint --fix` === `eslint --ext .ts src && eslint --plugin mocha test && eslint code_samples && eslint scripts --fix`

so it won't fix all the errors

so I added this one.

test/integration/backward_compatible/HazelcastClientTest.js
srknzl: >Also, there is an unnecessary space in TestUtil.assertTrueEventually( async.
To avoid this, I plan to add `"space-in-parens": "error"` to root eslint config.(no error with current code base)


ref: https://eslint.org/docs/rules/space-in-parens

srknzl:
Some additional changes I made here so a new review would be better:

* tests changed using `eslint --fix`
* Added lint:fix script
* I plan to add new rule(check out the comments above)

test/integration/backward_compatible/HazelcastClientTest.js
mdumandag: Sounds good to me

srknzl:
I added the new rule as well

mdumandag (approved):

PR #866, opened 2021-04-22
Jsdoc change in static client constructors

srknzl:
We return a promise not a client object.


mdumandag (commented):
I don't think we should make this change. Because, if you take a look at other places (such as IMap documentation), we do not mention the promise. I think we should be consistent across the code base about this.

srknzl:
I think we can convert all

srknzl:
Actually since typedoc takes return type and puts it in documentation, we can skip this change 

image

PR #871, opened 2021-04-27
[API-378] Make defaultNumberType usage clear

srknzl:
I encountered this while reading the documentation. 

The defaultNumberType option was not clear, this change makes it clear and allows it to accept string in any casing(uppercase or lowercase as opposed to accepting only lower case).

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -581,7 +581,7 @@ Hazelcast serializes all your objects before sending them to the server. Certain
 
 > **NOTE: The `Long` type means the type provided by [long.js library](https://github.com/dcodeIO/long.js).**
 
-> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the `serialization.defaultNumberType` config option. Refer to [API Documentation](http://hazelcast.github.io/hazelcast-nodejs-client/api/current/docs/) in order to learn how you can configure it**
+> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the [serialization.defaultNumberType](http://hazelcast.github.io/hazelcast-nodejs-client/api/4.1.0/docs/interfaces/_config_serializationconfig_.serializationconfig.html#defaultnumbertype) config option.
 
 Arrays of the `boolean`, `number`, `string`, and `Long` types can be serialized as `boolean[]`, `byte[]`, `short[]`, `int[]`, `float[]`, `double[]`, `string[]`, and `long[]` for the Java server side, respectively.
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -581,7 +581,7 @@ Hazelcast serializes all your objects before sending them to the server. Certain
 
 > **NOTE: The `Long` type means the type provided by [long.js library](https://github.com/dcodeIO/long.js).**
 
-> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the [serialization.defaultNumberType](http://hazelcast.github.io/hazelcast-nodejs-client/api/current/docs/interfaces/_config_serializationconfig_.serializationconfig.html#defaultnumbertype) config option.**
+> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the `serialization.defaultNumberType` config option. See [API Documentation](http://hazelcast.github.io/hazelcast-nodejs-client/api/current/docs/) for more information.**
 
 Arrays of the `boolean`, `number`, `string`, and `Long` types can be serialized as `boolean[]`, `byte[]`, `short[]`, `int[]`, `float[]`, `double[]`, `string[]`, and `long[]` for the Java server side, respectively.
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -581,7 +581,7 @@ Hazelcast serializes all your objects before sending them to the server. Certain
 
 > **NOTE: The `Long` type means the type provided by [long.js library](https://github.com/dcodeIO/long.js).**
 
-> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the `serialization.defaultNumberType` config option.**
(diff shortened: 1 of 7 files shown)


typedoc.json
```diff
@@ -1,11 +1,9 @@
...
-    "inputFiles": "src/",
+    "entryPoints": ["src/"],
     "exclude": "src/codec/**/*.ts",
-    "excludeNotExported": true,
     "excludeExternals": true,
-    "stripInternal": true,
-    "ignoreCompilerErrors": true,
+    "excludeInternal": true,
     "excludePrivate": true,
     "excludeProtected": true
```
srknzl, line 8: Refer to https://github.com/TypeStrong/typedoc/releases/tag/v0.20.0 for reasoning behind this change

src/config/SerializationConfig.ts
```diff
@@ -26,6 +26,18 @@ export interface SerializationConfig {
 
     /**
      * Defines how the `number` type is represented on the cluster side. By default, it is serialized as `Double`.
+     * This option can be one of the following strings. The casing of the string does not matter, it's converted to lower
+     * case internally.
+     * * `byte`(8-bit signed integer)
```
mdumandag, line 31: How is this rendered? (the `*`). IIRC, we should be using `<ul><li>elem1</li><li>elem2</li></ul>` like syntax for lists

src/config/SerializationConfig.ts
```diff
@@ -26,6 +26,18 @@ export interface SerializationConfig {
...
+     * * `short`(16-bit signed integer)
+     * * `integer`(32-bit signed integer)
+     * * `float`(single-precision 32-bit IEEE 754 floating point)
+     * * `double`(double-precision 64-bit IEEE 754 floating point)
+     * * `long`(64-bit integer from long.js library)
+     *
+     * Note:
+     * * If you are using byte, **for array of numbers** you need to use `Buffer.from([1,2,3])` instead of regular arrays.
```
mdumandag, line 39: I think it would be good to put `byte` into the backticks. A similar thing applies to `long`.

mdumandag, line 39: Can we use markdown in JSDoc(the `**for array of numbers**`)? I see that for it, we need a special plugin. But I think instead of it we can use `<b></b>`.

mdumandag, line 39: you need to use `Buffer.from([1,2,3])` -> you need to use `Buffer` objects

src/config/SerializationConfig.ts
```diff
@@ -26,6 +26,18 @@ export interface SerializationConfig {
...
     /**
      * Defines how the `number` type is represented on the cluster side. By default, it is serialized as `Double`.
+     * This option can be one of the following strings. The casing of the string does not matter, it's converted to lower
+     * case internally.
+     * * `byte`(8-bit signed integer)
+     * * `short`(16-bit signed integer)
+     * * `integer`(32-bit signed integer)
+     * * `float`(single-precision 32-bit IEEE 754 floating point)
+     * * `double`(double-precision 64-bit IEEE 754 floating point)
+     * * `long`(64-bit integer from long.js library)
```
mdumandag, line 36: Since we are mentioning about the server-side, I think just saying 64-bit signed integer would be enough instead of mentioning long.js

src/config/SerializationConfig.ts
srknzl: try and see :) markdown, for typedoc

src/config/SerializationConfig.ts
srknzl: again, used because of typedoc, is it a problem though?

srknzl: I checked out vscode and intellij, they are showing markdown properly in jsdoc. Other than that, we are using typedoc for api documentation; so it should be fine. Any modern IDE should support markdown in jsdoc(or any docstring). Therefore, I see no issue with using markdown.

https://typedoc.org/guides/doccomments/#markdown

src/config/SerializationConfig.ts
srknzl: This is not `exactly` server side. If user selects `long`, what he should use? We should state it here, this is a client related config.

DOCUMENTATION.md
```diff
@@ -581,7 +581,7 @@ Hazelcast serializes all your objects before sending them to the server. Certain
 
 > **NOTE: The `Long` type means the type provided by [long.js library](https://github.com/dcodeIO/long.js).**
 
-> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the `serialization.defaultNumberType` config option.**
+> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the `serialization.defaultNumberType` config option. Refer to [API Documentation](http://hazelcast.github.io/hazelcast-nodejs-client/api/current/docs/) in order to learn how you can configure it**
```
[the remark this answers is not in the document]
srknzl: Or better, let's refer to option itself

DOCUMENTATION.md
```diff
@@ -581,7 +581,7 @@ Hazelcast serializes all your objects before sending them to the server. Certain
 
 > **NOTE: The `Long` type means the type provided by [long.js library](https://github.com/dcodeIO/long.js).**
 
-> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the `serialization.defaultNumberType` config option.**
+> **NOTE: A `number` is serialized as `Double` by default. You can configure this behavior using the [serialization.defaultNumberType](http://hazelcast.github.io/hazelcast-nodejs-client/api/4.1.0/docs/interfaces/_config_serializationconfig_.serializationconfig.html#defaultnumbertype) config option.
```
mdumandag, line 584: Is there a way to not use the exact version number in the URL(something like `latest` instead of `4.1.0`)? We might forget to update the link sometime later, after the future releases

mdumandag: Also, the `**` is missing at the end of the sentence

srknzl: done thanks

srknzl: but looks full link does not work, this works `http://hazelcast.github.io/hazelcast-nodejs-client/api/current/docs`

mdumandag (approved):

PR #873, opened 2021-04-28
Add junit mocha reporter

srknzl:
With this we will be able to see running test and test results in jenkins.

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -6,9 +6,9 @@
         "test/user_code/**"
     ],
     "reporter": "mocha-jenkins-reporter",
-    "reporterOptions": {
-        "junit_report_name": "Node.js Client Tests",
-        "junit_report_path": "report.xml",
-        "junit_report_stack": 1
-    }
+    "reporter-option": [
+        "junit_report_name=Node.js Client Tests",
+        "junit_report_path=report.xml",
+        "junit_report_stack=1"
+    ]
 }
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -4,5 +4,11 @@
     "exclude": [
         "test/soak_test/**",
         "test/user_code/**"
-    ]
+    ],
+    "reporter": "mocha-jenkins-reporter",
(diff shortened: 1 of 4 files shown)


test/integration/backward_compatible/LoggingTest.js
```diff
@@ -69,15 +71,15 @@ describe('LoggingTest', function () {
...
+        consoleLogSpy = sinon.spy(console, 'log');
     });
 
     afterEach(async function () {
         if (client != null) {
             await client.shutdown();
             client = null;
         }
-        console.log.restore();
+        consoleLogSpy.restore();
```
srknzl, line 82: this reporter changes how console.log works, so we need to use variable here for this test to pass

package.json
```diff
@@ -21,6 +21,7 @@
         "eslint-plugin-mocha": "~8.1.0",
         "jsonschema": "~1.4.0",
         "mocha": "~8.3.2",
+        "mocha-jenkins-reporter": "^0.4.5",
```
mdumandag, line 24: Do we still need `mocha-junit-reporter": "~2.0.0",` dependency?

.mocharc.json
```diff
@@ -4,5 +4,11 @@
     "exclude": [
         "test/soak_test/**",
         "test/user_code/**"
+    ],
+    "reporter": "mocha-jenkins-reporter",
+    "reporter-option": [
+        "junit_report_name=Node.js Client Tests",
+        "junit_report_path=report.xml",
+        "junit_report_stack=1"
```
mdumandag, line 12: Will this affect the stack we see on test failures? It would be very helpful if we can see the full stack trace on test failures

srknzl: not it enables stack, we can see trace

package.json
srknzl: guess not

srknzl:
verify

mdumandag (approved):

PR #890, opened 2021-05-10
[API-377] Add reliable topic cluster restart tests

### LINKED ISSUE #758: [TRACKING ISSUE] Fix ClientReliableTopicOnClusterRestartTest

**Fix ClientReliableTopicOnClusterRestartTest**

See https://github.com/hazelcast/hazelcast/pull/16644 for details.

---

Fixes the test testing behaviour of a loss tolerant listener when the
instance is restarted. This test failure is similar to the one described
in https://github.com/hazelcast/hazelcast/issues/16430#issuecomment-571947692 and the [fix PR](https://github.com/hazelcast/hazelcast/pull/16440).

Basically, it might happen that the new item is added to the restarted
member before the listener sends a read operation to the new member.
Since it was previously on a higher sequence, the sequence will be
reset to the currently highest, which is one after the latest added item.

In the case when the item is added to the ringbuffer before the read
operation is executed, the listener will simply ignore the added item
and wait for a new one.

The test asserts that the listener is able to continue after the cluster
restarts so we modify the test so that we add new items until we
eventually get an item.

Also, made some minor modifications like using small instances in the
test and made the sequence field volatile since it is accessed from
different threads.

Fixes: https://github.com/hazelcast/hazelcast/issues/16623

srknzl:
fixes #758 

I added `shouldContinue_OnClusterRestart_whenDataLoss_LossTolerant_afterInvocationTimeout`, should we add more tests here?

### DIFF
--- a/test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
+++ b/test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
@@ -178,7 +178,6 @@ describe('ReliableTopicOnClusterRestartTest', function () {
         });
 
         await RC.shutdownMember(cluster.id, member.uuid);
-
         await RC.startMember(cluster.id);
 
         await topic2.publish('message');
--- a/test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
+++ b/test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
@@ -28,11 +28,23 @@ describe('ReliableTopicOnClusterRestartTest', function () {
     let member;
     let client1;
     let client2;
-    const invocationTimeoutMillis = 2000;
 
     before(async function () {
         cluster = await RC.createCluster(null, null);
         member = await RC.startMember(cluster.id);
+        client1 = undefined;
+        client2 = undefined;
+    });
+
+    after(async function() {
+        await RC.shutdownCluster(cluster.id);
+        if (client1) await client1.shutdown();
+        if (client2) await client2.shutdown();
+    });
+
+    it('should continue on cluster restart when data lost and after invocation timeout', async function () {
+        const invocationTimeoutMillis = 2000;
+
         client1 = await Client.newHazelcastClient({
             clusterName: cluster.id,
             connectionStrategy: {
@@ -55,15 +67,7 @@ describe('ReliableTopicOnClusterRestartTest', function () {
                 'hazelcast.client.invocation.timeout.millis': invocationTimeoutMillis
             }
         });
-    });
-    after(async function() {
-        await RC.shutdownCluster(cluster.id);
-        await client1.shutdown();
-        await client2.shutdown();
-    });
 
-    // https://github.com/hazelcast/hazelcast/pull/16644
-    it('should continue on cluster restart after invocation timeout', async function () {
         let messageArrived = false;
         let messageCount = 0;
 
@@ -75,12 +79,11 @@ describe('ReliableTopicOnClusterRestartTest', function () {
 
         const topic1 = await client1.getReliableTopic(topicName);
         topic1.addMessageListener(() => {
-            console.log('received message');
             messageCount++;
             messageArrived = true;
         });
 
-        await RC.terminateMember(cluster.id, member.uuid);
+        await RC.shutdownMember(cluster.id, member.uuid);
 
         await RC.startMember(cluster.id);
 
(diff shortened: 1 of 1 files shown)


srknzl:
verify

test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
```diff
@@ -0,0 +1,194 @@
...
+    });
+
+    after(async function() {
+        await RC.shutdownCluster(cluster.id);
+    });
+
+    it('should continue on cluster restart when data lost and after invocation timeout', async function () {
+        const invocationTimeoutMillis = 2000;
+
+        client1 = await Client.newHazelcastClient({
```
mdumandag, line 54: I think we can move client construction to a function to eliminate duplicate code

test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
```diff
@@ -0,0 +1,194 @@
...
+        // wait some time for subscription
+        await promiseWaitMilliseconds(invocationTimeoutMillis);
+
+        await assertTrueEventually(async () => {
+            const topic3 = await client2.getReliableTopic(topicName);
+            await topic3.publish(`newItem${randomString()}`);
+            await assertTrueEventually(async () => {
+                if (!messageArrived)
+                    throw new Error('message not arrived yet');
+            }, undefined, 5000);
```
mdumandag, line 104:
```suggestion
            await assertTrueEventually(() => messageArrived.should.be.true, undefined, 5000);
```

test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
```diff
@@ -0,0 +1,194 @@
...
+    afterEach(async function () {
+        if (client1) await client1.shutdown();
+        if (client2) await client2.shutdown();
+    });
+
+    after(async function() {
+        await RC.shutdownCluster(cluster.id);
+    });
+
+    it('should continue on cluster restart when data lost and after invocation timeout', async function () {
```
mdumandag, line 51:
```suggestion
    it('should continue on cluster restart when data loss and after invocation timeout', async function () {
```

test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
```diff
@@ -0,0 +1,194 @@
...
+            connectionStrategy: {
+                connectionRetry: {
+                    clusterConnectTimeoutMillis: Number.MAX_SAFE_INTEGER
+                }
+            }
+        });
+
+        const topic2 = await client2.getReliableTopic(topicName);
+        await topic2.publish('message');
+        await assertTrueEventually(async () => {
```
mdumandag, line 151: similar thing as above

test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
```diff
@@ -0,0 +1,194 @@
...
+
+        topic1.addMessageListener(() => {
+            messageArrived = true;
+        });
+
+        await RC.shutdownMember(cluster.id, member.uuid);
+        await RC.startMember(cluster.id);
+
+        await topic2.publish('message');
+        await assertTrueEventually(async () => {
```
mdumandag, line 189: similar thing as above

srknzl: these are done in https://github.com/hazelcast/hazelcast-nodejs-client/pull/890/commits/c6f8d1483c1ace172ab81f96dab1e495382cc8ad

mdumandag (approved):

test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
```diff
@@ -0,0 +1,159 @@
...
+        member = await RC.startMember(cluster.id);
+    });
+
+    beforeEach(async function () {
+        client1 = undefined;
+        client2 = undefined;
+    });
+
+    afterEach(async function () {
+        if (client1) await client1.shutdown();
```
mdumandag, line 43: BTW, can you enable https://eslint.org/docs/rules/curly and https://eslint.org/docs/rules/keyword-spacing? So, we don't have such statements

srknzl: ok will do in a seperate pr, opening an issue

PR #896, opened 2021-05-12
[API-374] Fire membership events when hot restart persistence is enabled

### LINKED ISSUE #805: [TRACKING ISSUE] Fire membership events when persistence enabled

See https://github.com/hazelcast/hazelcast/pull/18245 for details.

---

When persistence enabled, client does not fire membership
events because members starts with same uuid's.

With this pr, we make sure that when cluster restarted,
we always update the local memberlist with empty member list
first.
This way, we will be able to fire member removed/added events
even if member uuid's does not change.

Note that if cluster uuid does not change, we will not fire
any event. This happens client disconnected and connected
back to the same cluster.
Note that cluster uuid is not preserved on hotrestart with
presistence.

There was already a test to verify this behaviour but it
was working wrong.
"ClientHotRestartTest extends ClientClusterRestartEventTest"
In this pr, we make sure that the test uses same address
for the restarted member to test this correctly.

fixes https://github.com/hazelcast/hazelcast/issues/18234

srknzl:
fixes #805

### DIFF
--- a/src/invocation/ClusterService.ts
+++ b/src/invocation/ClusterService.ts
@@ -156,6 +156,14 @@ export class ClusterService implements Cluster {
         }
     }
 
+    clearMemberList(connectionRegistry: ConnectionRegistry): void {
+        this.logger.trace('ClusterService', 'Resetting the member list.');
+        const previousMembers = this.memberListSnapshot.memberList;
+        this.memberListSnapshot = EMPTY_SNAPSHOT;
+        const events = this.detectMembershipEvents(previousMembers, [], connectionRegistry);
+        this.fireEvents(events);
+    }
+
     handleMembersViewEvent(
         connectionRegistry: ConnectionRegistry,
         memberListVersion: number,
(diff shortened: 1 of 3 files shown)


src/invocation/ClusterService.ts
```diff
@@ -151,11 +151,25 @@ export class ClusterService implements Cluster {
...
         }
     }
 
+    clearMemberList(connectionRegistry: ConnectionRegistry): void {
+        this.logger.trace('ClusterService', 'Resetting the member list.');
+        // This check is necessary so that when handling auth response, it will not
+        // intervene with client failover logic
+        if (this.memberListSnapshot !== EMPTY_SNAPSHOT) {
+            const previousMembers = this.memberListSnapshot.memberList;
+            this.memberListSnapshot = Object.assign({}, EMPTY_SNAPSHOT);
```
mdumandag, line 167: We should set the snapshot to something that has version 0, and an empty map

test/integration/backward_compatible/ClientHotRestartEventTest.js
```diff
@@ -80,4 +80,45 @@ describe('ClientHotRestartEventTest', function () {
...
+                removedMemberUUID = event.member.uuid.toString();
+                memberRemoved = true;
+            }
+        });
+
+        await RC.shutdownMember(cluster.id, oldMember.uuid);
+        const newMember = await RC.startMember(cluster.id);
+
+        await TestUtil.assertTrueEventually(async () => {
+            if (!memberRemoved) throw new Error('');
```
mdumandag, line 110: Can't we use something like `memberRemoved.should.be.true` instead of this?

srknzl: yeah I think that would look better

mdumandag (approved):

PR #929, opened 2021-06-07
SQL related improvements

srknzl:
### Changes

* If connection is not alive, client's uuid is used instead of remote uuid. fixes that. Note: Due to this change I think we should backport this to 4.2.x 

* Changes isRowSet logic to make it faster

* Doc or API doc changes

* Changes 'should reject iteration after close()' test to run close() at the middle of iteration with page size of 1. It was flaky in the previous version because if the result is received before close() is called, it will fail. Also adds catch() to rejected fetchPromise to suppress UnhandledPromiseRejectionWarning  Note: Also due to this change I think we should backport this to 4.2.x
* ignore empty modules from api doc

### DIFF
--- a/src/sql/SqlPage.ts
+++ b/src/sql/SqlPage.ts
@@ -54,7 +54,6 @@ export class SqlPage {
 
     /**
      * This is needed for better mocking. Constructor mocking is not trivial with sinon.
-     * @internal
      */
     static fromColumns(columnTypes: SqlColumnType[], columns: any[][], last: boolean): SqlPage {
         return new SqlPage(columnTypes, columns, last);
--- a/src/sql/SqlPage.ts
+++ b/src/sql/SqlPage.ts
@@ -53,6 +53,7 @@ export class SqlPage {
 
     /**
      * This is needed for better mocking. Constructor mocking is not trivial with sinon.
+     * @internal
      */
     static fromColumns(columnTypes: SqlColumnType[], columns: any[][], last: boolean): SqlPage {
         return new SqlPage(columnTypes, columns, last);
--- a/src/sql/SqlPage.ts
+++ b/src/sql/SqlPage.ts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/** @ignore *//** */
 
 import {SqlColumnType} from './SqlColumnMetadata';
 
--- a/src/sql/SqlResult.ts
+++ b/src/sql/SqlResult.ts
@@ -211,7 +211,7 @@ export class SqlResultImpl implements SqlResult {
 
     isRowSet(): Promise<boolean> {
         return this.executeDeferred.promise.then(() => {
-            return this.updateCount === Long.fromInt(-1);
+            return this.rowMetadata !== null;
         });
     }
 
--- a/src/sql/SqlResult.ts
+++ b/src/sql/SqlResult.ts
@@ -189,7 +189,6 @@ export class SqlResultImpl implements SqlResult {
     /**
      * Useful for mocking. (Constructor mocking is hard/impossible)
      * @returns new result object.
-     * @internal
      */
     static newResult(
         sqlService: SqlServiceImpl,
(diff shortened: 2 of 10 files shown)


src/sql/SqlPage.ts
```diff
@@ -53,6 +54,7 @@ export class SqlPage {
 
     /**
      * This is needed for better mocking. Constructor mocking is not trivial with sinon.
+     * @internal
```
mdumandag, line 57: Do we need this, since the class is marked as internal?

src/sql/SqlResult.ts
```diff
@@ -211,7 +212,7 @@ export class SqlResultImpl implements SqlResult {
 
     isRowSet(): Promise<boolean> {
         return this.executeDeferred.promise.then(() => {
-            return this.updateCount === Long.fromInt(-1);
+            return this.rowMetadata !== null;
```
mdumandag, line 215: just to be sure, `rowMetadata` can't be `undefined`, right?

src/sql/SqlResult.ts
```diff
@@ -242,6 +243,7 @@ export class SqlResultImpl implements SqlResult {
         // Prevent ongoing/future fetch requests
         if (!this.fetchDeferred) {
             this.fetchDeferred = deferredPromise<SqlPage>();
+            this.fetchDeferred.promise.catch(()=>{});
```
mdumandag, line 246:
```suggestion
            this.fetchDeferred.promise.catch(() => {});
```

src/sql/SqlResult.ts
```diff
@@ -255,7 +257,11 @@ export class SqlResultImpl implements SqlResult {
         return this.closeDeferred.promise;
     }
 
-    /** Called when next page of the result is received. */
+    /**
+     * Called when next page of the result is received.
+     * @param page
+     * @internal
```
mdumandag, line 263: Do we need to mark private functions as internal?

src/sql/SqlResult.ts
```diff
@@ -282,6 +292,7 @@ export class SqlResultImpl implements SqlResult {
     /**
      * Used by {@link next}.
      * @returns the current row.
+     * @internal
```
mdumandag, line 295: Do we need to mark private functions as internal?

src/sql/SqlResult.ts
```diff
@@ -372,6 +389,7 @@ export class SqlResultImpl implements SqlResult {
     /**
      * Used by {@link next}.
      * @returns if there are rows to be iterated.
+     * @internal
```
mdumandag, line 392: Do we need to mark private functions as internal?

src/sql/SqlService.ts
```diff
@@ -378,11 +380,24 @@ export class SqlServiceImpl implements SqlService {
         return this.executeStatement(sqlStatement);
     }
 
+    /**
+     * Sends a close request on a connection for an SQL result using its query id.
+     * @param connection The connection the request will be sent to
+     * @param queryId The query id that defines the SQL result
+     * @internal
```
mdumandag, line 387: I see that this class is internal, do we need to mark methods as internal again?

src/sql/SqlPage.ts
srknzl: 👍🏻  no we don't

src/sql/SqlResult.ts
srknzl: no it cant be

src/sql/SqlResult.ts
srknzl: also the class is internal anyway, going to delete it

srknzl: I am going to delete @internal labels of all methods in an internal class

src/sql/SqlResult.ts
```diff
@@ -350,7 +362,6 @@ export class SqlResultImpl implements SqlResult {
 
     /**
      * Checks if there are rows to iterate in a recursive manner, similar to a non-blocking while block
-     * @internal
```
srknzl, line 353: I guess we had this before, therefore I added other internal labels. Let's delete this one too

src/sql/SqlService.ts
```diff
@@ -253,15 +251,15 @@ export class SqlServiceImpl implements SqlService {
 
     /**
      * Converts an error to HazelcastSqlException and returns it. Used by execute, close and fetch
-     * @internal
```
srknzl, line 256: Same here we had them before, these led me to add internal labels to other methods

mdumandag (approved):

srknzl:
verify

PR #938, opened 2021-06-14
Add client version check to hot restart test for back compatibility

srknzl:
ref https://github.com/hazelcast/hazelcast-python-client/pull/412#discussion_r650782366

### DIFF
--- a/test/integration/backward_compatible/ClientHotRestartEventTest.js
+++ b/test/integration/backward_compatible/ClientHotRestartEventTest.js
@@ -57,7 +57,9 @@ describe('ClientHotRestartEventTest', function () {
     });
 
     afterEach(async function () {
-        await client.shutdown();
+        if (client) {
+            await client.shutdown();
+        }
         await RC.terminateCluster(cluster.id);
     });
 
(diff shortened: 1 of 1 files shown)


test/integration/backward_compatible/ClientHotRestartEventTest.js
```diff
@@ -82,6 +82,7 @@ describe('ClientHotRestartEventTest', function () {
     });
 
     it('should receive membership events when the member is restarted on same address', async function () {
+        TestUtil.markClientVersionAtLeast(this, '4.2');
```
mdumandag, line 85: We should probably update `afterEach` as well, since the client won't be constructed

srknzl: good catch 👍🏻

test/integration/backward_compatible/ClientHotRestartEventTest.js
```diff
@@ -57,7 +57,9 @@ describe('ClientHotRestartEventTest', function () {
     });
```
srknzl, line 57: I want to add client = undefined here to be sure

srknzl: because first test may set client, the second might skip. in this case client.shutdown will be called again on the same client

mdumandag: yes, I was about the say the same thing

mdumandag (approved):

PR #949, opened 2021-06-21
Fix tests for back compatibility of 4.1.4 imdg release

srknzl:
* adds if checks to client and cluster shutdown on SQL tests.

* fixes ReliableTopicOnClusterRestartTest, member was being created in before, where it should be created in beforeEach

### DIFF
--- a/test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
+++ b/test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
@@ -29,13 +29,10 @@ describe('ReliableTopicOnClusterRestartTest', function () {
     let client1;
     let client2;
 
-    before(async function () {
-        cluster = await RC.createCluster(null, null);
-    });
-
     beforeEach(async function () {
         client1 = undefined;
         client2 = undefined;
+        cluster = await RC.createCluster(null, null);
         member = await RC.startMember(cluster.id);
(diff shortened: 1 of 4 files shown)


test/integration/backward_compatible/topic/ReliableTopicOnClusterRestart.js
```diff
@@ -31,12 +31,12 @@ describe('ReliableTopicOnClusterRestartTest', function () {
 
     before(async function () {
         cluster = await RC.createCluster(null, null);
-        member = await RC.startMember(cluster.id);
     });
 
     beforeEach(async function () {
         client1 = undefined;
         client2 = undefined;
+        member = await RC.startMember(cluster.id);
```
mdumandag, line 39: I think to make sure that the tests do not affect each other we should move `cluster = await RC.createCluster(null, null);` to beforeEach (and its shutdown to afterEach) as well

srknzl: Makes sense, done

mdumandag (approved):

srknzl:
I also simplified some util methods in https://github.com/hazelcast/hazelcast-nodejs-client/pull/949/commits/0596696e49da289adceadbd1f9383daa4e18fd0c

PR #953, opened 2021-06-24
Fix test that asserts originated member uuid

srknzl:
Updates client version to 5.0(should we do it? we should do for the test to run in master ci tests) and adds client version check of 4.2.1 to a test to fix back compatibility tests.

failing: https://github.com/srknzl/test/actions/runs/967207890
after tests update, the one that should pass: https://github.com/srknzl/test/actions/runs/967371138

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "hazelcast-client",
-    "version": "4.2.0",
+    "version": "5.0",
     "description": "Hazelcast - open source In-Memory Data Grid - client for Node.js",
     "main": "lib/index.js",
     "types": "lib/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "hazelcast-client",
-    "version": "5.0",
+    "version": "5.0.0",
     "description": "Hazelcast - open source In-Memory Data Grid - client for Node.js",
     "main": "lib/index.js",
     "types": "lib/index.d.ts",
--- a/test/integration/backward_compatible/sql/ExecuteTest.js
+++ b/test/integration/backward_compatible/sql/ExecuteTest.js
@@ -518,7 +518,7 @@ describe('SqlExecuteTest', function () {
             if (TestUtil.isClientVersionAtLeast('5.0.0')) {
                 error1.originatingMemberId.should.be.eq(client.connectionManager.getClientUuid());
             } else {
-                error1.originatingMemberId.should.be.eq(member.uuid);
+                error1.originatingMemberId.toString().should.be.eq(member.uuid);
             }
         });
 
@@ -551,7 +551,7 @@ describe('SqlExecuteTest', function () {
             if (TestUtil.isClientVersionAtLeast('5.0.0')) {
                 error1.originatingMemberId.should.be.eq(client.connectionManager.getClientUuid());
             } else {
-                error1.originatingMemberId.should.be.eq(member.uuid);
+                error1.originatingMemberId.toString().should.be.eq(member.uuid);
             }
         });
 
--- a/test/integration/backward_compatible/sql/ExecuteTest.js
+++ b/test/integration/backward_compatible/sql/ExecuteTest.js
@@ -499,6 +499,7 @@ describe('SqlExecuteTest', function () {
             client = await Client.newHazelcastClient({
                 clusterName: cluster.id
             });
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "hazelcast-client",
-    "version": "4.2.0",
+    "version": "5.0",
```
mdumandag, line 3: This should be `5.0.0` to follow the semantic versioning

test/integration/backward_compatible/sql/ExecuteTest.js
```diff
@@ -499,6 +499,7 @@ describe('SqlExecuteTest', function () {
             client = await Client.newHazelcastClient({
                 clusterName: cluster.id
             });
+            TestUtil.markClientVersionAtLeast(this, '4.2.1');
```
mdumandag, line 502: We haven't released the `4.2.1` client yet, and maybe we never will :smile: 

I think we could make this `5.0.0` and update it if we release the 4.2.1

mdumandag: Also, can you put the comment you put below to this as well?

test/integration/backward_compatible/sql/ExecuteTest.js
```diff
@@ -499,6 +499,10 @@ describe('SqlExecuteTest', function () {
             client = await Client.newHazelcastClient({
                 clusterName: cluster.id
             });
+            // There was a fix regarding originatingMemberId in 5.0 and 4.2.x after 4.2.
+            // TODO: Update here to be 4.2.1 if we release 4.2.1
+            // https://github.com/hazelcast/hazelcast-nodejs-client/pull/940
+            TestUtil.markClientVersionAtLeast(this, '5.0.0');
```
mdumandag, line 505: Just a thought, what do you think about not skipping this test for 5.0- clients, but change the assertion according to the client version? So for 4.2 client, we could use member.uuid, for 5.0 we could use client uuid?

srknzl: I thought of doing that, but I think the correct uuid is client uuid, isn't it? But at the same time, I think assertion of originatingMemberId is just a single part of this test and we should run the other parts for 4.2 clients.

In total, I think we can change the assertion as you suggested

mdumandag: Yes, maybe don't check the originatingMemberUuid on 5.0- clients, but do other assertions on all versions

mdumandag: Sorry, I didn't saw your last commit, I am fine with different assertions on different versions

srknzl: Ok. Let's go with the current version in 78b0c3f

srknzl:
passing test on 4.2.0 client against 4.2 server 

https://github.com/srknzl/test/runs/2903760274?check_suite_focus=true

mdumandag (approved):

test/integration/backward_compatible/sql/ExecuteTest.js
```diff
@@ -538,7 +545,14 @@ describe('SqlExecuteTest', function () {
...
             error1.should.be.instanceof(getHazelcastSqlException());
             error1.code.should.be.eq(getSqlErrorCode().CONNECTION_PROBLEM);
-            error1.originatingMemberId.should.be.eq(client.connectionManager.getClientUuid());
+            // There was a fix regarding originatingMemberId in 5.0 and 4.2.x after 4.2.0.
+            // TODO: Update here to be 4.2.1 if we release 4.2.1
+            // https://github.com/hazelcast/hazelcast-nodejs-client/pull/940
+            if (TestUtil.isClientVersionAtLeast('5.0.0')) {
+                error1.originatingMemberId.should.be.eq(client.connectionManager.getClientUuid());
+            } else {
+                error1.originatingMemberId.should.be.eq(member.uuid);
```
srknzl, line 554: should we need toString here? @mdumandag see https://github.com/srknzl/test/runs/2904025971?check_suite_focus=true

srknzl: This is the updated version that passes:

https://github.com/srknzl/test/runs/2904280680?check_suite_focus=true

see tests branch in srknzl/test for my latest commit

PR #959, opened 2021-06-28
Add config checks

### LINKED ISSUE #925: Enhancement: warn bad configuration
On 4.1.0, this is accepted but ignored without warning

```
    serializationCfg: {
        defaultNumberType: 'integer'    
    },
```

(as it should be `serialization`)

srknzl:
Originated from and closes #925

### DIFF
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -180,7 +180,7 @@ const DEFAULT_PROPERTIES: Properties = {
 };
 
 /** @internal */
-export const PROPERTY_LIST : string[] = Object.keys(DEFAULT_PROPERTIES);
+export const PROPERTY_LIST = new Set(Object.keys(DEFAULT_PROPERTIES));
 
 /** @internal */
 export class ClientConfigImpl implements ClientConfig {
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -155,32 +155,37 @@ export interface ClientConfig {
 }
 
 /** @internal */
+const defaultProperties: Properties = {
+    'hazelcast.client.heartbeat.interval': 5000,
+    'hazelcast.client.heartbeat.timeout': 60000,
+    'hazelcast.client.invocation.retry.pause.millis': 1000,
+    'hazelcast.client.invocation.timeout.millis': 120000,
+    'hazelcast.client.internal.clean.resources.millis': 100,
+    'hazelcast.client.cloud.url': 'https://coordinator.hazelcast.cloud',
+    'hazelcast.client.statistics.enabled': false,
+    'hazelcast.client.statistics.period.seconds': Statistics.PERIOD_SECONDS_DEFAULT_VALUE,
+    'hazelcast.invalidation.reconciliation.interval.seconds': 60,
+    'hazelcast.invalidation.max.tolerated.miss.count': 10,
+    'hazelcast.invalidation.min.reconciliation.interval.seconds': 30,
+    'hazelcast.logging.level': 'INFO',
+    'hazelcast.client.autopipelining.enabled': true,
+    'hazelcast.client.autopipelining.threshold.bytes': 65536,
+    'hazelcast.client.socket.no.delay': true,
+    'hazelcast.client.shuffle.member.list': true,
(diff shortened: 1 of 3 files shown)


src/config/ConfigBuilder.ts
```diff
@@ -207,22 +222,23 @@ export class ConfigBuilder {
...
+            }
             this.effectiveConfig.properties[key] = jsonObject[key];
         }
     }
 
     private handleLifecycleListeners(jsonObject: any): void {
         const listenersArray = tryGetArray(jsonObject);
-        for (const index in listenersArray) {
-            const listener = listenersArray[index];
+        for (const listener of listenersArray) {
```
srknzl, line 234: I believe this is simple and maybe faster

src/config/ConfigBuilder.ts
```diff
@@ -255,72 +271,82 @@ export class ConfigBuilder {
...
             }
             this.effectiveConfig.nearCaches[nearCacheConfig.name] = nearCacheConfig;
         }
     }
 
     private handleReliableTopics(jsonObject: any): void {
-        for (const key in jsonObject) {
-            const jsonRtCfg = jsonObject[key];
+        for (const name in jsonObject) {
+            const jsonRtCfg = jsonObject[name];
```
srknzl, line 320: renamed key to name(because it is reliable topic name) and name to key below

src/config/ConfigBuilder.ts
```diff
@@ -93,6 +93,8 @@ export class ConfigBuilder {
                 this.handleCredentials(value);
             } else if (key === 'backupAckToClientEnabled') {
                 this.effectiveConfig.backupAckToClientEnabled = tryGetBoolean(value);
+            } else {
+                console.log(`Warning: Unexpected config key '${key}' is passed to the Hazelcast Client`);
```
mdumandag, line 97: We don't use `console.log` in the client code but use loggers

src/config/ConfigBuilder.ts
```diff
@@ -207,22 +222,23 @@ export class ConfigBuilder {
 
     private handleProperties(jsonObject: any): void {
         for (const key in jsonObject) {
+            if (!PROPERTY_LIST.includes(key)) {
```
mdumandag, line 225: Can we make `PROPERTY_LIST` a set to make lookups faster?

src/config/ConfigBuilder.ts
```diff
@@ -181,6 +191,11 @@ export class ConfigBuilder {
     }
 
     private handleSSL(jsonObject: any): void {
+        for (const key in jsonObject) {
+            if (!['enabled', 'sslOptions', 'sslOptionsFactory', 'sslOptionsFactoryProperties'].includes(key)) {
```
mdumandag, line 195: I think it would be good to either use SSLConfig keys here or somehow define this and reuse it, instead of such usage

src/config/ConfigBuilder.ts
srknzl: We cannot use logger since when config is being built there is no logger.

srknzl: Logger depends on config. So config should be built in order a logger to be initialized

src/config/Config.ts
```diff
@@ -155,32 +155,36 @@ export interface ClientConfig {
...
-        'hazelcast.client.socket.no.delay': true,
-        'hazelcast.client.shuffle.member.list': true,
-        'hazelcast.client.operation.backup.timeout.millis': 5000,
-        'hazelcast.client.operation.fail.on.indeterminate.state': false,
-        // `null` is set as the default value here to use this property
-        // as a tri-state boolean (see TranslateAddressProvider)
-        'hazelcast.discovery.public.ip.enabled': null,
-    };
+/** @internal */
+export const PROPERTY_LIST = new Set(Object.keys(DEFAULT_PROPERTIES));
```
mdumandag, line 183: This is not a list anymore

src/config/ConfigBuilder.ts
mdumandag: Then I would throw exceptions on these cases

mdumandag (approved):

PR #963, opened 2021-06-30
API-574 API-581 SQL 5.0

srknzl:
Adds default serializers for date/time classes
Renames getSqlService to getSql
Adjusts tests
Adds jet sql tests

Docs & code samples pr will follow this pr


src/core/BigDecimal.ts
```diff
@@ -0,0 +1,326 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * A wrapper for BigDecimal values. You can use this class to store and query BigDecimal values in distributed objects.
+ *
+ * A `BigDecimal` consists of an arbitrary precision number {@link unscaledValue} and a {@link scale}. If zero or positive,
```
mdumandag, line 20:
```suggestion
 * A BigDecimal consists of an arbitrary precision number {@link unscaledValue} and a {@link scale}. If zero or positive,
```
Or put the other places we say `BigDecimal` into backticks

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,326 @@
...
+    private static readonly MAX_EXPONENT_DIGITS = Number.MAX_SAFE_INTEGER.toString().length;
+
+    /**
+     * Unscaled value of this `BigDecimal`. This value is a native JavaScript BigInt object, which is used to store
+     * digits of this BigDecimal.
+     */
+    readonly unscaledValue: BigInt;
+
+    /**
+     * Scale of this `BigDecimal`.  If zero or positive, the scale
```
mdumandag, line 39:
```suggestion
     * Scale of this `BigDecimal`. If zero or positive, the scale
```

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,326 @@
...
+        const adjustedScale = scl - exp;
+        if (!Number.isSafeInteger(adjustedScale)) {
+            throw new RangeError('Scale out of range.');
+        }
+        scl = adjustedScale;
+        return scl;
+    }
+
+    /** @internal */
+    static _new(bigDecimalString: string): BigDecimal {
```
mdumandag, line 107: Maybe new -> from

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,326 @@
...
+            offset++;
+            len--;
+        }
+
+        let dot = false;
+        let exp = 0;
+        let c: string;
+        let idx = 0;
+        const coeff = [];
+        for (; len > 0; offset++, len--) {
```
mdumandag, line 132: I think while loop might be more readable but I am fine with this

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,326 @@
...
+
+interface BigInterface {
+    (decimalString: string): BigDecimal;
+
+    new(decimalString: string): BigDecimal;
+}
+
+function _Big(decimalString: string): BigDecimal {
+    return BigDecimal._new(decimalString);
+}
```
mdumandag, line 270: Why do we have this? Wouldn't a public constructor with `constructor(unscaledValue: BigInt, scale: number)` and a static factory method like `BigDecimal.from(decimalString: string)` would suffice?

The name `BigInterface`, or the exported `Big` does not ring a bell for me. I think they don't tell much about what they are

src/core/DatetimeClasses.ts
```diff
@@ -0,0 +1,475 @@
...
+            nanoStr = nanoStr.padEnd(9, '0');
+            nano = +nanoStr;
+        }
+
+        if (isNaN(hours) || isNaN(minutes) || isNaN(seconds) || isNaN(nano)) {
+            throw new RangeError('Illegal time string. Expected a string in HH:mm:ss.SSS format');
+        }
+        return new HzLocalTime(hours, minutes, seconds, nano);
+    }
+
```
mdumandag, line 85: should we provide asDate function that converts this to native Date object with zero values on year mont day etc.?

mdumandag: Maybe similarly a function like fromDate?

src/core/DatetimeClasses.ts
```diff
@@ -0,0 +1,475 @@
...
+/**
+ * Constructor function, can be invoked with or without new keyword.
+ *
+ * @param year Must be between -999999999-999999999
+ * @param month Must be between 1-12
+ * @param date Must be between 1-31 depending on year and month
+ * @throws RangeError if any of the arguments are invalid
+ */
+export const LocalDate: HzLocalDateInterface = <HzLocalDateInterface>_LocalDate;
```
mdumandag, line 417: Why do we have those? How it is different than the constructor we have? If you want to provide a static factory method, then make it public on the class itself.

I don't think this abstraction is useful

Similar for others

mdumandag (commented):
I haven't reviewed the whole thing but went over it quickly. I will do the detailed review once the big stones are out of the way

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,326 @@
...
+            len--;
+        }
+
+        let dot = false;
+        let exp = 0;
+        let c: string;
+        let idx = 0;
+        const coeff = [];
+        for (; len > 0; offset++, len--) {
+            c = bigDecimalString[offset];
```
mdumandag, line 133: This code is possibly inspired from the JDK and frankly, quite hard to understand. Can we put the comments available on there to here to make the life of future readers easier?

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,326 @@
...
+        }
+        const stringValue = coeff.join('');
+        if (isneg) {
+            bigIntValue = BigInt('-' + stringValue);
+        } else {
+            bigIntValue = BigInt(stringValue);
+        }
+        return new BigDecimal(bigIntValue, scale);
+    }
+
```
mdumandag, line 184:
```suggestion
```

src/core/BigDecimal.ts
srknzl: This pattern is needed for exporting a constructor method that can be used with or without new keyword. `_Big` is the constructor method exported. The `BigInterface` interface is needed for type hints to be available on the exported constructor method. There may be a better method of doing this, but this worked for me.

We may rename them

srknzl: Also I can add some comments explaining this

src/core/DatetimeClasses.ts
srknzl: I explained above. The normal constructor cannot be used without `new`. I think static factory methods are a little bit harder to use given that some date/time classes accept each other, the code needed to build a HzOffsetDateTime will be too long.

src/core/DatetimeClasses.ts
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    /**
+     * Constructs a {@link HzLocalDate} object from string.
+     * @throws RangeError if string is not passed, or string format is wrong
+     * @param dateString String in the form of yyyy-mm-dd
+     */
+    static fromString(dateString: string): HzLocalDate {
+        if (typeof dateString !== 'string') {
+            throw new RangeError('String expected.');
```
mdumandag, line 200: Would https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError be better? Similar for others

src/core/DatetimeClasses.ts
srknzl: For year 0-99 maps to 1900-1999 in a Date constructor. So some values won't be able to be represented as `asDate` result.
image


I think it is not very hard to use the constructor since there are 3 or 4 fields

srknzl: Still I am not sure though

src/core/BigDecimal.ts
mdumandag: Why is instantiating the `BigDecimal` class without `new` keyword necessary?

src/core/DatetimeClasses.ts
mdumandag: I am not sure about that. So, with the constructor we have
```js
new HzLocalDateTime(new HzLocalDate(2000, 1, 1), new HzLocalTime(12, 12, 12))
```
with factory methods we have
```js
HzLocalDateTime.from(HzLocalDate.from(200, 1, 1), HzLocalTime.from(12, 12, 12))
```
They seem the same to me. Note that, we can even make the constructor private and just provide the static methods

srknzl: In the current(constructor method export) version it is like this:

```js
OffsetDateTime(LocalDateTime(LocalDate(2000, 1, 1), LocalTime(12, 12, 12)), 1000)
```

I removed `hz` prefix from constructors. The ones with Hz are class names. API docs include both constructor and class documentation.

 Class constructor is private with the current approach

But I realized that maybe we can export a static method of the class, instead of defining a new function. If it works without new, it is ok.

Note that one of the most popular BigDecimal library big.js uses a similar constructor approach that can be used with or without the `new` keyword. Another benefit of exporting a named constant is that you can rename the constructor to a shorter variable while keeping the class name long to make it more descriptive.

mdumandag: I still don't think the constructor without `new` is necessary. I mean, the argument that `big.js` does it like that is not enough IMHO. I could probably find dozens of popular libraries that do not provide such things as well.

BTW, I am OK with omitting Hz suffix from the classes and naming them to LocalTime, etc.

srknzl: I see no problem since my approach is wider, it can be also used with new. However, to be consistent with other classes we provide, I will remove the newless approach and user will use consturtors for construction. I will rename the classes to their names without HZ. I will add a from static method to localdatetime and offsetdatetime that takes all the input at once, instead of nested class arguments.

mdumandag: I could be Ok with something like `BigDecimal()` and `new BigDecimal()`, but I believe the name should be the same as the class

mdumandag: Maybe something like this, from TypeScript
```ts
interface Boolean {
    /** Returns the primitive value of the specified object. */
    valueOf(): boolean;
}

interface BooleanConstructor {
    new(value?: any): Boolean;
    <T>(value?: T): boolean;
    readonly prototype: Boolean;
}

declare var Boolean: BooleanConstructor;
```

srknzl: If you are ok with newless BigDecimal, then why not with datetime classes?

> I still don't think the constructor without new is necessary. I mean, the argument that big.js does it like that is not enough IMHO. I could probably find dozens of popular libraries that do not provide such things as well.

Also finding exported traditional classes in the popular libraries are very hard. I struggled to find one, I don't remember which one I found. Usually libraries(express, fastify, axios ) just export a function that is called and it is ready.

mdumandag: I might be OK with them also. I just want to have just one name for them to avoid any confusion

srknzl: I feel like I can try to follow big.js's approach it exports one name that can be used both for type and constructor. In this case we will probably need to stop defining class and use classic function approach.

see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/big.js/index.d.ts#L343

mdumandag: Is such thing a pattern to avoid in TypeScript codebases? I don't know much about conventions in TypeScript but I assume mixing the `class` and `functions` in the same codebase does not look good. That's just my opinion though, if it is OK, feel free to do it

srknzl: How about the new approach? An interface can be exported with same name of another variable.

commits:
https://github.com/hazelcast/hazelcast-nodejs-client/pull/963/commits/242eaefd055eb9ba790f5fde55383bceea2e75fc
https://github.com/hazelcast/hazelcast-nodejs-client/pull/963/commits/38799e33f607866b463e9f9cfcb6bf5666533f34
https://github.com/hazelcast/hazelcast-nodejs-client/pull/963/commits/fc4370fde3cf95602006a0c0068935e25e655020

A single name is exported, can be used with or without new. The intellisense works perfectly now, in order it to happen, for intellij ide I needed to move constructor method above interface. I think it picks the first one it finds when there are more than one exports with same name.

srknzl: What do you think? Should we remove this newless approach altogether since it makes the code complex? If it will stay we may consider renaming BigDecimal to Big.

src/core/BigDecimal.ts
srknzl: for convenience but I started think that we can remove it altogether

mdumandag: Well, for me adding `new` before creating an object is not a big burden but I don't know much about conventions for libraries. Personally, my vote goes for providing construction with new only but you may try to get some opinions about it from the team

src/core/DatetimeClasses.ts
mdumandag: Well, I am not sure about this approach because I don't know how that fits into the best practices of TypeScript. We now using the old way of doing things instead of the `class`es, right? Is this discouraged?

srknzl: I think not discouraged, but not very commonly used by libraries. Anyways, I will revert to the classic approach.

srknzl: See https://github.com/hazelcast/hazelcast-nodejs-client/commit/319f18ec1649fa95f3061de4cf8de10f780ee32b for reverted constructor change for BigDecimal and Date/Time

Added `static new()` to OffsetDatetime and LocalDateTime for easier construction. 

Also changed errors to TypeError as you suggested.

src/core/BigDecimal.ts
srknzl: I think this is fine since for loop is too long

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,326 @@
...
+    /** @internal */
+    private constructor(unscaledValue: BigInt, scale: number) {
+        this.unscaledValue = unscaledValue;
+        this.scale = scale;
+    }
+
+    /** @internal */
+    private static parseExp(input: string, offset: number, len: number): number {
+        let exp = 0;
```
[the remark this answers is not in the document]
srknzl: parseExp parses a exponent string like e+23, e-23, e23 and checks for invalid ones. TBH I do not trust js parseInt behaviour. To keep the consistency of jdk implementation inspiration, and since I tested this method 

For example 

`parseInt('1232312322222.2')` returns `1232312322222` but I want it to throw since it is not a valid integer.

test/integration/backward_compatible/sql/JetTest.js
```diff
@@ -0,0 +1,198 @@
...
+
+        await result.executeDeferred.promise; // wait for execution to end
+
+        const result2 = client.getSql().execute(`
+            CREATE MAPPING ${mapName2} (__key DOUBLE, name VARCHAR, height DOUBLE) TYPE IMap OPTIONS (
+              'keyFormat'='double',
+              'valueFormat'='json')
+        `);
+
+        await result2.executeDeferred.promise; // wait for execution to end
```
mdumandag, line 110: maybe instead of this private API, await something `result.updateCount()`?

src/serialization/DefaultSerializers.ts
```diff
@@ -428,3 +435,158 @@ export class UuidSerializer implements Serializer<UUID> {
...
+    id = -27;
+
+    read(input: DataInput): BigDecimal {
+        const body = input.readByteArray();
+        const scale = input.readInt();
+
+        return new BigDecimal(fromBufferAndScale(body, scale));
+    }
+
+    write(output: DataOutput, bigDecimal: BigDecimal): void {
```
mdumandag, line 557: Have you verified we can read the different kinds of values written by the Node.js client in Java? I mean, yes I see that we test some number in serialization tests but are things like negative values etc. tested?

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /**
+     * Parses exponent and returns it
+     *
+     * @param expString Exponent string
+     * @param offset Offset of where to start reading
+     * @param len How many chars to read
+     * @internal
+     * @private
```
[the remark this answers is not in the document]
srknzl: jetbrains auto generates this. not sure why

test/integration/backward_compatible/sql/DataTypeTest.js
```diff
@@ -78,6 +80,8 @@ describe('Decode/Serialize test', function () {
 
     before(async function () {
         TestUtil.markClientVersionAtLeast(this, '4.2');
+        const clientVersion = BuildInfo.calculateServerVersionFromString(BuildInfo.getClientVersion());
```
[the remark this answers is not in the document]
srknzl: Actually there was a function. I will try to use it instead

test/integration/backward_compatible/sql/JetTest.js
```diff
@@ -0,0 +1,198 @@
...
+    let map;
+    let map2;
+
+    const mapName = 'a';
+    const mapName2 = 'b';
+    const jetEnabledConfig = `<?xml version="1.0" encoding="UTF-8"?>
+        <hazelcast xmlns="http://www.hazelcast.com/schema/config"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://www.hazelcast.com/schema/config
+            http://www.hazelcast.com/schema/config/hazelcast-config-4.0.xsd">
```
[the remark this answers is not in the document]
srknzl: I don't see any warning. should jetbrains show it? btw 5.0 link does not exist

test/integration/backward_compatible/sql/JetTest.js
srknzl: Yeah it can be done and I am going to do that. this was my initial point of mentioning a new API that can be used to wait result execution. I think awaiting updateCount or other things are not very intuitive

src/serialization/DefaultSerializers.ts
srknzl: it should be in DefaultSerializersLiveTest.js I guess?

srknzl: I mean am I supposed to write it there, they are not written yet

srknzl: @mdumandag Added tests to `DefaultSerializersLiveTest.js`, and fixed your other comments(other than the ones that I am not sure what to do) you can rereview this pr.

Also do you think `static new()` constructors are ok, or should I rename it to `from()` ?

srknzl:
Not an important test failure, overlooked one, fixed it.

src/codec/builtin/FixSizedTypesCodec.ts
```diff
@@ -35,34 +40,35 @@ export class FixSizedTypesCodec {
         return buffer.readInt32LE(offset);
     }
 
-    static decodeLocalDate(buffer: Buffer, offset: number): string {
+    static decodeLocalDate(buffer: Buffer, offset: number): LocalDate {
         const year = FixSizedTypesCodec.decodeShort(buffer, offset);
```
mdumandag, line 44: Shouldn't this be readInt?

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,364 @@
...
+     * "12.3"         [123,1]
+     * "0.00123"      [123,5]
+     * "-1.23E-12"    [-123,14]
+     * "1234.5E-4"    [12345,5]
+     * "0E+7"         [0,-7]
+     * "-0"           [0,0]
+     * </pre>
+     *
+     * @param value a non-null `BigDecimal` string
+     * @throws RangeError if value is not a valid representation of a `BigDecimal`.
```
mdumandag, line 168: Maybe add TypeError as well to the documentation

src/core/BigDecimal.ts
```diff
@@ -0,0 +1,364 @@
...
+            throw new RangeError('Scale is out of range.');
+        }
+        scale = adjustedScale;
+        return scale;
+    }
+
+    /**
+     * Returns absolute value of a BigInt
+     * @param val
+     * @internal
```
mdumandag, line 277: Should we put `internal` to private functions? If so, `adjustScale` is missing one. If it is not necessary, I am OK with leaving them as it is or removing them

src/core/DatetimeClasses.ts
mdumandag: Anyways, I think we can go over this later and add it if necessary in the future

src/serialization/DefaultSerializers.ts
mdumandag: I think `from` might sound better and be less confusing (`new XX()` and `XX.new()` look alike), but I don't have a strong opinion here

src/serialization/SerializationService.ts
```diff
@@ -281,6 +290,7 @@ export class SerializationServiceV1 implements SerializationService {
         this.registerSerializer('arrayList', new ArrayListSerializer());
         this.registerSerializer('linkedList', new LinkedListSerializer());
         this.registerSerializer('uuid', new UuidSerializer());
+        this.registerSerializer('bigdecimal', new BigDecimalSerializer());
```
mdumandag, line 293: Minor:
```suggestion
        this.registerSerializer('bigDecimal', new BigDecimalSerializer());
```

src/util/BigDecimalUtil.ts
```diff
@@ -0,0 +1,82 @@
...
+    let bigint = BigInt(hexString);
+    if (isNegative) {
+        // When converting from 2 s complement, need to add 1 to the inverted bits.
+        // Since adding 1 to a buffer is hard, it is done here.
+        bigint += BigInt(1);
+    }
+
+    if (isNegative) {
+        bigint *= -BigInt(1);
+    }
```
mdumandag, line 42: Can we merge this into one check?

src/util/BigDecimalUtil.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    return bigint;
+}
+
+export function bigintToBuffer(bigintValue: BigInt): Buffer {
```
mdumandag, line 47:
```suggestion
export function bigIntToBuffer(bigIntValue: BigInt): Buffer {
```

src/util/BigDecimalUtil.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    // prepend 0 to get a even length string
+    if (hex.length % 2) {
+        hex = '0' + hex;
+    }
+
+    // we need to add the zero byte if the value is positive
+    // js BigInt toString(16) omits it
+    hex = '00' + hex;
```
mdumandag, line 68: In the comment, you say that we need to add `00` when the value is positive but we are adding this without a check. Is the comment wrong?

test/TestUtil.js
```diff
@@ -20,7 +20,8 @@ const { BuildInfo } = require('../lib/BuildInfo');
 
 exports.promiseLater = function (time, func) {
     if (func === undefined) {
-        func = () => {};
+        func = () => {
```
mdumandag, line 23: Minor: I think the old version was better. If the linter is not happy about `{}`, maybe put a space in between `{ }`

test/integration/backward_compatible/serialization/DefaultSerializersLiveTest.js
```diff
@@ -18,8 +18,10 @@
 const Long = require('long');
 const { expect } = require('chai');
 const RC = require('../../RC');
-const { Client, RestValue, UUID } = require('../../../../');
+const { Lang } = require('../../remote_controller/remote-controller_types');
+const { Client, RestValue, UUID, BigDecimal, LocalDate, LocalTime, LocalDateTime, OffsetDateTime } = require('../../../../');
 const TestUtil = require('../../../TestUtil');
+const { getTimezoneOffsetFromSeconds } = require('../../../../lib/util/DatetimeUtil');
```
mdumandag, line 24: Will these imports work on all clients?

test/integration/backward_compatible/sql/DataTypeTest.js
```diff
@@ -19,7 +19,7 @@
 const { Lang } = require('../../remote_controller/remote-controller_types');
 const RC = require('../../RC');
 const TestUtil = require('../../../TestUtil');
-const { Client } = require('../../../../');
+const { Client, LocalDateTime, LocalTime, LocalDate, OffsetDateTime, BigDecimal } = require('../../../../');
```
mdumandag, line 22: Same comment for imports

test/integration/backward_compatible/sql/JetTest.js
```diff
@@ -0,0 +1,198 @@
...
+        });
+    });
+
+    afterEach(async function () {
+        if (map) {
+            await map.destroy();
+        }
+        if (map2) {
+            await map2.destroy();
+        }
```
mdumandag, line 64: Minor: Since you destroy the cluster on each test, destroying maps are unnecessary. This way, you could also not define them above, and only define them in tests where you use maps

test/.eslintrc
```diff
@@ -18,5 +18,8 @@
         "mocha/prefer-arrow-callback": ["error", {
             "allowNamedFunctions": true
         }]
+    },
+    "parserOptions": {
+        "ecmaVersion": 11
```
mdumandag, line 23: Just wondering, why do we need this?

mdumandag (commented):
Looks good overall, left some more comments

src/codec/builtin/FixSizedTypesCodec.ts
srknzl: yeah somehow it is left like this, thanks for pointing out. Also to catch this, I will update DataTypeTest.js to use a year that is greater than the short range

src/core/BigDecimal.ts
srknzl: putting `internal` removes it from emitted type definition files, I will add to adjustScale as well. Thanks for the heads up

src/util/BigDecimalUtil.ts
srknzl: I renamed bigIntValue to big so no need to change that part. Changing method name

src/util/BigDecimalUtil.ts
srknzl: yeah probably left over good catch, we need to add a check there

test/TestUtil.js
srknzl: This are unfortunately auto changed by the IDE reverting them, I will check if there are others.

test/.eslintrc
srknzl: The root level eslintrc overrides ecmaVersion setting to be 8. This one in the test folder extends it so it has 8 too. However bigint literals like 1n requires ecmaVersion 11 or above. Normally in test eslint config, we set env to es2020 that automatically sets the ecmaVersion to 11, but since the root level config overrides it, test eslint config needs this additional override.

mdumandag (approved):

PR #967, opened 2021-07-06
API-631 Add portable methods for DECIMAL and DATE/TIME

srknzl:
merge #963  first

samples and docs will follow


src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     * Writes a Portable.
+     * Use {@link writeNullPortable} to write a `null` Portable
+     *
+     * @param fieldName name of the field
+     * @param portable  Portable to be written
+     */
+    writePortable(fieldName: string, portable: Portable | null): void;
 
+    /**
+     * To write a null portable value, user needs to provide class and factoryIds of related class.
```
mdumandag, line 161:
```suggestion
     * To write a `null` Portable value, you need to provide class and factory IDs of the related class.
```

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     * Write a time field consisting of hour, minute, seconds and nanos parts
+     *
+     * @param fieldName name of the field
+     * @param value     LocalTime value to be written
+     */
+    writeTime(fieldName: string, value: LocalTime | null): void
 
-    writeCharArray(fieldName: string, chars: string[]): void;
+    /**
+     * Writes a date field consisting of year, month of the year and day of the month
```
mdumandag, line 186:
```suggestion
     * Writes a date field consisting of year, month of the year, and day of the month
```

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     * Writes a decimal which is arbitrary precision and scale floating-point number
+     *
+     * @param fieldName name of the field
+     * @param value     BigDecimal value to be written
+     */
+    writeDecimal(fieldName: string, value: BigDecimal | null): void;
 
-    writeBooleanArray(fieldName: string, booleans: boolean[]): void;
+    /**
+     * Write a time field consisting of hour, minute, seconds and nanos parts
```
mdumandag, line 178:
```suggestion
     * Write a time field consisting of hour, minute, seconds, and nanoseconds parts
```

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     *
+     * @param fieldName name of the field
+     * @param value     LocalDate value to be written
+     */
+    writeDate(fieldName: string, value: LocalDate | null): void;
 
-    writeIntArray(fieldName: string, ints: number[]): void;
+    /**
+     * Writes a timestamp field consisting of
+     * year, month of the year, day of the month, hour, minute, seconds, nanos parts
```
mdumandag, line 195:
```suggestion
     * year, month of the year, day of the month, hour, minute, seconds, and nanoseconds parts
```

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     *
+     * @param fieldName name of the field
+     * @param value     LocalDateTime value to be written
+     */
+    writeTimestamp(fieldName: string, value: LocalDateTime | null): void;
 
-    writeLongArray(fieldName: string, longs: Long[]): void;
+    /**
+     * Writes a timestamp with timezone field consisting of
+     * year, month of the year, day of the month, offset seconds, hour, minute, seconds, nanos parts
```
mdumandag, line 204:
```suggestion
     * year, month of the year, day of the month, offset seconds, hour, minute, seconds, and nanoseconds parts
```

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     * @param value     OffsetDateTime value to be written
+     */
+    writeTimestampWithTimezone(fieldName: string, value: OffsetDateTime | null): void;
 
-    writeDoubleArray(fieldName: string, doubles: number[]): void;
+    /**
+     * Writes a primitive byte-array.
+     *
+     * @param fieldName name of the field
+     * @param bytes     byte array to be written
```
mdumandag, line 215: `byte array` or `byte-array`? I think it would be good to pick one style and apply it to everywhere. Similar comment for others. Personally, I would prefer `byte array` but I am OK with other also

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+    /**
+     * Writes a primitive short-array.
+     *
+     * @param fieldName name of the field
+     * @param values    short array to be written
+     */
+    writeShortArray(fieldName: string, values: number[] | null): void;
+
+    /**
+     * Writes a String-array.
```
mdumandag, line 276:
```suggestion
     * Writes a string-array.
```

mdumandag: Same for other such methods and documentations

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     *
+     * @param fieldName name of the field
+     * @param values    String array to be written
+     */
+    writeStringArray(fieldName: string, values: string[] | null): void;
+
+    /**
+     * Writes a an array of Portables.
```
mdumandag, line 294: We can also be consistent here. `array of Portables` vs `Portable array`?

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     *
+     * @param fieldName name of the field
+     * @param values    portable array to be written
```
mdumandag, line 297:
```suggestion
     * @param values    Portable array to be written
```

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     */
+    writePortableArray(fieldName: string, values: Portable[] | null): void;
+
+    /**
+     * Writes an array of Decimals
```
mdumandag, line 302:
```suggestion
     * Writes an array of BigDecimals
```

I would use the name of the class here

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     *
+     * @param fieldName name of the field
+     * @param values    BigDecimal array to be written
+     * @see {@link writeDecimal}
+     */
+    writeDecimalArray(fieldName: string, values: BigDecimal[] | null): void;
+
+    /**
+     * Writes an array of Time's
```
mdumandag, line 311:
```suggestion
     * Writes an array of LocalTime's
```

Same for others. I think we should use the full class name of the thing we write

src/serialization/portable/DefaultPortableReader.ts
```diff
@@ -126,59 +133,109 @@ export class DefaultPortableReader implements PortableReader {
...
+        if (DefaultPortableReader.isNullOrEmpty(pos)) {
+            return null;
+        }
         return this.input.readByteArray(pos);
     }
 
-    readBooleanArray(fieldName: string): boolean[] {
+    readBooleanArray(fieldName: string): boolean[] | null {
         const pos = this.positionByField(fieldName, FieldType.BOOLEAN_ARRAY);
+        if (DefaultPortableReader.isNullOrEmpty(pos)) {
```
mdumandag, line 166: Can you write a test code that produces `-1` as pos? I couldn't find a way how that might happen. It either throws or returns a valid position

src/serialization/portable/DefaultPortableWriter.ts
```diff
@@ -115,56 +125,85 @@ export class DefaultPortableWriter implements PortableWriter {
         this.output.writeInt(classId);
     }
 
-    writeByteArray(fieldName: string, bytes: Buffer): void {
+    writeDecimal(fieldName: string, value: BigDecimal): void {
```
mdumandag, line 128: Should this be `: BigDecimal | null`? similar for others

src/util/IOUtil.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    static writeOffsetDatetime(out: DataOutput, value: OffsetDateTime): void {
+        IOUtil.writeLocalDatetime(out, value.localDateTime);
+        out.writeInt(value.offsetSeconds);
+    }
+
+}
+
+
```
mdumandag, line 103: Unnecessary newline

src/util/IOUtil.ts
```diff
@@ -0,0 +1,103 @@
...
+    static readLocalDatetime(inp: DataInput): LocalDateTime {
+        const localDate = IOUtil.readLocalDate(inp);
+        const localTime = IOUtil.readLocalTime(inp);
+
+        return new LocalDateTime(localDate, localTime);
+    }
+
+    static readOffsetDatetime(inp: DataInput): OffsetDateTime {
+        const localDate = IOUtil.readLocalDate(inp);
+        const localTime = IOUtil.readLocalTime(inp);
```
mdumandag, line 60:
```suggestion
        const localDateTime = IOUtil.readLocalDateTime();
```

src/util/IOUtil.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    static readLocalDate(inp: DataInput): LocalDate {
+        const year = inp.readInt();
+        const month = inp.readByte();
+        const date = inp.readByte();
+        return new LocalDate(year, month, date);
+    }
+
+    static readLocalDatetime(inp: DataInput): LocalDateTime {
```
mdumandag, line 51:
```suggestion
    static readLocalDateTime(inp: DataInput): LocalDateTime {
```

src/util/PortableUtil.ts
```diff
@@ -0,0 +1,75 @@
...
+    LocalDate,
+    LocalDateTime,
+    OffsetDateTime,
+} from '../core';
+import {IOUtil} from './IOUtil';
+
+/** @internal */
+export class PortableUtil {
+
+    static readLocalDateForPortable(inp: DataInput): LocalDate {
```
mdumandag, line 29: I think we might omit the `ForPortable` name, as the class name suggests that it is for Portable

src/util/PortableUtil.ts
```diff
@@ -0,0 +1,75 @@
...
+        const year = inp.readShort(); // this is different due to backward compatibility
+        const month = inp.readByte();
+        const date = inp.readByte();
+        return new LocalDate(year, month, date);
+    }
+
+    static readLocalDatetimeForPortable(inp: DataInput): LocalDateTime {
```
mdumandag, line 36:
```suggestion
    static readLocalDateTimeForPortable(inp: DataInput): LocalDateTime {
```

src/util/PortableUtil.ts
```diff
@@ -0,0 +1,75 @@
...
+
+        return new OffsetDateTime(
+            new LocalDateTime(
+                localDate,
+                localTime
+            ),
+            offsetSeconds
+        );
+    }
+
```
mdumandag, line 57: Unnecessary newline

test/unit/serialization/PortableObjects.js
```diff
@@ -281,10 +355,32 @@ class Child {
 
 }
 
+class SameFieldNamePortable {
```
mdumandag, line 358: Maybe `DuplicateFieldNamePortable`?

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     */
+    writeLong(fieldName: string, value: Long): void;
 
-    /** @deprecated since version 4.2.
-     *  This method will be removed in next major version. Please use {@link writeString} instead.
+    /**
+     * Writes an UTF string.
+     *
+     * @param fieldName name of the field
+     * @param value     utf string value to be written
```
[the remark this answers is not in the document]
srknzl: I think your suggestion is better, will do

src/util/IOUtil.ts
srknzl: FYI: I did a mass renaming following this

src/serialization/Portable.ts
```diff
@@ -41,127 +48,552 @@ export enum FieldType {
...
+     * @deprecated since version 4.2 for the sake of better naming. Please use {@link writeString} instead.
+     * This method will be removed in the next major version.
```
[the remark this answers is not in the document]
srknzl: Let's discuss this tomorrow standup

src/serialization/portable/DefaultPortableReader.ts
srknzl: Yeah I think you are right, this checks are added quite a while ago in https://github.com/hazelcast/hazelcast/commit/c7b0adc65aa28ae55e43c71a730b62e38c8bf37a. I think we can remove them.

Can you check isNullOrEmpty usages in Java side and verify that it is useless. We should remove it from Java side too.

src/serialization/Portable.ts
```diff
@@ -41,127 +48,607 @@ export enum FieldType {
...
  * Writer helper for {@link Portable} objects.
  */
 export interface PortableWriter {
+
+    /**
+     * Writes a number as 32-bit signed integer.
+     *
+     * @param fieldName name of the field
+     * @param value     int value to be written. The value must be a valid signed 32-bit integer.
+     * Behavior is undefined when value is anything other than a signed 32-bit integer.
```
mdumandag, line 74:
```suggestion
     * Behavior is undefined when value is out of the signed 32-bit integer range.
```
Maybe this one is a bit more clear. WDYT? If you agree with that, we should also update the other ones.

src/serialization/Portable.ts
```diff
@@ -41,127 +48,607 @@ export enum FieldType {
...
+     */
+    writeChar(fieldName: string, value: string): void;
 
-    writeDouble(fieldName: string, double: number): void;
+    /**
+     * Writes a number as double.
+     *
+     * @param fieldName name of the field
+     * @param value     double value to be written. The value must be a JavaScript number.
+     * Behavior is undefined when value is anything other than a JavaScript number.
```
mdumandag, line 136: How can it be anything other than a number? Also, as far as I remember, JS number and Java double have the same range so this warning might be unnecessary

src/serialization/Portable.ts
```diff
@@ -41,127 +48,607 @@ export enum FieldType {
...
+     */
+    writeDouble(fieldName: string, value: number): void;
 
-    writeFloat(fieldName: string, float: number): void;
+    /**
+     * Writes a number as float.
+     *
+     * @param fieldName name of the field
+     * @param value     float value to be written. The value must be a JavaScript number.
+     * Behavior is undefined when value is anything other than a JavaScript number.
```
mdumandag, line 145: Do you mean that the behavior is undefined when the number is bigger than the 32-bit single-precision floating-point number range? I think we can be more clear here

src/serialization/Portable.ts
```diff
@@ -41,127 +48,607 @@ export enum FieldType {
...
+     * @param fieldName name of the field
+     * @param value     boolean value to be written
+     */
     writeBoolean(fieldName: string, value: boolean): void;
 
+    /**
+     * Writes a number as 8-bit unsigned integer.
+     *
+     * @param fieldName name of the field
+     * @param value     Byte value to be written. Must be a valid unsigned 8-bit integer.
```
mdumandag, line 115:
```suggestion
     * @param value     byte value to be written. Must be a valid unsigned 8-bit integer.
```

src/serialization/Portable.ts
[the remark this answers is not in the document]
srknzl: I asked Sancar, they removed `This method will be removed in the next major version.` part from Java, and they did not decide to remove it. So I am updating this documentation

src/serialization/Portable.ts
```diff
@@ -41,127 +48,607 @@ export enum FieldType {
...
+     * Reads a timestamp.
+     *
+     * @param fieldName name of the field
+     * @return the LocalDateTime value read
+     */
+    readTimestamp(fieldName: string): LocalDateTime | null;
 
-    readLongArray(fieldName: string): Long[];
+    /**
+     * Reads a timestampWithTimezone.
```
mdumandag, line 507:
```suggestion
     * Reads a timestamp with timezone.
```
to be consistent with the documentation of the write method

src/serialization/portable/DefaultPortableReader.ts
mdumandag: To me, Java side checks also seem unnecessary.

mdumandag:
It seems that we also have a conflict in the `test/unit/core/DatetimeClasses.js`. Can you resolve it?

src/serialization/Portable.ts
srknzl: Actually, these are directly taken from [Node.js docs.](https://nodejs.org/docs/latest-v14.x/api/buffer.html#buffer_buf_writeint32be_value_offset). I think both versions are fine. The current version has a wider meaning, i.e when value is not a number behaviour is undefined. I think it can stay like this because I have no strong choice.

src/serialization/Portable.ts
srknzl: Value has number type but in runtime anything can be passed to it. Especially in environments where typescript is not used.

src/serialization/Portable.ts
srknzl: Yeah actually this seemed interesting to me too. Again, I took this directyl from nodejs docs https://nodejs.org/docs/latest-v14.x/api/buffer.html#buffer_buf_writefloatbe_value_offset

I think the intent of saying this is that when a value other than a number is passed, behaviour is undefined. Nothing related to the number range.

mdumandag (approved):

PR #973, opened 2021-07-13
API-579 Add code samples for decimal and date/time classes usage

srknzl:
follow up to #967 merge it first


srknzl:
verify

src/proxy/IQueue.ts
```diff
@@ -154,8 +154,8 @@ export interface IQueue<E> extends DistributedObject {
     size(): Promise<number>;
 
     /**
-     * Retrieves and removes the head of this queue. It waits to return until
-     * an element becomes available if neccessary.
+     * Retrieves and removes the head of this queue, waiting if necessary
+     * until an element becomes available.
```
srknzl, line 158: this was outdated

code_samples/bigdecimal_sample.js
```diff
@@ -0,0 +1,94 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+// This sample code demonstrates big decimal usage.
+
+const { Client, Big } = require('..');
```
mdumandag, line 20: I think the code samples are not up to date, as we shouldn't have something called `Big`

srknzl: Yeah you are right, I will change this

code_samples/bigdecimal_sample.js
```diff
@@ -0,0 +1,97 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+// This sample code demonstrates BigDecimal usage.
+
+const { Client, BigDecimal } = require('..');
```
mdumandag, line 20:
```suggestion
const { Client, BigDecimal } = require('hazelcast-client');
```
so that the users would have an easier time copy-pasting and running the code in their locals

code_samples/bigdecimal_sample.js
```diff
@@ -0,0 +1,97 @@
...
+
+class CustomNumber {
```
mdumandag, line 22: Where do we use this? I saw that we register the factory on the client config but never used this

code_samples/datetime_classes_sample.js
```diff
@@ -0,0 +1,104 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+// This sample code demonstrates datetime classes usage.
+
+const { Client, LocalDate, OffsetDateTime } = require('..');
```
mdumandag, line 20:
```suggestion
const { Client, LocalDate, OffsetDateTime } = require(hazelcast-client');
```

code_samples/datetime_classes_sample.js
```diff
@@ -0,0 +1,104 @@
...
+
+class CustomNumber {
```
mdumandag, line 22: the name CustomNumber is not in line with the fields we have, right? Maybe we can rename it

mdumandag: Same comment above, we don't use this. If you intend to just show the usage, without using it, I believe we need to document this with a comment. Also, in this case, I don't think we need the factory definition or config registration.

But I would like to see a simple usage, put it to some data structure (like set etc.)

code_samples/.eslintrc
```diff
@@ -0,0 +1,11 @@
...
+    "env": {
+        "es2020": true
+    },
+    "extends": [
+        "../.eslintrc"
+    ],
+    "parserOptions": {
+        "ecmaVersion": 11
+    }
+}
```
srknzl, line 11: This is needed for bigint usage in code samples directory (big decimal sample)

mdumandag (approved):

PR #978, opened 2021-07-14
API-657 GH actions

srknzl:
adds a pr builder and branch builder. we need to add HAZELCAST_ENTERPRISE_KEY secret to the repo


srknzl:
verify

.github/workflows/test_branch.yml
```diff
@@ -0,0 +1,47 @@
+name: Test Branch
```
mdumandag, line 1: Can we merge these two files into one? They are almost exact copies of each other.
I think we can add another `on` rule below for PRs and pass `HAZELCAST_ENTERPRISE_KEY` conditionally as an environment variable. WDTY?

.github/workflows/test_branch.yml
```diff
@@ -0,0 +1,47 @@
...
+        steps:
+            - name: Setup Java
+              uses: actions/setup-java@v1
+              with:
+                  java-version: 8
+            - name: Setup Node.js
+              uses: actions/setup-node@v2
+              with:
+                  node-version: 10
+            - name: Disable line ending conversion of git
```
mdumandag, line 27: Why is this necessary?

srknzl: On windows, git converts line endings from linux format to windows format. Eslint was printing a lot of warnings since we have a rule for line endings.

.github/workflows/test_branch.yml
srknzl: Makes sense, let me try

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
+name: Test Branch
```
mdumandag, line 1:
```suggestion
name: Run tests
```
WDYT? Test branch seems a little bit off to me, after the recent changes

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+on:
+    push:
+        branches-ignore:
+            - 'gh-pages'
+        tags-ignore:
+            - '*'
+    pull_request:
+
+jobs:
+    test-branch:
```
mdumandag, line 11:
```suggestion
    run-tests:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+        name: Test Branch (${{ matrix.os }})
```
mdumandag, line 12:
```suggestion
        name: Run tests on (${{ matrix.os }})
```

mdumandag (approved):
Looks good overall. I left some naming-related comments.

BTW, we should work on shortening the test duration as it is close to 30 min mark now. It is almost as long as the IMDG itself. We can maybe make use of parallel test runs or work on the long tests to make them simpler/shorter

srknzl:
I checked mocha parallel test feature but I remember it was hard to enable because running tests in parallel affect other tests. I will try to do again. Opening a issue.

srknzl:
After this is merged, should we ditch Jenkins @mdumandag ?

PR #989, opened 2021-07-29
API-664 Add code sample checker

### LINKED ISSUE #962: Consider adding a code sample runner script for ci
Needed to check if code samples are runnable

srknzl:
fixes #962 

* Adds a gh actions workflow that checks validity of code samples by running them

* Adds process.exit(1); to code samples to ensure they exit in case of an error, otherwise code sample checker will hang

merge #973  first


.github/workflows/code_sample_checker.yml
```diff
@@ -0,0 +1,34 @@
...
+on: pull_request
+
+jobs:
+    test-code-samples:
+        name: check-code-samples
+        runs-on: ${{matrix.os}}
+
+        strategy:
+            matrix:
+                os: [ ubuntu-latest, windows-latest ]
```
mdumandag, line 11: Do we really need to test this on Windows too? I mean, for a code sample, it shouldn't really matter

.github/workflows/code_sample_checker.yml
```diff
@@ -0,0 +1,34 @@
+name: Code sample checker
+on: pull_request
+
+jobs:
+    test-code-samples:
+        name: check-code-samples
```
mdumandag, line 6:
```suggestion
        name: Check code samples
```
Let's make it more human-readable :smile:

scripts/code-sample-checker.js
```diff
@@ -0,0 +1,67 @@
...
+    'hazelcast_cloud_discovery.js', // needs a token
+    'jaas_sample' // this example is hard to run
+];
+
+// Recursively walks inside a directory, returning all the JavaScript files inside of it
+async function walkJsFiles(dir) {
+    let files = await fs.readdir(dir);
+    files = await Promise.all(files.map(async file => {
+        const filePath = path.join(dir, file);
+        const stats = await fs.stat(filePath);
```
mdumandag, line 20: Minor: we can move this to below, so that we don't have to await for it, if the file is in the ignore list

scripts/code-sample-checker.js
```diff
@@ -0,0 +1,67 @@
...
+        console.log(`Running ${file}, ${counter}/${numberOfFiles}`);
+        counter++;
+
+        const subprocess = spawnSync('node', [file], {
+            stdio: ['ignore', 'inherit', 'pipe'] // redirect stderr and stdout to parent process, ignore stdin
+        });
+
+        await RC.terminateMember(cluster.id, member.uuid);
+
+        const stderrString = subprocess.stderr.toString().trim();
```
mdumandag, line 62: Can we check the error code of the subprocess, instead of the stderr? I think the former would be more reliable way

scripts/test-runner.js
```diff
@@ -201,15 +217,26 @@ process.on('SIGINT', shutdownProcesses);
...
-        console.log(`Running tests... Test type: ${testType}, Test command: ${testCommand}`);
+        console.log(`Running ${testType}, Command: ${testCommand}`);
         testProcess = spawn(testCommand, [], {
             stdio: ['ignore', 'inherit', 'inherit'],
             shell: true
         });
         testProcess.on('exit', shutdownRC);
+    } else if (testType === 'check-code-samples') {
+        const RC = getRC();
+        cluster = await RC.createClusterKeepClusterName(null, TestUtil.createClusterConfig({clusterName: 'dev'}));
```
mdumandag, line 231: I think you should be able to pass `null` for the cluster config. After all, the default value for the clusterName is `dev`, so the default config should also be enough for us.

Then you can un-export the createClusterConfig.

.github/workflows/code_sample_checker.yml
srknzl: Does it hurt? I think it may be useful to check that code samples work on windows which is important.

scripts/test-runner.js
srknzl: default value is not dev, it is random uuid. check out https://github.com/hazelcast/hazelcast-remote-controller/blob/master/src/main/java/com/hazelcast/remotecontroller/HzCluster.java

mdumandag: Then I would just pass the config as a string. I think using the test utilities here is not a good practice.
```
<?xml version="1.0" encoding="UTF-8"?>
<hazelcast xmlns="http://www.hazelcast.com/schema/config"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://www.hazelcast.com/schema/config
    http://www.hazelcast.com/schema/config/hazelcast-config-4.0.xsd">
    <cluster-name>dev</cluster-name>
</hazelcast>
```

.github/workflows/code_sample_checker.yml
mdumandag: It hurts me (I care for the poor Github) but I can live with that :smile:

srknzl: Don't worry this one does not take long :)

mdumandag (approved):

PR #1000, opened 2021-08-02
Add paging predicate sample

### LINKED ISSUE #903: Consider adding a link to this example in documentation
Example: https://github.com/srknzl/hazelcast-paging-predicate-example

Background:
https://hazelcastcommunity.slack.com/archives/C01PP55A667/p1621233287000900

srknzl:
fixes #903


DOCUMENTATION.md
```diff
@@ -2733,7 +2733,11 @@ You can configure this using `metadata-policy` element for the map configuration
 
 #### 8.8.1.5. Filtering with Paging Predicates
 
+> **WARNING: Starting from version 4.2, [the SQL querying feature](#87-sql) is introduced.
+> The SQL querying feature is more efficient than paging predicates, and it is recommended to be used instead of paging predicates.**
```
mdumandag, line 2737:
```suggestion
> The SQL querying feature is more efficient than paging predicates, and it is recommended to be used instead.**
```
Maybe we could make it shorter like this

DOCUMENTATION.md
```diff
@@ -2733,7 +2733,11 @@ You can configure this using `metadata-policy` element for the map configuration
...
+
 The Node.js client provides paging for defined predicates. With its `PagingPredicate` object, you can get a list of keys, values or entries page by page by filtering them with predicates and giving the size of the pages. Also, you can sort the entries by specifying comparators.
+You can find an example code [here.](https://github.com/srknzl/hazelcast-paging-predicate-example)
```
mdumandag, line 2740: I am not okay with having some links pointing to the code sample at a place that is not in control of Hazelcast, for Hazelcast features. I mean, there is no guarantee that you won't delete it in the future :smile: So, I would leave the documentation as it is.

BTW, why do you used some custom comparator for integer values? Can't it be more suitable if you show some sort of complex object, which cannot be used out-of-the-box with paging predicates?

srknzl: we can move the repository to hazelcast organization. Should I use a complex object to showcase how it is used? Why it is more suitable? Also, see the new reported issue from Andrey #1009, is the sort method redundant in nodejs client. If so I can remove it in this pr.

mdumandag: Yeah, that method is redundant and I feel like can be removed in 5.0. But, I feel like we can do it another PR.

I feel like if we are going such long measures, let's build something more complex BUT I don't feel like this feature (paging predicate) deserves such thing. Was there a user request for it, for the extended documentation?

srknzl: Yes someone asked how he can use the paging predicate with sorting. And I prepared a the repo for him. That's why I opened this PR. 

Then, I told him that he can use SQL query which is easier and he was happy to hear it. However, I think for legacy users we can include such a guide.

mdumandag (approved):

DOCUMENTATION.md
mdumandag: Then, maybe I would provide this repo you prepared as a code sample in this repo, similar to the JAAS sample, with its own README etc. explaining the necessary steps to run it, BUT that is definitely not a must.

Like I said, I don't think we need to promote this feature more as it is highly inefficient

srknzl: I have already added a warning about its unefficiency against sql. I think it fine to add a sample for legacy users.

code_samples/paging_predicate_sample/Comparator.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void writeData(ObjectDataOutput out) throws IOException {
+    }
+
+    @Override
+    public void readData(ObjectDataInput in) throws IOException {
+    }
+
+    // We needed a Double comparator since the default number type is 'double' in Node.js client. You can change it via defaultNumberType serialization setting.
```
mdumandag, line 22:
```suggestion
    // We needed a Double comparator since the default number type is 'double' in Node.js client. 
    // You can change it via defaultNumberType serialization setting.
```
To make it more readable

code_samples/paging_predicate_sample/hazelcast.xml
```diff
@@ -0,0 +1,345 @@
...
+  or the Reference Manual at https://hazelcast.org/documentation/
+-->
+
+<!--suppress XmlDefaultAttributeValue -->
+<hazelcast xmlns="http://www.hazelcast.com/schema/config"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://www.hazelcast.com/schema/config
+           http://www.hazelcast.com/schema/config/hazelcast-config-4.2.xsd">
+
+    <cluster-name>dev</cluster-name>
```
mdumandag, line 34: Can you just leave what you need in this file (factory registration) and remove everything else?

I see that it is just using the default values and other settings are not related to this PR

code_samples/paging_predicate_sample/main.js
```diff
@@ -0,0 +1,69 @@
...
+const { Client, Predicates } = require('hazelcast-client');
+
+class Comparator {
+    constructor() {
+        this.factoryId = 66;
+        this.classId = 103;
+    }
+
+    sort(a, b) {
+        if (a[1] > b[1]) {
```
mdumandag, line 11: Can you remove the method body and add a comment saying that the actual sorting happens on the member side and nothing is required here?

code_samples/paging_predicate_sample/main/src/main/java/Comparator.java
```diff
@@ -0,0 +1,47 @@
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+
+import java.io.IOException;
+import java.util.Map;
+
+public class Comparator implements java.util.Comparator<Map.Entry<String, Double>>, IdentifiedDataSerializable {
```
mdumandag, line 8: We have the same files above right? What is the point?

code_samples/paging_predicate_sample/README.md
```diff
@@ -0,0 +1,10 @@
+# hazelcast-paging-predicate-example
+
+1. Download hazelcast. Extract it somewhere.  https://hazelcast.org/imdg/download/
```
mdumandag, line 3: This needs a general fix.

First, we need to change the wording here so it doesn't contain `I` etc. and make it a bit more formal.

I also would like to see what is this code sample for, and why do we need to do such things etc.

code_samples/paging_predicate_sample/main/src/main/java/Comparator.java
srknzl: yeah we can remove them

.gitignore
```diff
@@ -1,5 +1,5 @@
 target/*
-.idea/*
```
srknzl, line 2: needed to avoid .idea folders inside code samples directory ( when opening a code sample in intellij it is created)

code_samples/paging_predicate_sample/README.md
```diff
@@ -0,0 +1,23 @@
...
+The following describes how to do it:
+
+You can use the sample project in [main](./main) folder:
+
+1. Run the [Main](./main/src/main/java/Main.java) class. The started member will already have the `Comparator` class registered.
+2. Run the sample code in main.js via `node main.js`.
+
+Alternatively, you can start from scratch:
+
+1. Download hazelcast. Extract it somewhere. https://hazelcast.org/imdg/download/
```
mdumandag, line 14:
```suggestion
1. Download Hazelcast. Extract it somewhere. https://hazelcast.org/imdg/download/
```
Can you do the same for other instances?

Also, the link needs to be updated after .com site migration

code_samples/paging_predicate_sample/README.md
```diff
@@ -0,0 +1,23 @@
...
+2. To add a custom comparator you need to compile and add it to hazelcast classpath.
+3. In order to compile use a java project that includes hazelcast as a dependency, you can use an IDE like Intellij for simplicity. Just open the IDE and press build button, then the classes will be in `target` folder. The java project used is in the `main` folder.
```
mdumandag, line 16:
```suggestion
3. In order to compile use a Java project that includes hazelcast as a dependency, you can use an IDE like Intellij for simplicity. Just open the IDE and press build button, then the classes will be in `target` folder. The java project used is in the `main` folder.
```
Can you do the same for other instances?

Also, `Intellij` is not the name of the IDE, it is `IDEA` or `Intellij IDEA`

code_samples/paging_predicate_sample/hazelcast.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<hazelcast xmlns="http://www.hazelcast.com/schema/config"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://www.hazelcast.com/schema/config
+           http://www.hazelcast.com/schema/config/hazelcast-config-4.2.xsd">
+
+    <serialization>
+        <portable-version>0</portable-version>
```
mdumandag, line 8: This is not necessary

code_samples/paging_predicate_sample/main/src/main/java/Comparator.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public void writeData(ObjectDataOutput out) throws IOException {
+    }
+
+    @Override
+    public void readData(ObjectDataInput in) throws IOException {
+    }
+
```
mdumandag, line 21: Unnecessary newline

mdumandag (approved):
Approving with minor comments

srknzl:
verify

PR #1001, opened 2021-08-02
API-668 Document and Validate Properties

### LINKED ISSUE #894: Improve documentation regarding properties
It would be better to have a separate section about client system properties like the Java client does.
https://docs.hazelcast.com/imdg/4.2/clients/java.html#client-system-properties

We will check which properties the Node.js client has and then copy the their documentation from Java docs. Also default values might be different, need to check that as well.

srknzl:
also fixes test runner about exit status codes for gh actions to work properly

adds padded-blocks: 'never' for consistency of block newlines

fixes #894


DOCUMENTATION.md
```diff
@@ -564,6 +565,31 @@ const client = await Client.newHazelcastClient(cfg);
 
 In the following chapters you will learn the description of all options supported by Hazelcast Node.js client.
 
+# 3.1 Client Properties
+
+The following is the list of all client properties in alphabetical order.
+
+> **NOTE: When you want to reconfigure a client property, you need to restart the client.**
```
mdumandag, line 572: Isn't this the same for all of the config elements? We don't support dynamic updates to config elements (or let's say the behavior, in this case, is undefined), so I would remove this warning

DOCUMENTATION.md
```diff
@@ -564,6 +565,31 @@ const client = await Client.newHazelcastClient(cfg);
...
+| hazelcast.client.autopipelining.threshold.bytes        | 65536 (64 KB)                       | number          | This is the coalescing threshold for the internal queue used by automated pipelining. Once the total size of operation payloads taken from the queue reaches this value during batch preparation, these operations are written to the socket. Notice that automated pipelining will still send operations if their total size is smaller than the threshold and there are no more operations in the internal queue. |
+| hazelcast.client.cloud.url                             | https://coordinator.hazelcast.cloud | string          | Hazelcast Cloud URL.                                                                                                                                                                                                                                                                                                                                                                                                | <!-- markdown-link-check-disable-line -->
```
mdumandag, line 578: This shouldn't be public. It is only suitable for the cloud team to use, to test the client in different environments. For users, they don't need to bother with this and only use the production environment, which is the default value.

DOCUMENTATION.md
```diff
@@ -564,6 +565,31 @@ const client = await Client.newHazelcastClient(cfg);
...
+| hazelcast.client.heartbeat.interval                    | 5000                                | number          | Frequency of the heartbeat messages sent by the clients to members.                                                                                                                                                                                                                                                                                                                                                 |
```
mdumandag, line 579:
```suggestion
| hazelcast.client.heartbeat.interval                    | 5000                                | number          | Frequency of the heartbeat messages sent by the clients to members in milliseconds.                                                                                                                                                                                                                                                                                                                                                 |
```

DOCUMENTATION.md
```diff
@@ -564,6 +565,31 @@ const client = await Client.newHazelcastClient(cfg);
...
+| hazelcast.client.heartbeat.timeout                     | 60000                               | number          | Timeout for the heartbeat messages sent by the client to members. If no messages pass between the client and member within the given time via this property in milliseconds, the connection will be closed.                                                                                                                                                                                                         |
```
mdumandag, line 580:
```suggestion
| hazelcast.client.heartbeat.timeout                     | 60000                               | number          | Timeout for the heartbeat messages sent by the client to members in milliseconds. If no messages pass between the client and member within the given time via this property, the connection will be closed.                                                                                                                                                                                                         |
```

DOCUMENTATION.md
```diff
@@ -564,6 +565,31 @@ const client = await Client.newHazelcastClient(cfg);
...
+| hazelcast.client.invocation.retry.pause.millis         | 1000                                | number          | Pause time between each retry cycle of an invocation in milliseconds.                                                                                                                                                                                                                                                                                                                                               |
+| hazelcast.client.invocation.timeout.millis             | 120000                              | number          | Period, in milliseconds, to give up the invocation when a member in the member list is not reachable.                                                                                                                                                                                                                                                                                                               |
```
mdumandag, line 582: There are also other scenarios where we use this property. From the reference manual
```
You can set a timeout for retrying the operations sent to a member. This can be provided by using the property hazelcast.client.invocation.timeout.seconds in ClientProperties. The client retries an operation within this given period, of course, if it is a read-only operation or you enabled the redoOperation as stated in the above paragraph. This timeout value is important when there is a failure resulted by either of the following causes:

Member throws an exception.

Connection between the client and member is closed.

Client’s heartbeat requests are timed out.
```

DOCUMENTATION.md
```diff
@@ -564,6 +565,31 @@ const client = await Client.newHazelcastClient(cfg);
...
+| hazelcast.client.operation.fail.on.indeterminate.state | false                               | boolean         | When it is `true`, if an operation has sync backups and acks are not received from backup replicas in time, or the member which owns primary replica of the target partition leaves the cluster, then the invocation fails. However, even if the invocation fails, there will not be any rollback on other successful replicas.                                                                                     |
+| hazelcast.client.shuffle.member.list                   | true                                | boolean         | The client shuffles the given member list to prevent all the clients to connect to the same member when this property is true. When it is set to false, the client tries to connect to the members in the given order.                                                                                                                                                                                              |
```
mdumandag, line 585:
```suggestion
| hazelcast.client.shuffle.member.list                   | true                                | boolean         | The client shuffles the given member list to prevent all the clients to connect to the same member when this property is `true`. When it is set to `false`, the client tries to connect to the members in the given order.                                                                                                                                                                                              |
```

DOCUMENTATION.md
```diff
@@ -564,6 +565,31 @@ const client = await Client.newHazelcastClient(cfg);
...
+| hazelcast.client.statistics.period.seconds             | 3                                   | number          | Period in seconds the client statistics are collected and sent to the cluster.                                                                                                                                                                                                                                                                                                                                      |
+| hazelcast.discovery.public.ip.enabled                  | null(detection enabled)             | boolean or null | When set to `true`, the client will assume that it needs to use public IP addresses reported by members. When set to `false`, the client will always use private addresses reported by members.                                                                                                                                                                                                                     |
```
mdumandag, line 589: I think we should document that, if it is null, we try to infer how the cluster is structured and act accordingly. I also think that, we should say that this inference is not %100 reliable and may result in false negatives (acting like `false`, while the cluster is set up to act like `true`)

DOCUMENTATION.md
```diff
@@ -564,6 +565,31 @@ const client = await Client.newHazelcastClient(cfg);
...
+| hazelcast.invalidation.max.tolerated.miss.count        | 10                                  | number          | If missed invalidation count is bigger than this value, relevant cached data will be made unreachable.                                                                                                                                                                                                                                                                                                              |
```
mdumandag, line 590:
```suggestion
| hazelcast.invalidation.max.tolerated.miss.count        | 10                                  | number          | If missed invalidation count is greater than this value, relevant cached data will be made unreachable.                                                                                                                                                                                                                                                                                                              |
```
Maybe we can put the word `Near cache` somewhere, as this is related to that.

src/config/ConfigBuilder.ts
```diff
@@ -223,10 +224,76 @@ export class ConfigBuilder {
...
+                        value = tryGetNumber(value);
+                        break;
+                    case 'hazelcast.invalidation.max.tolerated.miss.count':
+                        value = tryGetNumber(value);
+                        break;
+                    case 'hazelcast.invalidation.min.reconciliation.interval.seconds':
+                        value = tryGetNumber(value);
+                        break;
+                    case 'hazelcast.logging.level':
+                        tryGetEnum(LogLevel, value); // only validate
```
mdumandag, line 264: can't we change the implementation so that we use the value? I see that in the LoggingService, we already do that. Am I missing something?

DOCUMENTATION.md
srknzl: Yeah this is taken from properties section of IMDG, but let's remove as you said

DOCUMENTATION.md
srknzl: Why this section does not include them? https://docs.hazelcast.com/imdg/4.2/clients/java.html#client-system-properties

srknzl: opened this one https://github.com/hazelcast/hz-docs/issues/94

DOCUMENTATION.md
srknzl: Why java client does not have null value but node.js client have?

DOCUMENTATION.md
srknzl: Ok I will try to add near cache word. I think `bigger` is ok btw.

src/config/ConfigBuilder.ts
srknzl: we need to preserve value as string. If we change it to be number via tryGetEnum, the logging service will not work. Should I update it too?

DOCUMENTATION.md
mdumandag: It also has that by default value

srknzl: Then I will open an issue since it is not documented. the property description in https://docs.hazelcast.com/imdg/4.2/clients/java.html#client-system-properties is not enough is it?

src/config/ConfigBuilder.ts
srknzl: Implemented in ef07339c

DOCUMENTATION.md
```diff
@@ -564,6 +565,28 @@ const client = await Client.newHazelcastClient(cfg);
...
+| hazelcast.invalidation.max.tolerated.miss.count        | 10                                  | number          | If missed invalidation count is bigger than this value, relevant cached data in a near cache will be made unreachable.                                                                                                                                                                                                                                                                                                              |
+| hazelcast.invalidation.reconciliation.interval.seconds | 60                                  | number          | This is a periodic task that scans cluster members periodically to compare generated invalidation events with the received ones from the client Near Cache.                                                                                                                                                                                                                                                         |
```
mdumandag, line 588:
```suggestion
| hazelcast.invalidation.reconciliation.interval.seconds | 60                                  | number          | Period of the task that scans cluster members to compare generated invalidation events with the received ones from the client Near Cache.                                                                                                                                                                                                                                                         |
```

src/config/Config.ts
```diff
@@ -167,7 +170,7 @@ const DEFAULT_PROPERTIES: Properties = {
     'hazelcast.invalidation.reconciliation.interval.seconds': 60,
     'hazelcast.invalidation.max.tolerated.miss.count': 10,
     'hazelcast.invalidation.min.reconciliation.interval.seconds': 30,
-    'hazelcast.logging.level': 'INFO',
+    'hazelcast.logging.level': LogLevel.INFO,
```
mdumandag, line 173: I think you should revert this. The default value should be `INFO`, not `2`

mdumandag (commented):
BTW, despite this failure, the Github Actions report the tests are passed 
```
  1624 passing (23m)
  73 pending
  1 failing

  1) ConfigBuilderTest
       properties:
     AssertionError: expected -1 to equal 'OFF'
      at Context.<anonymous> (test/unit/config/ConfigBuilderTest.js:155:61)
```

You should probably check that out

src/config/Config.ts
srknzl: no this is correct. The default properties should have the same types with config builder's return type so an integer(or enum). if logging level property is not given and its default value is used, logging level is directly passed to the default logger which expects an integer.  

otherwise these fail:
```js
it('default logging in case of empty property', async function () {
        client = await Client.newHazelcastClient({
            clusterName: cluster.id
        });
        sinon.assert.called(console.log);
    });
```
```js
it('default logging in case of empty property', async function () {
        client = await Client.newHazelcastClient({
            clusterName: cluster.id
        });
        sinon.assert.called(console.log);
    });
```

srknzl:
verify

src/config/Config.ts
mdumandag: Hmm, the default value is `2` (`LogLevel.INFO`), but if you pass something like 
```
newHazelcastClient({properties: {"hazelcast.logging.level": LogLevel.INFO}}) 
```
(or whatever the correct thing is :smile: ), it will fail, right?

srknzl: DEFAULT_PROPERTIES is internal isn't it? the user knows that default value is 'INFO', not LogLevel.INFO. we document it like this. Using enums is harder for users, so we decided to go with strings in the past.

mdumandag: OK, I am not really enjoying this but I can live with that

test/unit/config/ConfigBuilderTest.js
```diff
@@ -152,7 +153,7 @@ describe('ConfigBuilderTest', function () {
         expect(properties['hazelcast.invalidation.reconciliation.interval.seconds']).equal(50);
         expect(properties['hazelcast.invalidation.max.tolerated.miss.count']).to.equal(15);
         expect(properties['hazelcast.invalidation.min.reconciliation.interval.seconds']).to.equal(60);
-        expect(properties['hazelcast.logging.level']).to.be.equal('OFF');
+        expect(properties['hazelcast.logging.level']).to.be.equal(LogLevel.OFF);
```
mdumandag, line 156: Can we add some tests for the property validation and make sure that it works as expected?

src/config/Config.ts
mdumandag, line 173: Sorry to be paranoid but I have to come back to this again :smile: 

So, if I have a code like this, that won't work, right?

```
const c1 = await newClient({properties:{"hazelcast.logging.level": "INFO"}});
const c2 = await newClient(c1.getConfig());
```

I think to make this work, we either
- Revert the changes we did (LoggingService etc.)
- Make it accept values of the enum as well

mdumandag (commented):
Apart from one thing, it looks good

src/config/Config.ts
srknzl: I think reverting it is the way to go. And tryGetEnum will be used to just verify it. I think it is ok the code to be asymmetric there :)

srknzl: I did not know we have such an API(getConfig), so thanks for the heads up

mdumandag (approved):

PR #1003, opened 2021-08-02
Return saved promise for shutdown

### LINKED ISSUE #852: Refactor functions to return deferred promise
* The second call to such functions resolves immediately. e.g client.shutdown
* Using a deferred promise instead of a flag(e.g lifecycleService.active), we can return the same promise on concurrent calls.

srknzl:
Also hides replicated map proxy, and refactors listener service

closes #852

### DIFF
--- a/test/integration/backward_compatible/HazelcastClientTest.js
+++ b/test/integration/backward_compatible/HazelcastClientTest.js
@@ -80,22 +80,16 @@ configParams.forEach((cfg) => {
         let client;
         let managed;
 
-        before(async function () {
+        beforeEach(async function () {
             cluster = await RC.createCluster(null, null);
             await RC.startMember(cluster.id);
             cfg.clusterName = cluster.id;
             client = await Client.newHazelcastClient(cfg);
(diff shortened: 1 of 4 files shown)


test/integration/backward_compatible/HazelcastClientTest.js
```diff
@@ -80,22 +80,16 @@ configParams.forEach((cfg) => {
         let client;
         let managed;
 
-        before(async function () {
```
mdumandag, line 83: Why do we need to make it `before/afterEach`?

srknzl: Actually I don't know how this test was passing but, first running `getDistributedObjects returns all dist objects` and then `getDistributedObjects does not return removed object` causes names in the second test to be like this: 
```
Array(4) [map,
set,
map2,
map3]
```

So they affect each other. Plus, our new test shuts down the client so we need to create another one for other tests

mdumandag (approved):

PR #1013, opened 2021-08-09
Update dependencies

srknzl:


package.json
```diff
@@ -14,8 +14,8 @@
     },
     "devDependencies": {
         "@types/node": "~10.17.60",
-        "@typescript-eslint/eslint-plugin": "~4.28.5",
-        "@typescript-eslint/parser": "~4.28.5",
+        "@typescript-eslint/eslint-plugin": "~4.29.0",
+        "@typescript-eslint/parser": "~4.29.0",
```
mdumandag, line 18: These guys are releasing new versions non-stop :smile:

mdumandag (approved):

package.json
srknzl: They have a strange release cycle, they release every week even if nothing changes.

mdumandag: maybe we need to increase the period of the new version check to something like 2-4 weeks then?

srknzl: Maybe. Are these prs too much? :D If so, we should

srknzl: Do you see any way to set it 2 weeks etc. 

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#scheduleinterval

only weekly, monthly or daily is available

mdumandag: No, I am not aware of such a method. Personally, I would set it to monthly if these are the only options, but I am leaving the decision to you.

srknzl: I don't think we have critical dependencies that needs to be updated immediately, we have only `long.js` which should be mature already. As you can see from dependencies update PRs, it is not updated often. if our repo settings enable security warnings about dependencies, it should be enough.

So I will make it monthly.

PR #1021, opened 2021-08-16
Exclude codecs folder from coverage

srknzl:


mdumandag:
Can you share the coverage ratio before and after this change?

srknzl:
I think somehow coverage tool started to include test directory in the coverage. I will add a `include`: `src` to the nyc config

srknzl:
verify

.nycrc.json
```diff
@@ -1,4 +1,9 @@
 {
     "cache": true,
-    "reporter": "cobertura"
+    "reporter": "cobertura",
+    "exclude": [
+        "src/codec/custom/*.ts",
+        "src/codec/*.ts",
+        "test/**"
```
srknzl, line 7: default exclude is replaced, so we need to add test folder here. https://github.com/istanbuljs/schema/blob/master/default-exclude.js

srknzl: I tried to add "include": "src/**" etc. all variants of it, but it did not work(produced empty coverage). Whatever, this version works for us for now

mdumandag: what about other folders like benchmarks code_samples etc.? Are they also included now?

srknzl: no they are not, since they are not visited during test run

srknzl:
The new coverage is 90.29, master coverage is 92.55 and the drop is due to the fact that almost all of codecs are used. 

see codec folder in master https://codecov.io/gh/hazelcast/hazelcast-nodejs-client/tree/19cc6386d1fb2b1e228ad65b208ba23c6df7b247/src and this pr https://codecov.io/gh/hazelcast/hazelcast-nodejs-client/tree/e297c9c7b3caa892f5452fe1d83055fa5f73c1bb/src


this pr's latest commit: 
https://codecov.io/gh/hazelcast/hazelcast-nodejs-client/commit/e297c9c7b3caa892f5452fe1d83055fa5f73c1bb/

master latest commit: 
https://codecov.io/gh/hazelcast/hazelcast-nodejs-client/commit/19cc6386d1fb2b1e228ad65b208ba23c6df7b247#

srknzl:
should be unnecesary with #1082

PR #1024, opened 2021-08-25
SQL Tests adjustments around jet enabled config for backward compat

srknzl:
tested with 5.0 snapshot, 4.2, 4.2.1 servers


srknzl:
will be merged to another pr that adds explicit mappings.

PR #1027, opened 2021-08-27
API-868 Disable asserts in prod build

### LINKED ISSUE #482: Disable asserts in final build
There are multiple asserts done with standard `assert` module in the library. Those asserts should be enabled in dev and test environments, but disabled in the final library artifact as they have some performance impact (though, it's not critical).

A custom wrapper module could be created. This module would act as a proxy for the `assert` in dev and test environment, but in the final build it would act as a no-op object.

srknzl:
Adds a env variable check to decide whether to run assertions. Ideally we should skip runtime assertions but we should keep assertions that are used for validation. (Specified in the API docs). Let me know if there are any changes needed.

resolves #482

### DIFF
--- a/src/invocation/ClusterService.ts
+++ b/src/invocation/ClusterService.ts
@@ -107,7 +107,7 @@ export class ClusterService implements Cluster {
     }
 
     addMembershipListener(listener: MembershipListener): string {
-        assertNotNull(listener);
+        assertNotNull(listener, true);
 
         const registrationId = UuidUtil.generate().toString();
         this.listeners.set(registrationId, listener);
@@ -125,7 +125,7 @@ export class ClusterService implements Cluster {
     }
 
     removeMembershipListener(listenerId: string): boolean {
-        assertNotNull(listenerId);
+        assertNotNull(listenerId, true);
         return this.listeners.delete(listenerId);
     }
 
--- a/src/invocation/InvocationService.ts
+++ b/src/invocation/InvocationService.ts
@@ -15,7 +15,6 @@
  */
 /** @ignore *//** */
 
-import * as assert from 'assert';
 import {
     ClientNotActiveError,
     HazelcastInstanceNotActiveError,
@@ -39,7 +38,8 @@ import {
     cancelRepetitionTask,
     Task,
     deferredPromise,
-    DeferredPromise
+    DeferredPromise,
+    assert
 } from '../util/Util';
 import {ClientConfig} from '../config';
 import {ListenerService} from '../listener/ListenerService';
--- a/src/serialization/ObjectData.ts
+++ b/src/serialization/ObjectData.ts
@@ -15,7 +15,7 @@
  */
 /** @ignore *//** */
 
-import * as assert from 'assert';
+import { assert } from '../util/Util';
 import * as Long from 'long';
 import {BitsUtil} from '../util/BitsUtil';
 import {Data, DataInput, DataOutput, PositionalDataOutput} from './Data';
--- a/src/util/Util.ts
+++ b/src/util/Util.ts
@@ -29,8 +29,8 @@ const IS_DEVELOPMENT_MODE = process.env.HZ_NODEJS_ENV === 'development';
  * Used to only turn on asserts in development mode
  * @internal
  */
-export function assert(value: any, message?: string | Error): asserts value {
-    if (IS_DEVELOPMENT_MODE) {
+export function assert(value: any, message?: string | Error, alwaysAssert?: boolean): asserts value {
+    if (IS_DEVELOPMENT_MODE || alwaysAssert) {
         _assert(value, message);
     }
 }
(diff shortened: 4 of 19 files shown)


mdumandag:
What is the performance impact of these asserts? Can you share some before and after numbers? Maybe you can do a short test in the lab environment.

srknzl:
I added some missing changes. We need to be careful with assertions, which should always run etc. Also we are missing some api doc for errors thrown. I opened an issue about it. #1037

mdumandag:
I would like to see this change has a positive impact. Yes, you can run the benchmarks in the library or try to write some artificial benchmark that compares having asserts (old state) and not having them with runtime checks (new state)

Something like this

```
def f1():
  alwaysDoAssertion()

def f1():
  doNotPerformAssertionIfRuntimeCheckFails()
```

BTW, I see that you forced lots of assertion all the time (which is the correct thing IMHO), how many asserts that are left which can be disabled at runtime? If the number is low, maybe we can live with them

srknzl:
I see these:

in getMember(uuid) to check uuid is not null
in getType() in Util.ts which gets type string of a value for serialization
in notify() in an Invocation to check if client message is null
in invocation service send() to check if connection is not null
in ObjectDataInput assertAvailable() which I think is used in all readX() methods

srknzl:
Closing since performance dropped, not increased. I used the benchmark in the repo. The following numbers denote ops per second.

Local machine:

random:

192128 when asserts disabled conditionally
195624 when asserts always enabled

set:

131481 when asserts disabled conditionally
154000 when asserts always enabled


Lab:

1 client 1 server on different machines

random:

82116 when asserts disabled conditionally
112000 when asserts always enabled

set:

 67800 when asserts disabled conditionally
100000 when asserts always enabled

srknzl:
I think performance dropped due to the if check we perform for each assertion. Can you think of any other way to achive this without an if?

srknzl:
I think I will try this again with replacing assertion methods will empty methods depending on the environment variable.

srknzl:
@mdumandag The benchmark result of my second version:

node version 12, server version 4.2, 1 client 1 server on different machines, 20 million operations with default concurrency:


**random**:

old 103243
new 112429

**set**:

 old 76621
 new 84626

**get**:

 old 76641
 new 78934

srknzl:
If you are ok merging this, I will update some docstrings of methods in util.ts.

mdumandag:
but now you don't perform null checks in the get/set methods for example now, right? I don't think we should skip checking them

srknzl:
No, the only conditional checks are the ones listed in the comment https://github.com/hazelcast/hazelcast-nodejs-client/pull/1027#issuecomment-911639339

srknzl:
another benchmark after https://github.com/hazelcast/hazelcast-nodejs-client/pull/1027/commits/2cd5f8bf43d1740ce95fc1f9ce2a1162737e8914

same setup as the previous benchmark(lab)

**random**:

old 92439
new 110170

**get**:

old 74380
new 105834

**set**:

old 80461
new 81933

srknzl:
Are we going to merge this?

src/serialization/ObjectData.ts
```diff
@@ -561,8 +561,8 @@ export class ObjectDataInput implements DataInput {
     }
 
     private assertAvailable(numOfBytes: number, pos: number = this.pos): void {
-        assert(pos >= 0);
-        assert(pos + numOfBytes <= this.buffer.length);
+        assertCond(pos >= 0);
+        assertCond(pos + numOfBytes <= this.buffer.length);
```
mdumandag, line 565: Looking at the Java side, these seem to be regular checks, not asserts. So, they are checked in the release builds as well

src/util/Util.ts
```diff
@@ -15,45 +15,112 @@
...
+            expected: null,
+            actual: v
+        });
+    }
 }
 
 /** @internal */
 export function assertArray(x: any): void {
-    assert(Array.isArray(x), 'Should be array.');
+    const isArray = Array.isArray(x);
```
mdumandag, line 51: TBH, the old version looks more compact. I can say a similar thing for other assertions as well. What is the reason we changed them?

src/invocation/InvocationService.ts
```diff
@@ -172,7 +172,7 @@ export class Invocation {
     }
 
     notify(clientMessage: ClientMessage): void {
-        assert(clientMessage != null, 'Response can not be null');
+        assertCond(clientMessage != null, 'Response can not be null');
```
mdumandag, line 175: I feel like we don't even need a conditional check here. Looking at the code, it is not possible to pass a null client message here

src/invocation/ClusterService.ts
```diff
@@ -78,7 +79,7 @@ export class ClusterService implements Cluster {
      * @return The member that was found, or undefined if not found.
      */
     getMember(uuid: string): MemberImpl | undefined {
-        assertNotNull(uuid);
+        assertNotNullCond(uuid);
```
mdumandag, line 82: Same as below, I have checked all code paths that use this. It is not possible to pass a null UUID here

src/invocation/InvocationService.ts
```diff
@@ -529,7 +529,7 @@ export class InvocationService {
     }
 
     private send(invocation: Invocation, connection: Connection): Promise<void> {
-        assert(connection != null);
+        assertCond(connection != null);
```
mdumandag, line 532: I also feel like this cannot be null anywhere.

src/util/Util.ts
```diff
@@ -15,45 +15,112 @@
...
+import {AssertionError} from 'assert';
+
+// Used to assert only in development mode
+const IS_DEVELOPMENT_MODE = process.env.HZ_NODEJS_ENV === 'development';
+
+/** @internal */
+export const assert : (value: any, message?: string | Error) => void = _assert;
+
+/** @internal */
+export const assertCond : (value: any, message?: string | Error) => void = IS_DEVELOPMENT_MODE ? _assert : () => {};
```
mdumandag, line 33: Minor: Maybe it would be good to be more explicit on the naming: `assertCond` -> `assertConditional`

mdumandag (commented):
I have gone over the places that use the assertions you made conditional. To me, they are not very useful, as it is not possible for the current code base to fail such assertions. If you want to keep them as sanity checks, I would be okay with that. So, I am leaving the decision to you.

srknzl:
Ok let's resolve left conditional checks: `assertNotNullConditional` in Util.ts and ClusterService's `getMember`.

After that I will remove all these assertions in this pr

PR #1034, opened 2021-09-01
Fix sql tests and code samples 

srknzl:
Fixes sql tests for backward compat tests and normal tests.

fixes include:

create mapping need
make checking server version is >= 5 correct.
conditionally give jet enabled flag
updates code samples to include create mapping

### DIFF
--- a/test/integration/backward_compatible/sql/DataTypeTest.js
+++ b/test/integration/backward_compatible/sql/DataTypeTest.js
@@ -82,6 +82,72 @@ describe('Data type test', function () {
         await RC.startMember(cluster.id);
     });
 
+    /**
+     * Creates mapping for SQL queries. In 5.0, users started to write explicit mapping for SQL queries against maps.
+     * @param valueFormat SQL column type, case insensitive
+     */
+    const createMapping = async (valueFormat) => {
+        if (!serverVersionNewerThanFive) {
+            // Before 5.0, mappings are created implicitly, thus we don't need to create explicitly.
+            return;
+        }
+        const createMappingQuery = `
+            CREATE MAPPING ${mapName} (
+                __key INT,
+                this ${valueFormat.toUpperCase()}
+            )
+            TYPE IMAP
+            OPTIONS (
+                'keyFormat' = 'int',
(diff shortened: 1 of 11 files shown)


test/integration/backward_compatible/sql/DataTypeTest.js
```diff
@@ -73,11 +74,54 @@ describe('Data type test', function () {
...
+            return;
+        }
+
+        const entries = Object.entries(columns);
+
+        const numberOfColumns = entries.length;
+
+        const columnsString = entries.reduce((accumulator, current, currentIndex) => {
+            return accumulator + `${current[0]} ${current[1].toUpperCase()}${currentIndex !== numberOfColumns - 1 ? ',' : ''}\n`;
+        }, '');
```
mdumandag, line 104: Would something like this be more readable?

```js
Object.entries(columns).map(column => `${column[0]} ${column[1].toUpperCase()}`).join(',\n');
```

test/integration/backward_compatible/sql/DataTypeTest.js
```diff
@@ -73,11 +74,54 @@ describe('Data type test', function () {
...
     });
 
+    /**
+     * Creates portable mapping for SQL queries. In 5.0, users started to write explicit mapping for SQL queries against maps.
+     * @param keyFormat Key format
+     * @param factoryId Portable's factory id
+     * @param classId Portable's class id
+     * @param columns Columns as a dict where keys are column names, and values are case insensitive value formats.
+     */
+    const createMappingForPortable = async (keyFormat, factoryId, classId, columns) => {
```
mdumandag, line 92: Why this is defined here, both other in TestUtil?

mdumandag (approved):
Approving with minor comments

test/integration/backward_compatible/sql/DataTypeTest.js
srknzl: Since this is used only in here I wanted to keep it here.

test/integration/backward_compatible/sql/DataTypeTest.js
srknzl: yeah thanks

PR #1036, opened 2021-09-02
Eslint config changes

srknzl:
Added ignorePatterns so that IDEs does not show up an error like "the file does not match your project config:" when .eslintrc.js file is opened. Our cli command already ignores that since we use --ext .ts flag.

Change config in eslintrc in test folder so that typescript files in test folder is linted. There is only one typescript file in test folder which is UserCodeTest.ts, changing it to the following was not give linting error, not it gives:

```ts
/**
 * The following file is used to validates correctness of
 * declarations (.d.ts) of the client library
 * produced by TypeScript compiler.
 */

import { Client } from '../../lib';

(async (): Promise<void> => {
    const cliensddddddddddddddddddddddddddddddsddddddddddddddddddddddddddddddddddddddddddddddddddddddt = await Client.newHazelcastClient();
    const rb = await cliensddddddddddddddddddddddddddddddsddddddddddddddddddddddddddddddddddddddddddddddddddddddt.getRingbuffer('my-rb');
    await rb.add(1);
})();
```


test/.eslintrc
```diff
@@ -10,6 +10,10 @@
         "../.eslintrc",
         "plugin:mocha/recommended"
     ],
+    "overrides": [{
+        "files": ["**/*.ts"],
+        "extends": ["../src/.eslintrc.js"]
```
mdumandag, line 15: I don't know much about this but is this overrides the `extends` defined just above?

srknzl: this defines separate config for files under this directory matching the pattern. we use the config in src for typescript files for them.

srknzl: which means it overrides the config of certain files with different config

src/.eslintrc.js
```diff
@@ -24,6 +24,9 @@ module.exports = {
     plugins: [
         '@typescript-eslint',
     ],
+    ignorePatterns: [
+        '.eslintrc.js'
```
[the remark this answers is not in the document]
srknzl: Another possibility is that IDEs run eslint in the src directory. But .eslintignore file is in the root directory. Therefore it does not cause any affect.

test/.eslintrc
mdumandag: But we have some config options defined in the top level .eslintrc file that are not there in src/.eslintrc.js, so I feel like we change something (but not sure :smile: as I don't know eslint much).

BTW, I am really confused by the amount of different .eslint config files we have :smile: Ideally, I guess we just need to have one defined in the top level, shouldn't we?

.eslintignore
```diff
@@ -1,3 +1,4 @@
 src/codec/*.ts
 src/codec/custom/*.ts
 src/.eslintrc.js
+lib/**
```
srknzl, line 4: This is needed because now we call eslint with: "eslint ."

srknzl:
I forgot the change for checking ts files in test folder adding it now..

srknzl:
I have to create another eslintrc file in test/user_code otherwise I have to extend the eslintrc file in the root folder. This is how eslint works unfortunately...

.eslintrc
```diff
@@ -101,5 +101,43 @@
...
+                "mocha"
+            ],
+            "extends": [
+                "plugin:mocha/recommended"
+            ],
+            "rules": {
+                "mocha/no-hooks-for-single-case": "off",
+                "mocha/no-setup-in-describe": "off",
+                "camelcase": "off",
+                "max-len": ["warn", 130],
```
mdumandag, line 122: There are some configurations defined for them above. Can we use the settings above to simplify this a bit more?

.eslintrc
```diff
@@ -101,5 +101,43 @@
...
+            "files": ["code_samples/**/*"],
+            "env": {
+                "es2020": true
+            },
+            "parserOptions": {
+                "ecmaVersion": 11
+            }
+        }
+    ],
+    "ignorePatterns": ["ts.eslintrc.js"]
```
mdumandag, line 142: Does this have some special meaning? I don't see a file called `ts.eslintrc.js`

srknzl: Ah sorry I created it and deleted it, fixing..

.eslintrc
srknzl: What do you mean? We don't use them since we have different values (130 for line length, root config has 120) for test folder.

Do you mean we should have the same values?

.eslintrc
mdumandag: Yes, I mean they are not hugely different, if there is no special reason for the difference, I would just use what is defined on the top level

.eslintrc
mdumandag: I mean they are not hugely different, if there is no special reason for the difference, I would just use what is defined on the top level

srknzl: Due to test names and nesting, line length is generally longer in test files. I think it can stay like this. Otherwise the code will be ugly with line breaks in test names.

Regarding "mocha/" rules they come from plugin so we cannot add them to root rules.

camelcase: "off" exists since there are some an_integer like variable names in a serialization test I suppose.

and the "padded-blocks" exists only on test files since recently I wanted some consistency about it. I think enabling it for typescript files is not too necessary since I don't feel files look inconsistent in style. This is probably due to every file has different structure whereas every test file has the describe block.

 I think making overrides shorter is not a simplification, it restricts out ability to choose rules for each glob pattern

mdumandag (approved):

PR #1046, opened 2021-09-14
Make sql execute return promise api-877

srknzl:
Makes `execute` and `executeStatement` methods in `SqlService` return promises of `SqlResult`. Update tests, code samples and documentation accordingly.

Note that we do not need any conditional tests for back compatibility because using `await` with a non async function works. It will convert the returned value to a promise like `Promise.resolve(returnedValue)` implicitly but I think it is better not to change tests for readability. The promise overhead only affects ~50 tests and the overhead is very small.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,4 @@ src/codec/*.ts
 src/codec/custom/*.ts
 src/.eslintrc.js
 lib/**
+docs/**
--- a/code_samples/sql-basic-usage.js
+++ b/code_samples/sql-basic-usage.js
@@ -15,7 +15,7 @@
  */
 'use strict';
 
-const { Client, SqlColumnType, HazelcastSqlException } = require('hazelcast-client');
+const { Client, SqlColumnType } = require('hazelcast-client');
 
 (async () => {
     try {
@@ -42,50 +42,31 @@ const { Client, SqlColumnType, HazelcastSqlException } = require('hazelcast-clie
         await map.put('key3', 3);
 
         let result;
-        try {
-            result = await client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1]);
-            const rowMetadata = result.getRowMetadata();
-            const columns = rowMetadata.getColumns();
 
-            console.log('Columns:');
-            for (const column of columns) {
-                console.log(`${column.name}: ${SqlColumnType[column.type]}`);
-            }
+        result = await client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1]);
+        const rowMetadata = result.rowMetadata;
+        const columns = rowMetadata.getColumns();
 
-            console.log('Rows from query 1:');
-            for await (const row of result) {
-                // By default a row is a plain javascript object. Keys are column names and values are column values
-                console.log(`${row['__key']}: ${row['this']}`);
-            }
-        } catch (e) {
-            if (e instanceof HazelcastSqlException) {
-                // HazelcastSqlException is thrown if an error occurs during SQL execution.
-                console.log(`An SQL error occurred while running SQL: ${e}`);
-            } else {
-                // for all other errors
-                console.log(`An error occurred while running SQL: ${e}`);
-            }
+        console.log('Columns:');
+        for (const column of columns) {
+            console.log(`${column.name}: ${SqlColumnType[column.type]}`);
         }
 
-        try {
-            // You can set returnRawResult to true to get rows as `SqlRow` objects
-            result = await client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1], {
-                returnRawResult: true
-            });
+        console.log('Rows from query 1:');
+        for await (const row of result) {
+            // By default a row is a plain javascript object. Keys are column names and values are column values
+            console.log(`${row['__key']}: ${row['this']}`);
+        }
 
-            console.log('Rows from query 2:');
-            for await (const row of result) {
-                console.log(`${row.getObject('__key')}: ${row.getObject('this')}`);
-                console.log(JSON.stringify(row.getMetadata().columns)); // SqlRow has a getter for row metadata
-            }
-        } catch (e) {
-            if (e instanceof HazelcastSqlException) {
-                // HazelcastSqlException is thrown if an error occurs during SQL execution.
-                console.log(`An SQL error occurred while running SQL: ${e}`);
-            } else {
-                // for all other errors
-                console.log(`An error occurred while running SQL: ${e}`);
-            }
+        // You can set returnRawResult to true to get rows as `SqlRow` objects
+        result = await client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1], {
+            returnRawResult: true
+        });
+
+        console.log('Rows from query 2:');
+        for await (const row of result) {
+            console.log(`${row.getObject('__key')}: ${row.getObject('this')}`);
+            console.log(JSON.stringify(row.getMetadata().columns)); // SqlRow has a getter for row metadata
         }
 
         await client.shutdown();
--- a/code_samples/sql-basic-usage.js
+++ b/code_samples/sql-basic-usage.js
@@ -44,8 +44,8 @@ const { Client, SqlColumnType, HazelcastSqlException } = require('hazelcast-clie
         let result;
         try {
             result = await client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1]);
-            const rowMetadata = await result.getRowMetadata();
-            const columns = await rowMetadata.getColumns();
+            const rowMetadata = result.getRowMetadata();
+            const columns = rowMetadata.getColumns();
 
             console.log('Columns:');
             for (const column of columns) {
--- a/code_samples/sql-basic-usage.js
+++ b/code_samples/sql-basic-usage.js
@@ -35,8 +35,7 @@ const { Client, SqlColumnType, HazelcastSqlException } = require('hazelcast-clie
                 'valueFormat' = 'double'
             )
         `;
-        // executions are async, await on update count to wait for execution.
-        await client.getSql().execute(createMappingQuery).getUpdateCount();
+        await client.getSql().execute(createMappingQuery);
 
         await map.put('key1', 1);
         await map.put('key2', 2);
@@ -44,7 +43,7 @@ const { Client, SqlColumnType, HazelcastSqlException } = require('hazelcast-clie
 
         let result;
         try {
-            result = client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1]);
+            result = await client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1]);
             const rowMetadata = await result.getRowMetadata();
             const columns = await rowMetadata.getColumns();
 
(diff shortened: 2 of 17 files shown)


.eslintignore
```diff
@@ -2,3 +2,4 @@ src/codec/*.ts
 src/codec/custom/*.ts
 src/.eslintrc.js
 lib/**
+docs/**
```
srknzl, line 5: Needed because of linting does `eslint .`. I forgot to add docs folder

test/integration/backward_compatible/sql/DataTypeTest.js
```diff
@@ -491,14 +490,14 @@ describe('Data type test', function () {
...
+            result = await TestUtil.getSql(client).execute(
                 `SELECT * FROM ${mapName} WHERE this > ? AND this < ? ORDER BY __key ASC`,
                 [new LocalDate(50001, 1, 1), new LocalDate(50005, 5, 5)]
             );
         } else {
-            result = TestUtil.getSql(client).execute(
+            result = await TestUtil.getSql(client).execute(
                 `SELECT * FROM ${mapName} WHERE this > CAST (? AS DATE) AND this < CAST(? AS DATE) ORDER BY __key ASC`,
-                ['50001-01-01', '50005-05-05']
+                ['+50001-01-01', '+50005-05-05']
```
srknzl, line 500: See https://github.com/hazelcast/hazelcast/issues/19532 for why this is needed.

test/integration/backward_compatible/sql/DataTypeTest.js
```diff
@@ -483,22 +482,22 @@ describe('Data type test', function () {
             `
             var map = instance_0.getMap("${mapName}");
             for (var key = 1; key < 12; key++) {
-                map.set(new java.lang.Integer(key), java.time.LocalDate.of(key+50002,key+1,key));
+                map.set(new java.lang.Integer(key), java.time.LocalDate.of(key+5002,key+1,key));
```
srknzl, line 485: I changed this since year became and 16 bit integer in 5.0

src/sql/SqlService.ts
```diff
@@ -359,21 +357,20 @@ export class SqlServiceImpl implements SqlService {
...
+            return this.invocationService.invokeOnConnection(connection, requestMessage).then(clientMessage => {
                 SqlServiceImpl.handleExecuteResponse(clientMessage, res);
+                return res;
             }).catch(err => {
-                res.onExecuteError(
-                    this.rethrow(err, connection)
-                );
+                const error = this.rethrow(err, connection);
+                res.onExecuteError(error);
+                throw error;
```
mdumandag, line 366: I honestly like this approach but I am not sure why the Java client returns the result regardless of whether the error is received or not.

src/sql/SqlService.ts
```diff
@@ -284,7 +282,7 @@ export class SqlServiceImpl implements SqlService {
         );
     }
 
-    executeStatement(sqlStatement: SqlStatement): SqlResult {
+    executeStatement(sqlStatement: SqlStatement): Promise<SqlResult> {
```
mdumandag, line 285: After this change, I assume the methods on the SqlResult (updateCount etc.) not have to return Promise, right? 
Should we change those as well? Maybe we can leave them like this for being future-proof, but I am not sure

srknzl: What do you mean by being future-proof? I think we can do that. This will simplify the SqlResult.

src/sql/SqlService.ts
srknzl: Can we make nodejs client different? Should we discuss this in the standup?

src/sql/SqlService.ts
srknzl: Since we will always return a promise from this execute()(the one that optionally waits execution response), the user will always wait in order to get a SqlResult object. Then we don't need to return promises from other methods.

Also I don't think we need another execute method in the future(sync or async version) because the user can use this execute method for both async and sync. By async, I mean without waiting the execute response. Therefore we can simplify SqlResult and delete executeDeferred variable altogether. WDYT?

code_samples/sql-data-types.js
```diff
@@ -54,15 +54,14 @@ const varcharExample = async (client) => {
...
+    await client.getSql().execute(createMappingQuery);
 
     for (let key = 0; key < 10; key++) {
         await someMap.set(key, key.toString());
     }
 
     try {
-        const result = client.getSql().execute('SELECT * FROM varcharMap WHERE this = ? OR this = ?', ['7', '2']);
+        const result = await client.getSql().execute('SELECT * FROM varcharMap WHERE this = ? OR this = ?', ['7', '2']);
         const rowMetadata = await result.getRowMetadata();
```
mdumandag, line 65: We don't need to await the result.xxx methods (apart from close) anymore. Can you update those as well?

src/sql/SqlService.ts
mdumandag: I meant if we want to use the same API when we introduce NonBlockingSqlResult in the future, but we could introduce a new type to represent that in the future. So, resolving this

src/sql/SqlResult.ts
```diff
@@ -98,39 +98,37 @@ export interface SqlResult extends AsyncIterable<SqlRowType> {
      * Returns row metadata of the result.
      * @returns SQL row metadata if rows exist in the result; otherwise null.
      */
-    getRowMetadata(): Promise<SqlRowMetadata | null>;
+    getRowMetadata(): SqlRowMetadata | null;
```
mdumandag, line 101: Not related to here but, above, we should update `### Usage for update count` section and remove await

src/sql/SqlResult.ts
```diff
@@ -330,8 +307,6 @@ export class SqlResultImpl implements SqlResult {
             this.updateCount = updateCount;
             this.closed = true;
         }
-        this.responseReceived = true;
```
mdumandag, line 333: There are some refactorings that can be done here. If you want, you can do it later. It is not required for correctness but for example, we might remove the code that assumes the user can close the result before the response arrives.

src/sql/SqlService.ts
```diff
@@ -128,26 +128,24 @@ import {SqlPage} from './SqlPage';
  */
 export interface SqlService {
     /**
-     * Executes SQL and returns an SqlResult.
+     * Executes SQL and returns an {@link SqlResult}.
```
mdumandag, line 131: ` * let result = client.getSqlService().execute('SELECT * FROM person');` above doc must be updated

code_samples/sql-basic-usage.js
```diff
@@ -35,18 +35,17 @@ const { Client, SqlColumnType, HazelcastSqlException } = require('hazelcast-clie
...
         await map.put('key2', 2);
         await map.put('key3', 3);
 
         let result;
         try {
-            result = client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1]);
-            const rowMetadata = await result.getRowMetadata();
-            const columns = await rowMetadata.getColumns();
+            result = await client.getSql().execute('SELECT __key, this FROM myMap WHERE this > ?', [1]);
+            const rowMetadata = result.getRowMetadata();
```
mdumandag, line 47: This is not valid after recent changes.

And, how does the code sample checker test passed? I see this error in logs 
```
2021-09-24T10:27:11.6504752Z An error occurred while running SQL: TypeError: result.getRowMetadata is not a function
```

but it didn't exit with a non-zero code. We should definitely fix that as well.. Maybe create an issue and fix it after 5.0?

code_samples/sql-data-types.js
```diff
@@ -54,16 +54,15 @@ const varcharExample = async (client) => {
...
 
     for (let key = 0; key < 10; key++) {
         await someMap.set(key, key.toString());
     }
 
     try {
-        const result = client.getSql().execute('SELECT * FROM varcharMap WHERE this = ? OR this = ?', ['7', '2']);
-        const rowMetadata = await result.getRowMetadata();
+        const result = await client.getSql().execute('SELECT * FROM varcharMap WHERE this = ? OR this = ?', ['7', '2']);
+        const rowMetadata = result.getRowMetadata();
```
mdumandag, line 65: Same as above

code_samples/sql-data-types.js
```diff
@@ -102,19 +101,18 @@ const integersExample = async (client) => {
...
     }
 
     try {
-        const result = client.getSql().execute(
+        const result = await client.getSql().execute(
             'SELECT * FROM bigintMap WHERE this > ? AND this < ?',
             [long.fromNumber(10), long.fromNumber(18)]
         );
-        const rowMetadata = await result.getRowMetadata();
+        const rowMetadata = result.getRowMetadata();
```
mdumandag, line 115: .

code_samples/sql-data-types.js
```diff
@@ -184,12 +181,12 @@ const objectExample = async (client, classId, factoryId) => {
...
         // instead of `this`.
         // This is true only for complex custom objects like portable and identified serializable.
-        const result = client.getSql().execute(
+        const result = await client.getSql().execute(
             'SELECT __key, this FROM studentMap WHERE age > CAST(? AS INTEGER) AND age < CAST(? AS INTEGER)',
             [3, 8]
         );
 
-        const rowMetadata = await result.getRowMetadata();
+        const rowMetadata = result.getRowMetadata();
```
mdumandag, line 189: .

src/sql/SqlResult.ts
```diff
@@ -95,42 +95,39 @@ export interface SqlResult extends AsyncIterable<SqlRowType> {
     close(): Promise<void>;
 
     /**
-     * Returns row metadata of the result.
+     * SQL row metadata if rows exist in the result; otherwise null.
```
mdumandag, line 98: Can you say `@returns` for properties?

src/sql/SqlService.ts
```diff
@@ -359,21 +356,20 @@ export class SqlServiceImpl implements SqlService {
...
 
-            this.invocationService.invokeOnConnection(connection, requestMessage).then(clientMessage => {
+            return this.invocationService.invokeOnConnection(connection, requestMessage).then(clientMessage => {
                 SqlServiceImpl.handleExecuteResponse(clientMessage, res);
+                return res;
             }).catch(err => {
-                res.onExecuteError(
-                    this.rethrow(err, connection)
-                );
+                const error = this.rethrow(err, connection);
```
mdumandag, line 363: I think this logic is a bit flawed. Imagine that you executed a query and received an error (but like valid response with `response_error`, so you throw it from `handleExecuteResponse` above). With this catch, we will also catch it. Not only we will call `onExecuteError` (no problem, just not ideal) a second time, but if the connection is closed in between, we will report to user a `Cluster topology changed while a query was executed` error, while in fact it was a different error.

I believe you should do invokeOnConnection().catch().then()

code_samples/sql-basic-usage.js
srknzl: 🤔 Sure let's fix let me see..

src/sql/SqlResult.ts
srknzl: Leftover deleted

src/sql/SqlService.ts
srknzl: Now we call onExecuteError once like my java pr https://github.com/hazelcast/hazelcast/pull/19612/files

code_samples/sql-basic-usage.js
srknzl: Fixed it, by removing unnecessary try catch blocks

mdumandag (approved):

PR #1053, opened 2021-09-20
ClientReconnectTest flaky test fix

srknzl:
Fix for https://github.com/hazelcast/client-compatibility-suites/actions/runs/1251558570

I explained the fail in comments

### DIFF
--- a/test/integration/backward_compatible/ClientReconnectTest.js
+++ b/test/integration/backward_compatible/ClientReconnectTest.js
@@ -18,7 +18,7 @@
 const { expect } = require('chai');
 const RC = require('../RC');
 const { Client } = require('../../../');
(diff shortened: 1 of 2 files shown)


srknzl:
This pr should fix 4.2.0 fail, I did not really understand why 4.0.0 test failed though. 🤔  https://github.com/hazelcast/client-compatibility-suites/runs/3646558223?check_suite_focus=true

@mdumandag Can you see why?

srknzl:
I think with this change we don't need the version check anymore and can delete the version check for `create proxy while member is down, member comes back`. also `member restarts, while map.put in progress 2` should pass since we wait for connections to be closed. 

I think 4.0.0 fail is due to connection close too, but since we don't close socket on error prior to 4.0.2 it hangs 120 seconds and times out.

mdumandag (approved):

PR #1055, opened 2021-09-24
Provide suggestions for `HazelcastSqlException`

### LINKED ISSUE #1045: [TRACKING ISSUE] Provide CREATE MAPPING suggestion on existing but not mapped IMap exception

See https://github.com/hazelcast/hazelcast/pull/19484 for details.

---

`HazelcastSqlException` gains `suggestion` (suggested SQL statement to remediate experienced error).
In a situation when an `IMap` exists but is not mapped, `suggestion` contains `CREATE MAPPING` statement for this particular `IMap`.

Added SqlMappingDdl client operation - one can pass a name of an `IMap` and get `CREATE MAPPING` statement for it.

Depends upon [#391](https://github.com/hazelcast/hazelcast-client-protocol/pull/391)

Fixes #19464

srknzl:
resolves #1045

Protocol pr: https://github.com/hazelcast/hazelcast-client-protocol/pull/401

### DIFF
--- a/src/core/HazelcastError.ts
+++ b/src/core/HazelcastError.ts
@@ -412,6 +412,8 @@ export class UndefinedErrorCodeError extends HazelcastError {
  */
 export class HazelcastSqlException extends HazelcastError {
     private readonly code: number;
+    /** Suggested SQL statement to remediate experienced error. */
+    readonly suggestion : string | null;
     /**
      * ID of the member that caused or initiated an error condition.
      * This can be the client's UUID if error is due to the client.
@@ -422,11 +424,13 @@ export class HazelcastSqlException extends HazelcastError {
         originatingMemberId: UUID,
         code: number,
         msg: string,
+        suggestion : string | null = null,
         cause?: Error,
-        serverStackTrace?: ServerErrorStackElement[]
+        serverStackTrace?: ServerErrorStackElement[],
     ) {
         super(msg, cause, serverStackTrace);
         this.code = code;
         this.originatingMemberId = originatingMemberId;
+        this.suggestion = suggestion;
     }
 }
--- a/src/sql/SqlError.ts
+++ b/src/sql/SqlError.ts
@@ -22,10 +22,17 @@ import {UUID} from '../core';
  * @internal
  */
 export class SqlError {
+
     constructor(
+        /** The error code. */
         readonly code: number,
+        /** The error message. */
         readonly message: string,
-        readonly originatingMemberId: UUID
+        /** UUID of the member that caused or initiated an error condition. */
+        readonly originatingMemberId: UUID,
+        _ : boolean,
+        /** Suggested SQL statement to remediate experienced error. */
+        readonly suggestion: string
     ) {
     }
 }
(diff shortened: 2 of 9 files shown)


src/sql/SqlError.ts
```diff
@@ -22,10 +22,17 @@ import {UUID} from '../core';
...
+        /** The error code. */
         readonly code: number,
+        /** The error message. */
         readonly message: string,
-        readonly originatingMemberId: UUID
+        /** UUID of the member that caused or initiated an error condition. */
+        readonly originatingMemberId: UUID,
+        _ : boolean,
+        /** Suggested SQL statement to remediate experienced error. */
+        readonly suggestion: string
```
mdumandag, line 35: Shouldn't this be `string | null` as well

src/sql/SqlError.ts
```diff
@@ -22,10 +22,17 @@ import {UUID} from '../core';
...
  */
 export class SqlError {
+
     constructor(
+        /** The error code. */
         readonly code: number,
+        /** The error message. */
         readonly message: string,
-        readonly originatingMemberId: UUID
+        /** UUID of the member that caused or initiated an error condition. */
```
mdumandag, line 31: This can also be the UUID of the client.

src/core/HazelcastError.ts
```diff
@@ -422,11 +424,13 @@ export class HazelcastSqlException extends HazelcastError {
         originatingMemberId: UUID,
         code: number,
         msg: string,
+        suggestion : string | null = null,
```
mdumandag, line 427: I think you should move this parameter to last, or places like `SqlService#rethrow` or `SqlService#toHazelcastSqlException` would fail (and I am not sure how the tests are passing? Even the TypeScript compiler should alert that, right?)

srknzl: when you provide default value to a parameter it is optional as well. So it is like `suggestion? : string | null`. The only difference is that `suggestion?`'s default value is undefined.

mdumandag: But, if you call `new HazelcastSqlException(uuid, errorcode, msg, err)` the `err` will be set to `suggestion` and that is wrong

srknzl: Ah you are right, in rethrow methods it should be: 

new HazelcastSqlException(uuid, errorcode, msg, undefined, err)

passing undefined to a argument leaves its default value.

test/unit/sql/SqlResult.js
```diff
@@ -212,8 +212,8 @@ describe('SqlResultTest', function () {
 
         beforeEach(function () {
             fakeSqlService = {
-                toHazelcastSqlException: sandbox.fake((err) => new HazelcastSqlException(null, 1, '', err)),
-                rethrow: sandbox.fake((err) => new HazelcastSqlException(null, 1, '', err)),
+                toHazelcastSqlException: sandbox.fake((err) => new HazelcastSqlException(null, 1, '', undefined, err)),
+                rethrow: sandbox.fake((err) => new HazelcastSqlException(null, 1, '', undefined, err)),
```
srknzl, line 216: These were not causing test fails it's good to fix them as well

mdumandag (approved):

PR #1092, opened 2021-10-18
New release workflow [API-929]

srknzl:
## New release workflow. 

The idea is coming from https://hazelcast.atlassian.net/wiki/spaces/HZC/pages/3803480072/Common+Release+Process+For+Clients

There are two new workflows that will be triggered manually: 
* one for testing and creating a release pack
*  the other for actually releasing the pack and generating api docs automatically.

The verification steps(i.e kubernetes, soak, back compat) and publish workflow will use a nodejs script to download the latest verified release pack for a given branch.  So, after the release pack is ready, the release verifier should only know the branch name from which the release will be done.

The nodejs script is used like this: 

`node scripts/download-release-candidate.js -b <branch-of-release> -u <username> -p <token> `

and downloads the latest release pack matched using github api.

The only thing left to do manually is to tag the commit of the release. If I can find a way to do that easily(a action that does that) I can add that too.

I have tried the release pack creation workflow in my fork, it works. Also tested the nodejs script. The other publish workflow is not tried. Maybe we can try it with a test registry.

## Description of the script:

It's job is only to download the `suitable` release pack.

See the script's main:
https://github.com/hazelcast/hazelcast-nodejs-client/pull/1092/files#diff-939c198801c9a02a1ff5d9c83e6a7a977c803f20887ff1fe2d2029c51aea8bb3R116

It will do that via: 

1. Find the latest commit of the given branch
2. Search for an artifact with that commit(Note that all artifacts are preverified, where verify means unit and integration tests of the client is run on that pack)
3. Download that release pack

So it will fail if there is no package created with that branch name. 

## Client compatibliity suites counter part

 https://github.com/hazelcast/client-compatibility-suites/pull/8


srknzl:
closing because this is discontinued. Our QA left and we need to revisit this when the new QA joins.

PR #1127, opened 2021-11-11
Change schedule of nightly tests

srknzl:
Tests fail to download 5.0 snapshot server jars. As an easy possible fix, I changed the time the tests run to reduce load on sonatype repositories. Note that python client uses 2AM. Hazelcast repos might be using the same runners, causing the sonatype to block downloads.

If the issue stills same after this, we can introduce retry mechanism to downloader script. But that can be hard for 4.x since they still use .sh files.

### DIFF
--- a/.github/workflows/nightly_runner_maintenance.yml
+++ b/.github/workflows/nightly_runner_maintenance.yml
@@ -2,7 +2,7 @@ name: Runs nightly tests on maintenance branches
 on:
     workflow_dispatch:
     schedule:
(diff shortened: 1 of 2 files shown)


mdumandag (approved):

srknzl:
I added caching change here too. It caches dependencies each date, according to download-rc script hash and OS.

srknzl:
wrong alarm not needed for the moment. closing but can be opened later.

PR #1130, opened 2021-11-11
Refactor some components and documentation

srknzl:
Change list:

* Comments of components changed and added for clarification and explanation
* `addDistributedObjectListener` is moved to ListenerService since it felt more suitable to me. We are registering and deregistering listeners. `removeDistributedObjectListener` on ProxyManager is removed as there is no need to it.
* ConnectionRegistry's "isActive" was being used only by ConnectionManager and ConnectionManager's component HeartbeatManager. I moved active field from ConnectionRegistry to ConnectionManager for simplicity.
* Connection registry can be get from ConnectionManager now, and due to this I removed the ConnectionRegistry service from service constructors that was taking both ConnectionManager and ConnectionRegistry
* [Parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) syntax refactor


src/LifecycleService.ts
```diff
@@ -62,14 +62,14 @@ export enum LifecycleState {
 const LIFECYCLE_EVENT_NAME = 'lifecycleEvent';
 
 /**
- * Lifecycle service for Hazelcast clients. Allows to determine lifecycle
- * state of the client and shut it down.
+ * Allows to determine if the client is active and emits client state events.
```
yuce, line 65: How about "Allows determining whether the ..."

src/core/LoadBalancer.ts
```diff
@@ -24,13 +24,12 @@ import {
 } from './MembershipListener';
 
 /**
- * {@link LoadBalancer} allows you to send operations to one of a number of endpoints (Members).
+ * A {@link LoadBalancer} allows you to send operations to one of a number of endpoints (Members).
```
yuce, line 27: How about something like: "A LoadBalancer selects one of the endpoints (members) for the next send operation when the operation is not directed to a specific endpoint."

src/core/LoadBalancer.ts
```diff
@@ -24,13 +24,12 @@ import {
...
  * It is up to the implementation to use different load balancing policies.
  *
- * If the client is configured with {@link ClientNetworkConfig.smartRouting},
- * only the operations that are not key based will be routed to the endpoint
- * returned by the LoadBalancer. If it is not, the LoadBalancer will not be used.
- *
+ * If the client is a smart client configured with {@link ClientNetworkConfig.smartRouting},
+ * only the operations that are not key based will be routed via LoadBalancer.
+ * If the client is not a smart client, the LoadBalancer will not be used.
```
yuce, line 32: How about: "The LoadBalancer is used only when the client is in the smart mode and an operation is not directed to a specific endpoint."

yuce (commented):
A few ideas about the documentation, will continue revew.

yuce:
@srknzl Is `ListenerService` considered internal API?

srknzl:
yes it is an internal API

src/LifecycleService.ts
```diff
@@ -62,14 +62,14 @@ export enum LifecycleState {
...
+ * To register lifecycle listeners, use {@link ClientConfig.lifecycleListeners} config.
  */
 export interface LifecycleService {
 
     /**
-     * Returns the active state of the client.
-     * @returns {boolean}
+     * Returns if the client is running or not. This will be true between the time the client is initialized
+     * and it's shutdown process is started.
```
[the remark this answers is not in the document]
srknzl: I guess this is better than both: WDYT?

`Returns true if the client is running. The client is considered to be running between its initialization and the start of its shutdown.`

src/core/LoadBalancer.ts
srknzl: Can a user direct an operation to a specific member? I think "when the operation is not directed to a specific endpoint." is internal detail and user does not need to know it. 

I checked grammarly it says "one of a number of endpoints" can be improved to be "one of many/some/several endpoints".

BTW, this explanation is the same of Java at the moment, maybe we should change that too.

I propose to use:

"A LoadBalancer allows you to send operations to one many endpoints (Members)." changing only "one of a number of" part

src/core/LoadBalancer.ts
srknzl: Again, can a user direct an operation to a specific member? I dont know. So let's hide that detail from the user. Which part do you propose to change by the way? Only the  "If the client is not a smart client," sentence?

src/core/LoadBalancer.ts
yuce: The load balancer selects a member, not send operations to it. I think mentioning that the load balancer is only used in some cases is useful information, but if you'd like to left it out, the doc becomes even simpler, how about: "A LoadBalancer selects one of the endpoints (members) for the next send operation"

src/core/LoadBalancer.ts
yuce: The user obviously cannot direct an operation to a specific member, but if the operation (invocation) is already directed to a key or member (PNCounter) then hte load balancer is not used. I think "only the operations that are not key based will be routed via LoadBalancer" has at least the same implementation detail with "The LoadBalancer is used only when the client is in the smart mode and an operation is not directed to a specific endpoint." (if not more).

I propose to replace lines 30-32 (the lines I've highlighted) with  "The LoadBalancer is used only when the client is in the smart mode and an operation is not directed to a specific endpoint." which I think summarizes those lines.

src/LifecycleService.ts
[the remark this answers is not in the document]
yuce: How about: "The client is considered to be running between initialization and the shutdown process starts."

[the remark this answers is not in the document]
srknzl: Chose 'The client is considered to be running between initialization and start of its shutdown process'. to have two nouns  around "and"

src/core/LoadBalancer.ts
srknzl: A {@link LoadBalancer} selects one of the endpoints(Members) for the next send operation that be directed to many endpoints.

What about this? Did not choose your version since a user cannot direct an operation to an endpoint directly, but they can understand that some operations need to go to a specific member with this sentence.

src/core/LoadBalancer.ts
srknzl: Hmm, ok but I think it is better to explain which operations are directed to a specific endpoint.(key based operations). So the current version gives more info to user. 

I realized that this info is important for user and do not have my opinion I have in the past anymore:  
>  I think "when the operation is not directed to a specific endpoint." is internal detail and user does not need to know it.

srknzl:
Link checker can be ignored. Wrong alarm.

src/core/LoadBalancer.ts
yuce: Key based operations are not the only ones exempt from the load balancer, member based ones are also exempt. Also if an operation has a specific target (connection) then the load balancer will not be used. So load balancer is only used when a random connection is needed.

srknzl: Hmm what about writing that instead

srknzl:
@yuce this is open for a long time, should we finalize this PR?

yuce (approved):
Looks good!

mdumandag (approved):

PR #1132, opened 2021-11-11
Fix flaky MapLockTest

srknzl:
The test fails when the key is tried to be generated before the second partition event including both members comes. Also fails on master on my local

fail:
https://github.com/hazelcast/hazelcast-nodejs-client/runs/4178339025?check_suite_focus=true

### DIFF
--- a/test/integration/backward_compatible/parallel/map/MapLockTest.js
+++ b/test/integration/backward_compatible/parallel/map/MapLockTest.js
@@ -27,10 +27,11 @@ const TestUtil = require('../../../../TestUtil');
 describe('MapLockTest', function () {
     const INVOCATION_TIMEOUT_FOR_TWO = 1000;
     const testFactory = new TestUtil.TestFactory();
+    const DEFAULT_PARTITION_COUNT = 271;
 
-    async function generateKeyOwnedBy(client, member) {
+    function generateKeyOwnedBy(client, member) {
         const partitionService = client.getPartitionService();
-        const MAX_ATTEMPTS = 20;
+        const MAX_ATTEMPTS = 10000;
         let attempt = 0;
         while (attempt++ < MAX_ATTEMPTS) {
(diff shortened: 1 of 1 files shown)


test/integration/backward_compatible/parallel/map/MapLockTest.js
```diff
@@ -19,6 +19,7 @@ const { expect } = require('chai');
 
 const RC = require('../../../RC');
 const TestUtil = require('../../../../TestUtil');
+const { promiseWaitMilliseconds } = require('../../../../TestUtil');
```
srknzl, line 22: Ah I can use TestUtil directly let me fix that

test/integration/backward_compatible/parallel/map/MapLockTest.js
```diff
@@ -85,7 +86,10 @@ describe('MapLockTest', function () {
             }, [member, m]);
         }).then((c) => {
             clientTwo = c;
-            key = generateKeyOwnedBy(client, keyOwner);
+        }).then(() => {
+            return generateKeyOwnedBy(client, keyOwner);
```
mdumandag, line 90: What about, instead of this, you write a new method, waitForNewPartitionTable, where in the method body, you loop from 0 to partition count, and check the owner of the partition, if it is equal to second member's uuid, return, if not wait a bit and try again (or, directly use assertTrueEventually).

My worry with the current implementation is that, we might be unlucky and 20 attempts might not be enough to generate a key owned by a second member

srknzl: ah ok I forgot the randomness here. sure.

srknzl: done

mdumandag (approved):

srknzl:
link checking fails but link is not broken

PR #1136, opened 2021-11-26
Deprecate statistics props and add metrics props API-1181

### LINKED ISSUE #990: [TRACKING ISSUE] Deprecate STATISTICS properties

See https://github.com/hazelcast/hazelcast/pull/19219 for details.

---

The following properties are not used anymore since 4.0

```
  public static final HazelcastProperty STATISTICS_ENABLED = new HazelcastProperty("hazelcast.client.statistics.enabled",
            false);
    /**
     * The period in seconds the statistics run.
     */
    public static final HazelcastProperty STATISTICS_PERIOD_SECONDS = new HazelcastProperty(
            "hazelcast.client.statistics.period.seconds", 3, SECONDS);
```
I traced it back to this pr. It seems that we are checking metricsConfig instead.

This pr:

1. puts @Deprecated to the property documents on ClientProperties class.
2. If user sets statistics, make sure it will set metrics config as well. This is because there
were no warning or documentation about the removal. According to doc, these properties still used.
So we fix the behaviour.
3. If both properties are set, ignore old STATISTICS properties.
4. Adds javadoc to ClientProperty and MetricsConfig to explain the effects of using these properties

fixes https://github.com/hazelcast/hazelcast/issues/18579

Checklist:
- [X] Labels (`Team:`, `Type:`, `Source:`, `Module:`) and Milestone set
- [X] Add `Add to Release Notes` label if changes should be mentioned in release notes or `Not Release Notes content` if changes are not relevant for release notes
- [X] Request reviewers if possible
- [X] New public APIs have `@Nonnull/@Nullable` annotations
- [X] New public APIs have `@since` tags in Javadoc
- [ ] Send backports/forwardports if fix needs to be applied to past/future releases

### LINKED ISSUE #1135: Metrics section not present [API-1181]
https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/src/config/Config.ts
has no Metrics section

This should exist for consistency with other clients.
Eg.https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/config/ClientMetricsConfig.java

Related https://github.com/hazelcast/hazelcast-nodejs-client/issues/990

srknzl:
1. Deprecate statistics props and add metrics props.
2. Update documentation.
3. Validate frequency to be positive.

closes #990 
closes #1135


srknzl:
LInk checker can be ignored safely. Wrong alarm

DOCUMENTATION.md
```diff
@@ -638,8 +638,10 @@ The following is the list of all client properties in alphabetical order.
...
+| hazelcast.client.metrics.enabled                       | false                               | boolean         | If set to `true`, it enables collecting the client metrics and sending them to the cluster. When it is `true` you can monitor the clients that are connected to your Hazelcast cluster, using Hazelcast Management Center.                                                                                                                                                                                          |
+| hazelcast.client.metrics.collection.frequency          | 5                                   | number          | Period in seconds the client metrics are collected and sent to the cluster. Must be positive.                                                                                                                                                                                                                                                                                                                       |
```
yuce, line 644: Should `metrics` be a configuration section instead of properties?

yuce (changes requested):
I think `metrics` should be its own configuration item. Statistic properties were deprecated in the Java client: https://github.com/hazelcast/hazelcast/pull/19219

DOCUMENTATION.md
srknzl: It could be. It may be easier to configure like that. @mdumandag WDYT?

If it will be a separate section, it will be a top level config I guess. Like this:
```js
{
      metrics: {
         enabled: true,
         frequency: 3
      }
}
```

yuce: Making it top level makes sense, it's like that in the Java client and the Go client (there we use a different name).

srknzl: Yeah working on it, almost done thanks for the heads up.

src/config/Config.ts
```diff
@@ -155,6 +155,12 @@ export interface ClientConfig {
      */
     properties?: Properties;
 
+    /**
+     * Metrics config. With this config, you enable collecting the client metrics and sending them to the cluster.
```
yuce, line 159: How about: "Using this configuration, you can enable client metrics collection"?

src/config/Config.ts
```diff
@@ -155,6 +155,12 @@ export interface ClientConfig {
...
+     * After enabling you can monitor the clients that are connected to your Hazelcast cluster,
```
yuce, line 160: How about: "You can monitor connected clients using Hazelcast Management Center once the metrics collection is enabled"?

src/config/ConfigBuilder.ts
```diff
@@ -383,6 +402,26 @@ export class ConfigBuilder {
...
 
+    /**
+     * Statistics properties are deprecated and metrics will be used instead. This method will ensure metrics properties
+     * are set if statistics properties are set.
+     */
+    private static ensureMetricsConfiguredIfStatisticsConfigured(config: ClientConfig) {
+        if (config.hasOwnProperty('properties') &&
+            config.properties.hasOwnProperty('hazelcast.client.statistics.enabled') &&
+            (!config.hasOwnProperty('metrics') || !config.metrics.hasOwnProperty('enabled'))) {
+            throw new RangeError('hazelcast.client.statistics.enabled property is deprecated in version 5.1. ' +
```
yuce, line 413: Why do you throw `RangeError` here? This seems more like an illegal argument error.

src/config/ConfigBuilder.ts
```diff
@@ -421,8 +460,8 @@ export class ConfigBuilder {
         // Throw in case both memberAdded and memberRemoved are invalid.
         if (typeof membershipListener.memberAdded !== 'function' && typeof membershipListener.memberRemoved !== 'function') {
             throw new RangeError(`Invalid membershipListener is given in 'membershipListeners': ${membershipListener}. `
-                                + 'Expected at least one of \'memberAdded\' and \'memberRemoved\' properties to exist and be a'
-                                + ' function.');
+                + 'Expected at least one of \'memberAdded\' and \'memberRemoved\' properties to exist and be a'
```
yuce, line 463: You could use double quotes in order not to escape the single quotes.

src/config/MetricsConfig.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Metrics config. With this config, you enable collecting the client metrics and sending them to the cluster.
```
yuce, line 18: The same suggestion I've made above.

src/config/MetricsConfig.ts
```diff
@@ -0,0 +1,44 @@
...
+ * using Hazelcast Management Center.
+ */
+export interface MetricsConfig {
+
+    /**
+     * Whether metrics collection should be enabled for the client. It's enabled by default.
+     *
+     * If set to `true`, it enables collecting the client metrics and sending them to the cluster.
+     * When it is `true` you can monitor the clients that are connected to your Hazelcast cluster,
+     * using Hazelcast Management Center.
```
yuce, line 29: These lines all tell the same thing. Would it make sense to make it shorter?

src/config/MetricsConfig.ts
```diff
@@ -0,0 +1,44 @@
...
+     */
+    enabled?: boolean;
+
+    /**
+     * Period in seconds the client metrics are collected and sent to the cluster. Must be positive. By default,
```
yuce, line 34: You meant "frequency"?

src/config/ConfigBuilder.ts
srknzl: Here we rethrow IllegalConfigurationError anyway: https://github.com/hazelcast/hazelcast-nodejs-client/blob/c77b9c8094132376ad18defee792b6c166089406/src/config/ConfigBuilder.ts#L58

RangeError can be thrown if user provides something illegal. Think of the whole config object as input. So, we throw RangeError in handleConfig.

src/config/ConfigBuilder.ts
srknzl: We have an eslint rule that forbids double quotes and forces single quotes.

src/config/ConfigBuilder.ts
yuce: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError it seems like the value in question needs to be related to some range.

src/config/ConfigBuilder.ts
yuce: Just out of curiosity, why is that rule is used? Why not double quotes? Just for consistency?

test/integration/backward_compatible/serial/statistics/StatisticsTest.js
```diff
@@ -65,6 +64,7 @@ describe('StatisticsTest (default period)', function () {
     let cluster;
     let client;
     let map;
+    const DEFAULT_STATISTICS_PERIOD = 3;
```
yuce, line 67: Should this one be frequency?

yuce (changes requested):
Just a few minor comments

src/config/ConfigBuilder.ts
srknzl: Yes we can change this error. However, to be consistent with other errors, should I go and check all validation errors and ensure they are not RangeError if there is not a range to talk about? WDYT?

test/integration/backward_compatible/serial/statistics/StatisticsTest.js
srknzl: sure np

src/config/ConfigBuilder.ts
yuce: Since `RangeError` is used for all other configuration errors, it's OK to leave it as is.

yuce (dismissed):
Thanks, looks good!

src/config/ConfigBuilder.ts
mdumandag, line 413: Is this a breaking change? If I have this property set, do I have to change my code?

src/config/MetricsConfig.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Frequency of client metrics collection. Must be positive. By default, metrics are collected every 5 seconds.
+     */
+    collectionFrequencySeconds?: number;
+}
+
+/** @internal */
+export class MetricsConfigImpl implements MetricsConfig {
+    enabled = true;
```
mdumandag, line 36: This is a behavior change. Is this expected in a minor release?

mdumandag (changes requested):

src/config/ConfigBuilder.ts
srknzl: Maybe instead of throwing we should just log a warning? WDYT? @yuce  We should not break existing code. If both configs exist the new one takes precedence

src/config/MetricsConfig.ts
srknzl: As long as it is not a breaking change, it should not be too annoying if we let the users know this. The similar thing should have happened to Java client as well. With one difference though, I think metrics are added in 4.0 so in a major verison. Again @yuce can you share your idea?

Is making wait this change until 6.0 reasonable?

srknzl:
I will change the logic so that: 

When no config is given metrics will be enabled.
When only statistics props is given it will take effect.
When only metrics config is given it will take effect.
When both are given statistics props will take effect.

I will check if it they are given or not by Object.hasOwnProperty() 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

mdumandag:
I would either throw an error on the 4th case or prioritize the metrics config. Why would anyone have two configurations for the same behavior?

srknzl:
> I would either throw an error on the 4th case or prioritize the metrics config. Why would anyone have two configurations for the same behavior?

🤔 Hmm I was thinking about compatibility, but in this case since the user explicitly sets MetricsConfig we don't break existing code. I will throw in this case then.

src/statistics/Statistics.ts
```diff
@@ -38,34 +38,32 @@ type GaugeDescription = {
  * This class is the main entry point for collecting and sending the client
  * statistics to the cluster. If the client statistics feature is enabled,
  * it will be scheduled for periodic statistics collection and sent.
+ *
+ * Uses metricsConfig as configuration. The config from statistics properties
+ * will be used if the metrics counterpart and statistics property is set. This logic
+ * is in ConfigBuilder.
  * @internal
  */
 export class Statistics {
```
srknzl, line 47: Should I rename this to Metrics in this PR?

srknzl:
This PR is almost fully updated. You can review now.

code_samples/client_statistics.js
```diff
@@ -24,9 +24,9 @@ const { Client } = require('hazelcast-client');
                 // Enable near cache with all defaults
                 'nearCachedMap': {}
             },
-            properties: {
-                'hazelcast.client.statistics.enabled': true,
-                'hazelcast.client.statistics.period.seconds': 2
+            metrics: {
```
mdumandag, line 27: Maybe we can now rename the file to `client_metrics`

src/config/ConfigBuilder.ts
```diff
@@ -383,6 +402,41 @@ export class ConfigBuilder {
...
+     * Statistics properties are deprecated and metrics will be used instead. This method will ensure statistics properties
+     * and metrics config are not set at the same time. Also, it will set metricsConfig conditionally in four different cases:
+     *
+     * 1. When no config is given metrics will be enabled.
+     * 2. When only statistics props is given it will take effect.
+     * 3. When only metrics config is given it will take effect.
+     * 4. When both statistics props and metrics config are given throw.
+     */
+    private static handleMetricsAndStatistics(config: ClientConfig, metricsConfig: MetricsConfig): void {
+        if (config.hasOwnProperty('properties') && config.properties.hasOwnProperty('hazelcast.client.statistics.enabled')) {
```
mdumandag, line 415: this will run without an error in a configuration like below
```js
{metrics: {collectionFrequencySeconds: 4}, properties: {'hazelcast.client.statistics.enabled': true}}
```

test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
```diff
@@ -102,7 +102,7 @@ describe('NearCachedMapTest', function () {
                 }
                 await map1.get('key1');
                 await map2.remove('key1');
-                const val = await TestUtil.promiseLater(1000, map1.get.bind(map1, 'key1'));
+                const val = await TestUtil.promiseLater(5000, map1.get.bind(map1, 'key1'));
```
mdumandag, line 105: Didn't we apply some other changes to this part on other PRs?

src/config/ConfigBuilder.ts
srknzl: yeah, but I think they don't create a conflict. I think we should allow that.

test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
srknzl: I am not sure this is left like this, let me rebase

code_samples/client_statistics.js
srknzl: done

src/config/ConfigBuilder.ts
mdumandag: Personally, I don't think this is valid usage. I don't think it should be possible to mix two configuration options. @yuce WDYT?

yuce: How does the Java client behave in this case?

srknzl:
```java
public static void main(String[] args) throws InterruptedException {
        ClientConfig c = new ClientConfig();
        c.getMetricsConfig().setCollectionFrequencySeconds(4);
        c.getProperties().setProperty("hazelcast.client.statistics.enabled", "true");
        HazelcastInstance h = HazelcastClient.newHazelcastClient(c);
        IMap<String, String> map = h.getMap("map");
        map.put("key", "value");
    }
```

results in 

Feb 10, 2022 1:07:03 PM com.hazelcast.client.impl.statistics.ClientStatisticsService
INFO: Client statistics is enabled with period 4 seconds.

with hazelcast commit b27dee57b8dca6e70071656e59b2cd8af100b1cb (9th Feb)

srknzl:
```java
public static void main(String[] args) throws InterruptedException {
        ClientConfig c = new ClientConfig();
        c.getMetricsConfig().setCollectionFrequencySeconds(4);
        c.setProperty("hazelcast.client.statistics.enabled", "false");
        HazelcastInstance h = HazelcastClient.newHazelcastClient(c);
        IMap<String, String> map = h.getMap("map");
        map.put("key", "value");
    }
```

with this code, statistics is disabled.

srknzl: Java does not throw in case both are set: 

```java
public static void main(String[] args) throws InterruptedException {
        ClientConfig c = new ClientConfig();
        c.setProperty("hazelcast.client.statistics.enabled", "false");
        c.getMetricsConfig().setEnabled(true);

        HazelcastInstance h = HazelcastClient.newHazelcastClient(c);
        IMap<String, String> map = h.getMap("map");
        map.put("key", "value");
    }
```

It gived precedence to statistics property and disables statistics. Should we do the same @yuce @mdumandag ?

mdumandag: Hmm, I think it is good to be consistent with Java client then

srknzl: done! @yuce stale review dismissed 😉

yuce (changes requested):

src/config/ConfigBuilder.ts
```diff
@@ -403,37 +403,24 @@ export class ConfigBuilder {
...
-     * 1. When no config is given metrics will be enabled.
-     * 2. When only statistics props is given it will take effect.
-     * 3. When only metrics config is given it will take effect.
-     * 4. When both statistics props and metrics config are given throw.
+     * 1. When no config is given metrics will be enabled. Because this is the default behaviour in metrics.
+     * 2. When only statistics props is given statistics config will take effect. So backward compatibility is kept.
+     * 3. When only metrics config is given metrics config will take effect.
+     * 4. When both statistics props and metrics config are given statistics config will take effect.
+     *
+     * The behaviour is inline with Java's behaviour.
```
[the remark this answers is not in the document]
srknzl: I don't think we need that. this is an internal function.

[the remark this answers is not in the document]
yuce: I thought that was part of the documentation. Is the documentation generated for the internal functions?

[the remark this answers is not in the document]
srknzl: They are excluded from type definition files and documentation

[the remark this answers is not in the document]
yuce: OK. Could you change `Java's` to `Java client's`?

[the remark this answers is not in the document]
srknzl: Sure. done

yuce (approved):
Looks good!

mdumandag (approved):

PR #1080, opened 2021-10-04
Validate methods in config

srknzl:
closes https://github.com/hazelcast/hazelcast-nodejs-client/issues/558

PR #1146, opened 2021-12-09
[API-1018] Introduce security configuration

### LINKED ISSUE #674: Support PasswordCredentials config option [API-1018]
Currently the client only supports `customCredentials` option which maps to `SimpleTokenCredentials` on server-side. We should support `PasswordCredentials`. See `ClientConnectionManager#encodeAuthenticationRequest` to understand how those credentials are send to the cluster.

mdumandag:
This PR introduces a new configuration element called `security`
which enables users to use username-password, token, or custom
authentication.

Formerly, we had a configuration element called `customCredentials`.
We were serializing it using the serialization service, and sending
it as it is.

The problem with that is, the users were not able to use username-password
or token based authentication.

Instead of that element, we introduced a new configuration element and
deprecated it.

The users are still able to provide a custom credentials with this new
element.

closes #674


DOCUMENTATION.md
```diff
@@ -3927,11 +3931,178 @@ const cfg = {
...
+are ``username`` and ``password`` strings.
+
+```js
+const config = {
+    security: {
+        credentials: {
+            type: 'username_password',
+            username: 'admin',
+            password: 'some-strong-password'
+        }
```
yuce, line 3966: Does it make sense to move type up, like `security` would contain one of `username_password`, `token` or `custom`? That would remove a level and makes the configuration a bit easier:
```
const config = {
    security: {
        username_password: {
            username: 'admin',
            password: 'some-strong-password'
        }
```

DOCUMENTATION.md
```diff
@@ -3927,11 +3931,178 @@ const cfg = {
 The client calls the `init()` method with the `properties` configuration option. Then the client calls the `getSSLOptions()`
 method of `SSLOptionsFactory` to create the `options` object.
 
-## 9.2. Credentials
+## 9.2 Authentication
 
-One of the key elements in Hazelcast security is the `Credentials` object, which can be used to carry all security attributes of
-the Hazelcast Node.js client to Hazelcast members. Then, Hazelcast members can authenticate the clients and perform access
-control checks on the client operations using this `Credentials` object.
+By default, the client does not use any authentication method and just uses the configured cluster name to connect to members.
```
yuce, line 3936: We may also add a warning about enabling SSL, otherwise the username/password/token may be easily eavesdropped by third parties. Do you think we should also mention that authentication is an enterprise feature?

code_samples/security/security.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const { Client } = require('hazelcast-client');
```
yuce, line 18: Would it make sense to add the Hazelcast config required for this example, or at least links that provides configuration?

DOCUMENTATION.md
mdumandag: Makes sense, I have simplified the configuration as suggested.

DOCUMENTATION.md
mdumandag: I had a warning about the SSL at the end, but I have moved it to top. Also, put a note about it requiring enterprise edition

code_samples/security/security.js
mdumandag: It is available in the README.md in the same folder

DOCUMENTATION.md
zpqrtbnk: FWIW the .NET configuration file (JSON) supports writing the following "shortcuts", (with the `"authentication"` block being a top-level configuration item):
```json
"authentication": {
    "username-password": {
        "username": "<username>",
        "password": "<password>"
    }
}
```
Or:
```json
"authentication": {
    "token": {
        "encoding": "<none>|<base64",
        "data": "..."
    }
}
```
Or:
```json
"authentication": {
    "kerberos": {
        "spn": "..."
    }
}
```
In addition to the generic one:
```json
"authentication": {
    "credentialsFactory": {
        "typeName": "My.Project.CredentialsFactory, My.Project",
        "args": {
            "arg1": "<arg>",
            "arg2": "123456"
        }
    }
}
```

src/config/ConfigBuilder.ts
```diff
@@ -91,15 +95,85 @@ export class ConfigBuilder {
...
     }
 
+    private static validateSecurityConfiguration(jsonObject: any): void {
+        if ('security' in jsonObject && 'customCredentials' in jsonObject) {
+            throw new RangeError('Ambiguous security configuration is found. ' +
+                'Use one of \'security\' or \'customCredentials\' elements, not both.')
+        }
+    }
+
+    private handleSecurity(jsonObject: any): void {
```
zpqrtbnk, line 116: So if we have the `security` thing, we must have `usernamePassword`, `token` or `custom`... and what is `custom` exactly, is it the thing that enables us to inject any kind of credentials factory?

mdumandag: Yes, exactly, if you have something in the `security` element, it must be one of those three.

`custom` simply means any kind of object: portable, identified etc. where we serialize it as it is and send it to the server. It is similar to credentials factory in that regard

DOCUMENTATION.md
mdumandag: The final form is similar to what we have in C#

zpqrtbnk (approved):

src/security/TokenEncoding.ts
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Supported token encodings.
+ */
+export enum TokenEncoding {
+    ASCII = 'ASCII',
+    BASE64 = 'BASE64',
+}
```
yuce, line 23: Should the values be in lower case?

yuce: Also, should `BASE64` replace `"base64"` constants?

mdumandag: I didn't get the second comment. Could you please explain it again?

yuce: For all examples/tests etc. we use `"base64"`, `"ascii"` etc. Wouldn't it better to use `TokenEncoding.ASCII`, etc. instead of them?

yuce: Also, in the documentation...

mdumandag: Oh, I see. I personally like the string constant usage, as it is much easier, but as you suggested, I have changed the usages to enum members.

yuce (approved):
Thanks, looks good!

src/config/ConfigBuilder.ts
```diff
@@ -91,15 +95,85 @@ export class ConfigBuilder {
...
+                token = tryGetString(value);
+            } else if (key === 'encoding') {
+                encoding = tryGetEnum(TokenEncoding, value);
+            } else {
+                throw new RangeError(`Unexpected token credentials option '${key}' is passed to the Hazelcast Client`)
+            }
+        }
+
+        if (token == null) {
+            return;
```
srknzl, line 171: Should we return here or throw? `token` is a mandatory field in TokenCredentials

src/config/ConfigBuilder.ts
```diff
@@ -91,15 +95,85 @@ export class ConfigBuilder {
             } else if (key === 'customLogger') {
                 this.handleLogger(value);
             } else if (key === 'customCredentials') {
-                this.effectiveConfig.customCredentials = jsonObject;
```
srknzl, line 94: 🤔 Why was it like that? Was it a bug?

src/config/ConfigBuilder.ts
```diff
@@ -91,15 +95,85 @@ export class ConfigBuilder {
...
+            if (key === 'username') {
+                username = tryGetStringOrNull(value);
+            } else if (key === 'password') {
+                password = tryGetStringOrNull(value);
+            } else {
+                throw new RangeError(`Unexpected username password credentials option '${key}' is passed to the Hazelcast Client`)
+            }
+        }
+
+        this.effectiveConfig.security.usernamePassword = new UsernamePasswordCredentialsImpl(username, password);
```
srknzl, line 153: We can have username and/or password `undefined` as a result of this, but it must be a string or null. Isn't this the case? We can fix that by validating both username and password are provided. Another way to fix this is that initializing username and password variables with `null` in `handleUsernamePasswordCredentials`

Actually `CodecUtil.encodeNullable` encodes undefined as `NULL_FRAME` since it checks for `value == null`:

```ts
static encodeNullable<T>(clientMessage: ClientMessage, value: T, encoder: (msg: ClientMessage, val: T) => void): void {
        if (value == null) {
            clientMessage.addFrame(NULL_FRAME.copy());
        } else {
            encoder(clientMessage, value);
        }
    }
```

Still, I think it is better to prevent `undefined` value to avoid confusing some users or hazelcast developers while debugging.

srknzl (commented):
LGTM except two minor issues and a question. @mdumandag Can you have a look?

src/config/ConfigBuilder.ts
mdumandag: Yes, that was mistakenly changed to this in https://github.com/hazelcast/hazelcast-nodejs-client/pull/1080. Since we had no tests for the custom credentials, we failed to catch it there

src/config/ConfigBuilder.ts
mdumandag: Yes, I believe it could be better to throw here

src/config/ConfigBuilder.ts
mdumandag: I think it would be best to initialize them with null :+1:

PR #1147, opened 2021-12-14
[API-1018] Introduce security configuration

mdumandag:
This PR introduces a new configuration element called `security`
which enables users to use username-password, token, or custom
authentication.

Formerly, we had a configuration element called `customCredentials`.
We were serializing it using the serialization service, and sending
it as it is.

The problem with that is, the users were not able to use username-password
or token based authentication.

Instead of that element, we introduced a new configuration element and
deprecated it.

The users are still able to provide a custom credentials with this new
element.

* Simplify the secuirty configuration

This commit gets rid of the `type` and `credentials` and introduces
`usernamePassword`, `token`, and `custom` sub-elements.

* ignore security code samples from the code sample checker

* make username/password text more clear

* use enums instead of string constants

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -135,7 +135,10 @@
     * [9.1.1. TLS/SSL for Hazelcast Members](#911-tlsssl-for-hazelcast-members)
     * [9.1.2. TLS/SSL for Hazelcast Node.js Clients](#912-tlsssl-for-hazelcast-nodejs-clients)
     * [9.1.3. Mutual Authentication](#913-mutual-authentication)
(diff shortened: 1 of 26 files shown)


mdumandag:
This is a backport of #1146. The implementation, documentation, and code samples are identical. The only difference is in the test code, where we don't have parallel/serial tests here, in the 5.0.x branch. The actual test code is the same, the only difference comes from the before/after hooks

yuce (approved):
Looks good!

mdumandag:
Merging with 1 approval as this is almost a clean backport

PR #1155, opened 2021-12-29
Update dependencies

srknzl:
In addition to updating deps, more changes: 

* I think from now on we can upgrade development dependencies who dropped node 10 support. I added a note to CONTRIBUTING.md about using Node 12 and above during development of Node.js client. This should be enough to warn contributors. Since Node 10 is end of life for a long time, not many people should be using it. 

* New eslint version gave errors about [this rule](https://eslint.org/docs/rules/no-loss-of-precision). I fixed related numbers by deleting two digits from each of them. 

* `Disable line ending conversion of git` steps are removed since we don't have linebreak-style rule anymore in eslint rules. Note that I did not removed the step from nightly_runner_maintenance.yml since the rule exists in older versions.

* Add an option to typedoc to preserve old behaviour. See [0.22.0 release notes](https://github.com/TypeStrong/typedoc/releases/tag/v0.22.0).  In the new behaviour all exported things are simply listed and does not look good. They are not grouped in terms of module. (what I mean by grouping:  sql/SqlService, sql/SqlPage etc.)

* Separated code quality and tests workflows to use node 12 and 10 seperately.


CONTRIBUTING.md
```diff
@@ -1,5 +1,7 @@
 # Contributing to Hazelcast
 
+**NOTE**: It is recommended to use Node 12 and above for development.
```
mdumandag, line 3: I am fine with that, but we will still have our tests running with Node.js 10, right?

srknzl:
CI failed since we use Node 10 with eslint 8 which does not support Node 10. Maybe we can separate linting and code quality from tests?

CONTRIBUTING.md
mdumandag: I don't know if it is possible, but one solution might be dividing the dev dependencies into two subgroups: One for testing, and one for the rest (things like linters etc). We can make sure that testing dependencies are Node.js 10 compatible, and be flexible on the rest

srknzl: Yes see my comment below for more.

srknzl:
I checked API documentation and coverage report and they look ok. I think this PR is ready to be merged after CI passes. To be sure, you can also check them.

yuce (approved):
Looks good!

mdumandag (approved):

PR #1157, opened 2021-12-30
Add version list to API documentation

srknzl:
Aims to add a released version list to API documentation rather than redirecting user to the latest version. In the old method users had no way to select an older API docs version other than changing website's URL. 

* Updates .gitignore
* Add author and description
* Add version list
* Delete current.yaml


mdumandag (approved):

srknzl:
Let met delete current.yaml since it is not used anymore.

yuce (approved):

PR #1162, opened 2022-01-05
API-1106 SQL JSON support

### LINKED ISSUE #1126: [TRACKING ISSUE] [HZ-563] JSON type support

See https://github.com/hazelcast/hazelcast/pull/19303 for details.

---

Adds `JSON` as a supported type (available in most contexts, aside from JDBC since that's not part of the SQL spec) as well as few functions from SQL JSON TR. 
Features: 
- `JSON_QUERY(jsonArg VARCHAR|JSON, jsonPath VARCHAR ... <extended syntax> ): JSON` - returns a json-object/array by given JSONPath (implemented according to SQL JSON TR)
- JSON_VALUE(jsonArg VARCHAR|JSON, jsonPath VARCHAR ... <extended syntax>): VARCHAR|<RETURNING TYPE>` - returns a primitive value by given JSONPath (implemented according to SQL JSON TR)
- `JSON_PARSE(jsonString VARCHAR): JSON` - workaround function to mitigate issues with aggressive `CAST(literal AS <type>)` optimizations, analyzer replaces any instance of `CAST(<varcharLiteral> AS JSON)` with a call of this function. 
- `CAST(x AS JSON)` support for `VARCHAR`, columns/literals and dynamic params are supported. 
- `JSON` as a first-class type, with IMap and Kafka Connector support, usable in most contexts, similarly to any other type.

srknzl:
Closes #1126 

Additionally, I changed SQL code samples:

1. Disable logging so that output is seen better
2. Add randomness to map names so that the examples can be run without needing to restart cluster.
3. Format prints to the console in code samples so that it looks better
4. Change root try/catch blocks in code samples to catch promise handler for readability. **most of the changes is from here, don't be scared to review :)** 

Corresponding client protocol PR: https://github.com/hazelcast/hazelcast-client-protocol/pull/409

Added a link to new JSON document in reference manual that only exist on 5.1-beta-1 version. Also, refactored executeStatement() a little bit.


srknzl:
@yuce @mdumandag

Guys can you review this? It is ready, I added the link to json documentation.

srknzl:
@yuce If you want I can ask from Stephan. As I said to you in DM, I have not requested from him for a long time.

code_samples/sql-basic-usage.js
```diff
@@ -19,8 +19,12 @@ const { Client, SqlColumnType } = require('hazelcast-client');
 
 (async () => {
     try {
-        const client = await Client.newHazelcastClient();
-        const mapName = 'myMap';
+        const client = await Client.newHazelcastClient({
+            properties: {
+                'hazelcast.logging.level': 'OFF'
```
yuce, line 24: There is no change in the logging stuff in this PR, but should there be a logging setting instead of the property? Also why is the logging turned off?

code_samples/sql-basic-usage.js
```diff
@@ -19,8 +19,12 @@ const { Client, SqlColumnType } = require('hazelcast-client');
...
+            }
+        });
+        const mapName = 'myMap' + Math.floor(Math.random() * 10000);
```
yuce, line 27: I think it would be better to use a constant map name here, since the user may want to inspect this map after running the example code. Other samples seems to have maps with random names as well.

code_samples/sql_json.js
```diff
@@ -0,0 +1,67 @@
...
+(async () => {
+    try {
+        const client = await Client.newHazelcastClient({
+            properties: {
+                'hazelcast.logging.level': 'OFF'
+            }
+        });
+        const mapName = 'jsonMap' + Math.floor(Math.random() * 10000);
+
+        // To be able to use our map in SQL we need to create mapping for it.
```
yuce, line 29: How about "In order to use the map in SQL a mapping should be created"?

code_samples/sql_json.js
```diff
@@ -0,0 +1,67 @@
...
+        for await (const row of result) {
+            console.log(`${row['__key']}: ${row['this']}`);
+            console.log(`Value is of type ${row['this'].constructor.name}`);
+        }
+
+        await client.shutdown();
+    } catch (err) {
+        console.error('Error occurred:', err);
+        process.exit(1);
+    }
```
yuce, line 66: This is a pretty big try/catch block. How does the exception message look like? Is it worse than the `console.error` output above?

test/integration/backward_compatible/parallel/sql/DataTypeTest.js
```diff
@@ -814,4 +815,59 @@ describe('Data type test', function () {
...
+
+        await client.getSql().execute(createMappingQuery);
+
+        await client.getSql().execute(`INSERT INTO ${mapName} VALUES (2, CAST('{"age": 2}' AS JSON)),` +
+            '(3, ?)',
+        [new HazelcastJsonValue(JSON.stringify({age: 3}))]);
+
+        const result = await TestUtil.getSql(client).execute(
+            `SELECT * FROM ${mapName} WHERE CAST(JSON_VALUE(this, '$.age') AS DOUBLE) > ?` +
+            ' AND CAST(JSON_VALUE(this, \'$.age\') AS DOUBLE) < ? ORDER BY __key ASC',
```
yuce, line 846: Is string concatenation required? Can't you just put the second part of the string in the backtick string above?
```
`SELECT * ...
 AND CAST ...
`
```

yuce (approved):
Looks good, just some minor comments.

code_samples/sql-basic-usage.js
yuce: Does the logging clutter the output that much? What happens if the user forgets to run a cluster?

code_samples/sql-basic-usage.js
yuce: You can use `CREATE OR REPLACE MAPPING` to replace the mapping if it already exists so the user doesn't have to delete the mapping/restart the cluster to run the same example again.

code_samples/sql-basic-usage.js
srknzl: Maybe I should revert this change. It will, a little bit. But if we want to do it we should do it for all code samples I guess. Reverting..

> There is no change in the logging stuff in this PR, but should there be a logging setting instead of the property?

There is a customLogger setting at the root of config object but not logging level. Maybe it is better have it in a more convinient place. However, adding another config should deprecate this one.  @mdumandag  What do you think?

code_samples/sql-basic-usage.js
srknzl: > I think it would be better to use a constant map name here, since the user may want to inspect this map after running the example code. Other samples seems to have maps with random names as well.

It is random since we run all code samples with a cluster in code sample checker. So if the map names are the same they will break with each other. I do not like to do that because of this. We can change code sample checker code to use a new cluster for each code sample. I think we should add this task to jira backlog. 

> You can use CREATE OR REPLACE MAPPING to replace the mapping if it already exists so the user doesn't have to delete the mapping/restart the cluster to run the same example again.


Since the map name is random it won't be needed but it won't harm so doing it.

code_samples/sql_json.js
srknzl: Applying..

code_samples/sql_json.js
srknzl: > Is it worse than the console.error output above?

What do you mean? 

> This is a pretty big try/catch block. 

I remember we put this try catch blocks to make sure that if an error occurs our code sample checker will shut down the process with an error code so that the checker will fail. Alternative approach can be to add a .catch() call to the big async function call and do the same there. It will look better. Applying..

test/integration/backward_compatible/parallel/sql/DataTypeTest.js
srknzl: it will include the spaces and newline characters after the `select * ...` part. It won't work most probably

code_samples/sql_json.js
srknzl: I will do the same thing to all code samples as well.

code_samples/sql-basic-usage.js
srknzl: Alternatively we can log map names to console after creating its name.

test/integration/backward_compatible/parallel/sql/DataTypeTest.js
srknzl: Ok changed that part

DOCUMENTATION.md
```diff
@@ -2827,6 +2827,43 @@ for await (const row of sqlResult) {
...
 ```
 
+The "json-flat" value format means that top level fields of the json object is treated as separate columns. There is also "json"
+value format which we cover in the next section.
+
+### 8.7.7. JSON Data Type
+
+JSON is a first-class SQL type with IMap and Kafka Connector support, usable in most contexts, similarly to any other type.
+
+In order to send a JSON parameter use a `HazelcastJsonValue` object:
```
mdumandag, line 2837: Can't this also be a regular JS object? IMHO, if JS objects work fine, we should use it in code snippets

DOCUMENTATION.md
```diff
@@ -2827,6 +2827,43 @@ for await (const row of sqlResult) {
...
+await client.getSql().execute('INSERT INTO test VALUES (2, ?)', [new HazelcastJsonValue(JSON.stringify({age: 3}))]);
+```
+
+The corresponding result type for JSON type is `HazelcastJsonValue` as well:
+
+```javascript
+const result = await client.getSql().execute('SELECT * FROM test');
+
+const rowMetadata = result.rowMetadata;
+rowMetadata.getColumn(rowMetadata.findColumn('this')).type.should.be.eq(SqlColumnType.JSON);
```
mdumandag, line 2849: Is this a test? `type.should.be.eq` will not be available for the most of the users

DOCUMENTATION.md
```diff
@@ -2827,6 +2827,43 @@ for await (const row of sqlResult) {
...
+
+for await (const row of result) {
+    console.log(row);
+}
+```
+
+Supported operations:
+
+1. `JSON_QUERY(jsonArg VARCHAR|JSON, jsonPath VARCHAR ... <extended syntax> ): returns JSON`: returns a json-object/array by
```
mdumandag, line 2858:
```suggestion
1. `JSON_QUERY(jsonArg VARCHAR|JSON, jsonPath VARCHAR ... <extended syntax> ): returns JSON`: returns a JSON-object/array by
```

DOCUMENTATION.md
```diff
@@ -2827,6 +2827,43 @@ for await (const row of sqlResult) {
...
+the given JSON path.
+2. `JSON_VALUE(jsonArg VARCHAR|JSON, jsonPath VARCHAR ... ): returns VARCHAR`: returns a primitive value as varchar by the
+given JSON path.
+3. `CAST(x AS JSON)` support for VARCHAR, columns/literals and dynamic params are supported.
+
+For more information about working with JSON using SQL see
+[Working with JSON](https://docs.hazelcast.com/hazelcast/5.1-beta-1/sql/working-with-json) in Hazelcast reference manual.
```
mdumandag, line 2865: We need to remember to update this before releasing 5.1

code_samples/bigdecimal_sample.js
```diff
@@ -43,20 +43,19 @@ function portableFactory(classId) {
...
-        };
+    };
 
-        const client = await Client.newHazelcastClient(cfg);
-        const mapName = 'decimalMap';
-        const map = await client.getMap(mapName);
-        // To be able to use our map in SQL we need to create mapping for it.
-        const createMappingQuery = `
+    const client = await Client.newHazelcastClient(cfg);
+    const mapName = 'decimalMap';
```
mdumandag, line 55: Let's either use a random map name or use CREATE OR REPLACE MAPPING or something similar to make the consecutive runs easier

code_samples/datetime_classes_sample.js
```diff
@@ -52,21 +52,20 @@ function portableFactory(classId) {
...
 
-        const client = await Client.newHazelcastClient(cfg);
+    const client = await Client.newHazelcastClient(cfg);
 
-        const mapName = 'timestampWithTimezoneMap';
-        const map = await client.getMap(mapName);
-        // To be able to use our map in SQL we need to create mapping for it.
-        const createMappingQuery = `
+    const mapName = 'timestampWithTimezoneMap';
+    const map = await client.getMap(mapName);
```
mdumandag, line 66: Same as above

code_samples/sql_json.js
```diff
@@ -0,0 +1,61 @@
...
+            TYPE IMAP
+            OPTIONS (
+                'keyFormat' = 'bigint',
+                'valueFormat' = 'json'
+            )
+        `;
+    await client.getSql().execute(createMappingQuery);
+
+    await client.getSql().execute(`INSERT INTO ${mapName} VALUES (1, ?)`,
+        [new HazelcastJsonValue(JSON.stringify({age: 1}))]
```
mdumandag, line 39: Same as documentation, it would be far easier to use JS objects here for the user

code_samples/sql_json.js
```diff
@@ -0,0 +1,61 @@
...
+
+    console.log('Columns:');
+    for (const column of columns) {
+        console.log(`${column.name}: ${SqlColumnType[column.type]}`);
+    }
+
+    console.log('\nRows from query:');
+    for await (const row of result) {
+        console.log(`${row['__key']}: ${row['this']}`);
+        console.log(`Value is of type ${row['this'].constructor.name}`);
```
mdumandag, line 54: What is `.constrcutor.name`? It seems a bit cryptic to me

src/sql/SqlColumnMetadata.ts
```diff
@@ -64,7 +64,10 @@ export enum SqlColumnType {
      * The type of the generic SQL `NULL` literal.
      * The only valid value of `NULL` type is `null`.
      */
-    NULL
+    NULL,
+
+    /** JSON type, represented by {@link HazelcastJsonValue} */
```
mdumandag, line 69: Also JS objects

test/integration/backward_compatible/parallel/sql/DataTypeTest.js
```diff
@@ -814,4 +815,59 @@ describe('Data type test', function () {
...
+        rows.length.should.be.eq(expectedKeys.length);
+
+        for (let i = 0; i < rows.length; i++) {
+            const jsonValue = rows[i]['this'];
+            jsonValue.should.be.instanceof(HazelcastJsonValue);
+            const obj = JSON.parse(jsonValue.toString());
+            obj.age.should.be.eq(expectedBaseValues.age + i);
+            rows[i]['__key'].should.be.eq(expectedKeys[i]);
+        }
+    });
```
mdumandag, line 872: Can we also add a test with JS objects?

DOCUMENTATION.md
srknzl: If a user configures a global serializer I think using object won't work. This way it is safer.

DOCUMENTATION.md
srknzl: Yeah when hazelcast 5.1 is released we should update this. Let me create a note for myself.

code_samples/bigdecimal_sample.js
srknzl: I think that is done in https://github.com/hazelcast/hazelcast-nodejs-client/pull/1162

code_samples/sql_json.js
srknzl: Every object has a constructor function that created it. For literal 1 it is Number, for '' it is String etc. And for class instances it is class constructor. 

`.name` gives name of that function.

test/integration/backward_compatible/parallel/sql/DataTypeTest.js
srknzl: done. the return type does not change, I added passing js object as parameter.

src/sql/SqlColumnMetadata.ts
srknzl: done

DOCUMENTATION.md
srknzl: done

code_samples/bigdecimal_sample.js
srknzl: Ah sorry, I meant to say please see this comment: https://github.com/hazelcast/hazelcast-nodejs-client/pull/1162#discussion_r791716878

DOCUMENTATION.md
mdumandag: The ones who configure a global serializer is probably around %0.1 of our users :) 

I would mention JS objects for the vast majority

code_samples/bigdecimal_sample.js
mdumandag: OK, I can accept constant map names, but then why don't we use CREATE OR REPLACE MAPPING so that the user can run the code multiple times, without an error?

code_samples/sql_json.js
mdumandag: Would the `typeof` operator work the same way?

code_samples/bigdecimal_sample.js
srknzl: Hmm I think I should align map names in all code samples. I will try to make all of them constant and use create or replace mapping. It is a little bit annoying to ensure they are unique, but being constant will be good for users to inspect the map afterwards like Yuce said. Maybe I should use the file name as part of the map name.

srknzl: Ah sorry, we are starting a new member for each test already. 🤦🏻‍♂️  Going to make all of them constant and add CREATE OR REPLACE MAPPING. It is enough to avoid clash in a single file. https://github.com/hazelcast/hazelcast-nodejs-client/blob/0af1d2037ca253cdb82584de1be22de8927b960b/scripts/code-sample-checker.js#L53

code_samples/sql_json.js
srknzl: nope. It would return 'object'. Constructor.name returns 'HazelcastJsonValue'

code_samples/sql_json.js
srknzl: ok changed with a note

mdumandag (commented):
Can you also resolve the merge conflict?

test/integration/backward_compatible/parallel/sql/DataTypeTest.js
```diff
@@ -814,4 +815,62 @@ describe('Data type test', function () {
...
     });
+    it('should be able to decode/serialize JSON', async function () {
+        const inputs = [new HazelcastJsonValue(JSON.stringify({age: 3})), {age: 3}];
+        for (const input of inputs) {
+            await basicSetup(this);
+            const SqlColumnType = TestUtil.getSqlColumnType();
+
+            // JSON support is added in 5.1.
+            TestUtil.markClientVersionAtLeast(this, '5.1');
+            TestUtil.markServerVersionAtLeast(this, client, '5.1');
```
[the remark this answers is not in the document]
srknzl: By top do you mean before or beforeEach? This JSON datatype needs additional skip logic since it is added later than others added. If you mean changing these two lines' orders, no problem I did it.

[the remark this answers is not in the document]
mdumandag: I meant line 819, just after the `it` block, so no code is executed if the checks do not pass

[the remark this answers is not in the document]
srknzl: We cannot run it before `basicTests()` since the client would be uninitialized. Putting it right below it.

mdumandag (approved):

yuce (approved):
Looks good!

PR #797, opened 2021-02-06
[BACKLOG] Cleanup related to service initialization #731

srknzl:
Refactor to pass client's properties to services/managers' constructors in HazelcastClient constructor instead of passing the client itself.

fixes #731 and #807

## Update:

 All changes as a list

* Service/Manager classes get services instead of client object when they are constructed.
* Client shutdown logic is moved from lifecycle service to client itself.
* Methods and attributes in connection manager that are used in other services are moved from connection manager to a seperate class called ConnectionRegistry. 
* Some attributes made readonly
* tests refactored due to service construction change
* Added connection registry test which includes unit tests for `getActiveConnection` and `checkIfInvocationAllowed` methods.
* Rename ClientConnection to Connection
* Rename ClientConnectionManager to ConnectionManager

PR #1164, opened 2022-01-06
Fix CPSessionManagerTest

### LINKED ISSUE #1022: CPSessionManagerTest fail on windows
image

failed on gh actions of pr

#1000

srknzl:
We forgot to change this test in #797. Invocation service is used in heartbeatTask in CPSessionManager and leads to an error.  I fixed the initialization of CPSessionManager. The test was flaky, and the situation this test passes is probably when it passes before the first heartbeat task runs.

closes #1022

### DIFF
--- a/test/unit/proxy/cpsubsystem/CPSessionManagerTest.js
+++ b/test/unit/proxy/cpsubsystem/CPSessionManagerTest.js
@@ -22,7 +22,6 @@ const sinon = require('sinon');
 const sandbox = sinon.createSandbox();
 const Long = require('long');
 const {
-    Client,
     IllegalStateError,
     SessionExpiredError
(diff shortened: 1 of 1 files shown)


yuce (approved):

mdumandag (approved):

mdumandag:
Can we catch such errors, if we were to use typescript in the tests as well? Maybe it would be good idea to invest some time in this area after 5.1

srknzl:
Good idea. We could use https://www.npmjs.com/package/ts-mocha to run our tests in the future if we switch our tests to typescript. This library only transpiles does not do type checking(still IDEs will show errors). To do also type checking, we can compile test files in `npm run compile` command.

PR #1169, opened 2022-01-12
Change server version calculation to allow more flexible versions

srknzl:
Tests failed due to version parser cannot parse the new version 5.1-beta-1-snapshot-1. This PR fixes that to support more versions. Python client's approach is used.

https://github.com/hazelcast/client-compatibility-suites/runs/4782247432?check_suite_focus=true

### DIFF
--- a/src/BuildInfo.ts
+++ b/src/BuildInfo.ts
@@ -24,25 +24,32 @@ export class BuildInfo {
     public static readonly UNKNOWN_VERSION_ID = -1;
     private static readonly MAJOR_VERSION_MULTIPLIER = 10000;
     private static readonly MINOR_VERSION_MULTIPLIER = 100;
-    private static readonly PATTERN = /^([\d]+)\.([\d]+)(?:\.([\d]+))?(-[\w]+)?(-SNAPSHOT)?(-BETA-.)?$/;
 
     public static calculateServerVersionFromString(versionString: string): number {
         if (versionString == null) {
             return BuildInfo.UNKNOWN_VERSION_ID;
         }
-        const info = BuildInfo.PATTERN.exec(versionString);
-        if (info === null) {
+        const mainParts = versionString.split('-');
+        const tokens = mainParts[0].split('.');
+
+        if (tokens.length < 2) {
             return BuildInfo.UNKNOWN_VERSION_ID;
         }
-        const major = Number.parseInt(info[1]);
-        const minor = Number.parseInt(info[2]);
+
+        const major = +tokens[0];
+        const minor = +tokens[1];
         let patch: number;
-        if (info[3] === undefined) {
+        if (tokens.length === 2) {
             patch = 0;
         } else {
-            patch = Number.parseInt(info[3]);
+            patch = +tokens[2];
         }
-        return this.calculateServerVersion(major, minor, patch);
+        const version = this.calculateServerVersion(major, minor, patch);
+        if (isNaN(version)) {
+            return BuildInfo.UNKNOWN_VERSION_ID;
+        } else {
+            return version;
+        }
     }
 
     public static calculateServerVersion(major: number, minor: number, patch: number): number {
--- a/src/BuildInfo.ts
+++ b/src/BuildInfo.ts
@@ -38,18 +38,12 @@ export class BuildInfo {
 
         const major = +tokens[0];
         const minor = +tokens[1];
-        let patch: number;
-        if (tokens.length === 2) {
-            patch = 0;
-        } else {
-            patch = +tokens[2];
-        }
+        const patch = (tokens.length === 2) ? 0 : +tokens[2];
+
         const version = this.calculateServerVersion(major, minor, patch);
-        if (isNaN(version)) {
-            return BuildInfo.UNKNOWN_VERSION_ID;
-        } else {
-            return version;
-        }
+
+        // version is NaN when one of major, minor and patch is not a number.
+        return isNaN(version) ? BuildInfo.UNKNOWN_VERSION_ID : version;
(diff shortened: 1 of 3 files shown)


src/BuildInfo.ts
```diff
@@ -24,25 +24,32 @@ export class BuildInfo {
...
-        const info = BuildInfo.PATTERN.exec(versionString);
-        if (info === null) {
+        const mainParts = versionString.split('-');
+        const tokens = mainParts[0].split('.');
+
+        if (tokens.length < 2) {
             return BuildInfo.UNKNOWN_VERSION_ID;
         }
-        const major = Number.parseInt(info[1]);
-        const minor = Number.parseInt(info[2]);
```
srknzl, line 38: Number.parseInt is different from + operator. Number.parseInt parses the first number and returns it. so Number.parseInt("3,7") === 3. We don't want this

src/BuildInfo.ts
```diff
@@ -24,25 +24,32 @@ export class BuildInfo {
...
         let patch: number;
-        if (info[3] === undefined) {
+        if (tokens.length === 2) {
             patch = 0;
         } else {
-            patch = Number.parseInt(info[3]);
+            patch = +tokens[2];
+        }
+        const version = this.calculateServerVersion(major, minor, patch);
+        if (isNaN(version)) {
```
mdumandag, line 48: Just wondering, how can this be NaN?

srknzl: If one of major, minor or patch is NaN it will be NaN, because arithmetic operations with NaN results in NaN.

mdumandag: Oh, OK. I thought `+"x"` or something like that would throw but it returns `NaN` instead :+1:

srknzl: Exactly.

mdumandag (commented):
BTW, I am OK with this change, but this file will not be used in the compatibility tests as it is a source file related to client. We should probably duplicate this logic in the tests folder and use it, instead of this

test/TestUtil.js
```diff
@@ -540,3 +540,35 @@ exports.TestFactory = class TestFactory {
...
+ */
+exports.calculateServerVersionFromString = (versionString) => {
+    if (versionString == null) {
+        return BuildInfo.UNKNOWN_VERSION_ID;
+    }
+    const mainParts = versionString.split('-');
+    const tokens = mainParts[0].split('.');
+
+    if (tokens.length < 2) {
+        return BuildInfo.UNKNOWN_VERSION_ID;
```
srknzl, line 556: Q: Should we copy these constants as well to this function or use them from BuildInfo?

mdumandag: I think it should be OK, as these constants are defined on every release but I wouldn't mind copying them either

mdumandag (approved):

src/BuildInfo.ts
yuce: How about `parseFloat`?

srknzl: It has the same logic as parseNumber:

> Number.parseFloat("1.1sasd")
1.1

src/BuildInfo.ts
```diff
@@ -24,25 +24,32 @@ export class BuildInfo {
...
+        const major = +tokens[0];
+        const minor = +tokens[1];
         let patch: number;
-        if (info[3] === undefined) {
+        if (tokens.length === 2) {
             patch = 0;
         } else {
-            patch = Number.parseInt(info[3]);
+            patch = +tokens[2];
+        }
```
yuce, line 46: These lines could be simplified as:
```
let patch = (tokens.length === 2)? 0 : +tokens[2];
```

src/BuildInfo.ts
```diff
@@ -24,25 +24,32 @@ export class BuildInfo {
...
+        const version = this.calculateServerVersion(major, minor, patch);
+        if (isNaN(version)) {
+            return BuildInfo.UNKNOWN_VERSION_ID;
+        } else {
+            return version;
         }
```
yuce: These lines can also be simplified using the ternary operator.

test/TestUtil.js
```diff
@@ -540,3 +540,35 @@ exports.TestFactory = class TestFactory {
...
+    }
+
+    const major = +tokens[0];
+    const minor = +tokens[1];
+    let patch;
+    if (tokens.length === 2) {
+        patch = 0;
+    } else {
+        patch = +tokens[2];
+    }
```
yuce: These lines can also be simplified using the ternary operator.

test/TestUtil.js
```diff
@@ -540,3 +540,35 @@ exports.TestFactory = class TestFactory {
...
+    const version = BuildInfo.MAJOR_VERSION_MULTIPLIER * major + BuildInfo.MINOR_VERSION_MULTIPLIER * minor + patch;
+    if (isNaN(version)) {
+        return BuildInfo.UNKNOWN_VERSION_ID;
+    } else {
+        return version;
+    }
```
yuce: Ternary operator may be used.

src/BuildInfo.ts
srknzl: done in 672546d6

also changed return part in the same way and added a comment

yuce (approved):
Looks good, a few minor comments.

src/BuildInfo.ts
srknzl: also changed all of these in 672546d66a6eefec63feb7d0583f4fef40107d31

srknzl:
thanks for review guys!

PR #1177, opened 2022-01-24
Fix NearCachedMapTest

### LINKED ISSUE #1176: NearCachedMapTest test fail [API-1161]
failed with node 10.24.1, on windows, client commit 8ad6e6b846e2c1b79e1d46534386562cda422d7b

```
1) NearCachedMapTest
       invalidate on change=true
         get returns null if the entry was removed by another client:

      AssertionError: expected 1 to equal 0
      + expected - actual

      -1
      +0
      
      at expectStats (test\integration\backward_compatible\parallel\map\NearCachedMapTest.js:71:43)
      at Context.<anonymous> (test\integration\backward_compatible\parallel\map\NearCachedMapTest.js:106:17)
```

can easily reproduce on another setup such as node 16 and macos.

srknzl:
This PR fixes a near cached map test race condition. My understanding is the following. If the delayed map1.get() call finishes before invalidation, the test fails with 1 hit count and 1 miss count where it should be 0 hit count and 2 miss counts. Increased timeout to be 5 seconds.

fixes #1176

### DIFF
--- a/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
+++ b/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
@@ -96,16 +96,15 @@ describe('NearCachedMapTest', function () {
                 expectStats(map1, 0, 2, 1);
             });
 
-            it('get returns null eventually after the entry is removed by another client', async function () {
+            it('get returns null if the entry was removed by another client', async function () {
                 if (!invalidateOnChange) {
                     this.skip();
                 }
                 await map1.get('key1');
                 await map2.remove('key1');
-                await TestUtil.assertTrueEventually(async () => {
-                    const val = await map1.get('key1');
-                    expect(val).to.be.null;
-                }, 1000);
+                const val = await TestUtil.promiseLater(5000, map1.get.bind(map1, 'key1'));
+                expectStats(map1, 0, 2, 1);
+                expect(val).to.be.null;
             });
 
             it('clear clears nearcache', async function () {
--- a/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
+++ b/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
@@ -96,15 +96,17 @@ describe('NearCachedMapTest', function () {
                 expectStats(map1, 0, 2, 1);
             });
 
-            it('get returns null if the entry was removed by another client', async function () {
+            it('get returns null eventually after the entry is removed by another client', async function () {
                 if (!invalidateOnChange) {
                     this.skip();
                 }
                 await map1.get('key1');
                 await map2.remove('key1');
-                const val = await TestUtil.promiseLater(1000, map1.get.bind(map1, 'key1'));
-                expectStats(map1, 0, 2, 1);
-                expect(val).to.be.null;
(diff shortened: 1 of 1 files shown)


test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
```diff
@@ -96,15 +96,17 @@ describe('NearCachedMapTest', function () {
...
+            it('get returns null eventually after the entry is removed by another client', async function () {
                 if (!invalidateOnChange) {
                     this.skip();
                 }
                 await map1.get('key1');
                 await map2.remove('key1');
-                const val = await TestUtil.promiseLater(1000, map1.get.bind(map1, 'key1'));
-                expectStats(map1, 0, 2, 1);
-                expect(val).to.be.null;
+                const val = await map1.get('key1');
```
mdumandag, line 105: Can't this result in a hit and return non-null from the client's near-cache, if the invalidation is not yet received by the client?

srknzl: 🤔 yes it can, my code feels wrong. The get call was in assertTrueEventually but I don't know if it will have side affects that will make test fail.

test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
```diff
@@ -102,7 +102,7 @@ describe('NearCachedMapTest', function () {
                 }
                 await map1.get('key1');
                 await map2.remove('key1');
-                const val = await TestUtil.promiseLater(1000, map1.get.bind(map1, 'key1'));
+                const val = await TestUtil.promiseLater(5000, map1.get.bind(map1, 'key1'));
                 expectStats(map1, 0, 2, 1);
```
zpqrtbnk, line 106: just to be sure I understand correctly: `map1` puts a value into the map, `map2` removes that value, and it's therefore expected that `map1` NearCache gets notified and stops returning the value. Method `promiseLater` is going to wait for the specified time, *then* get the value. If the time value is too small, the test fail because you *still* get the value, most probably because the value has not been removed *yet*. Correct?

We have the same issues in .NET and at times some tests fail because the GitHub runner seems to pause for 5-10 seconds. I would *really* like to use mock servers for this, so we can control timeouts without depending on actual servers.

mdumandag (approved):

test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
srknzl: > just to be sure I understand correctly: map1 puts a value into the map, map2 removes that value, and it's therefore expected that map1 NearCache gets notified and stops returning the value. Method promiseLater is going to wait for the specified time, then get the value. If the time value is too small, the test fail because you still get the value, most probably because the value has not been removed yet. Correct?

exactly.

> We have the same issues in .NET and at times some tests fail because the GitHub runner seems to pause for 5-10 seconds. I would really like to use mock servers for this, so we can control timeouts without depending on actual servers.

I can't imagine how that mocking would work but I like the idea of solving these failures.

zpqrtbnk (approved):
all good to me

PR #1179, opened 2022-01-26
Fix nearcachedmaptest 2nd attempt

srknzl:
Second attempt to fix near cached map test. The test failed again on master during nigthly tests with the same exception. 

In order to fix it I removed expectStats call as Metin thinks that it's not needed. Also now I assert that the client does not go to cluster after some point and uses the near cache. I achieve that by adding a spy(a thing that counts call count to a function) to `MapProxy.getInternal`. `NearCachedMapProxy` calls `MapProxy.getInternal` via `super.getInternal()` only when near cache doesnt not have the entry.

Along with the test I added error stack trace logging to test errors originated from assertTrueEventually since it will be easier to reason test failures.

Example output with my change:

```
Error: Rejected due to timeout of 3000ms. Errors occurred in order: 

AssertionError: expected 1 to equal 0
    at /var/git/forked/hazelcast-nodejs-client/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js:120:60
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
AssertionError: expected null not to be null
    at /var/git/forked/hazelcast-nodejs-client/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js:121:42

    at Timeout._onTimeout (test/TestUtil.js:88:20)
    at listOnTimeout (internal/timers.js:557:17)
    at processTimers (internal/timers.js:500:7)
```

This output is only seen in the test results not in the output, so I don't expect the test duration will increase.

fixes #1177 

EDIT: Since public API usage will make the test more resilient to back compat test failure hit count increase is asserted now.

### DIFF
--- a/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
+++ b/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
@@ -96,16 +96,15 @@ describe('NearCachedMapTest', function () {
                 expectStats(map1, 0, 2, 1);
             });
 
-            it('get returns null eventually after the entry is removed by another client', async function () {
+            it('get returns null if the entry was removed by another client', async function () {
                 if (!invalidateOnChange) {
                     this.skip();
                 }
                 await map1.get('key1');
                 await map2.remove('key1');
-                await TestUtil.assertTrueEventually(async () => {
-                    const val = await map1.get('key1');
-                    expect(val).to.be.null;
-                }, 1000);
+                const val = await TestUtil.promiseLater(5000, map1.get.bind(map1, 'key1'));
+                expectStats(map1, 0, 2, 1);
+                expect(val).to.be.null;
             });
 
             it('clear clears nearcache', async function () {
--- a/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
+++ b/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
@@ -105,13 +105,10 @@ describe('NearCachedMapTest', function () {
                 await map1.get('key1');
                 await map2.remove('key1');
 
-                let previousHitCount;
-
                 await assertTrueEventually(async () => {
-                    previousHitCount = getNearCacheStats(map1).hitCount;
                     const val = await map1.get('key1');
-                    // Client should eventually use near cache
-                    expect(getNearCacheStats(map1).hitCount).to.be.greaterThan(previousHitCount);
+                    // Client should eventually fetch invalidated entry from cluster due to invalidation
+                    expect(getNearCacheStats(map1).missCount).to.be.greaterThan(1);
                     expect(val).to.be.null;
                 }, 1000);
             });
--- a/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
+++ b/test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
@@ -110,10 +110,10 @@ describe('NearCachedMapTest', function () {
                 await assertTrueEventually(async () => {
                     previousHitCount = getNearCacheStats(map1).hitCount;
                     const val = await map1.get('key1');
-                    expect(getNearCacheStats(map1).hitCount).to.be.greaterThan(previousHitCount);
                     // Client should eventually use near cache
+                    expect(getNearCacheStats(map1).hitCount).to.be.greaterThan(previousHitCount);
                     expect(val).to.be.null;
-                }, 1000, 4000);
+                }, 1000);
             });
 
             it('clear clears nearcache', async function () {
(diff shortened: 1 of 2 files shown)


zpqrtbnk (approved):
Oh my. These "true eventually" things look so dirty - and yes, we have them too in .NET. No better solution right now. Go with it :)

test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
```diff
@@ -102,9 +107,19 @@ describe('NearCachedMapTest', function () {
...
+
+                let previousCallCount = mapProxyGetSpy.callCount;
+
+                await assertTrueEventually(async () => {
+                    previousCallCount = mapProxyGetSpy.callCount;
+                    const val = await map1.get('key1');
+                    // Client should eventually use near cache
+                    expect(mapProxyGetSpy.callCount).to.be.eq(previousCallCount);
+                    expect(val).to.be.null;
+                }, 1000);
```
mdumandag, line 122: I think spying internals in a backward-compatible test should be avoided. What about a version like below?
```js
                await map1.get('key1');
                await map2.remove('key1');

                await assertTrueEventually(async () => {
                    const val = await map1.get('key1');
                    getNearCacheMissCount(map1).to.be.greaterThan(0); // Or whatever the syntax is
                    expect(val).to.be.null;
                }, 1000);
```

srknzl:
```
                console.log(getNearCacheStats(map1).missCount);
                await map1.get('key1');
                console.log(getNearCacheStats(map1).missCount);
                await map2.remove('key1');

                console.log(getNearCacheStats(map1).missCount);

                await assertTrueEventually(async () => {
                    const val = await map1.get('key1');
                    // Client should eventually use near cache
                    console.log(getNearCacheStats(map1).missCount);
                    expect(val).not.to.be.null;
                }, 1000);
```

when I do this it logs:

```
0
1
1
and then
2
2
2
2
2
```

So first get() increases miss count by 1 so it is always more than 0. Also you were saying we should make sure that the client does not send invocation to cluster anymore and use near cache. So we should assert that hitCount increases. Am I right?

srknzl: New version of the test is in expect

mdumandag: Then, I think it would be ok to make it `expect(getNearCacheStats(map1).hitCount).to.be.greaterThan(1);`. I just want to see that, we can't find that key in the near cache, because it is invalidated after some time.

Your version is also OK

test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
```diff
@@ -100,11 +101,19 @@ describe('NearCachedMapTest', function () {
...
-                const val = await TestUtil.promiseLater(5000, map1.get.bind(map1, 'key1'));
-                expectStats(map1, 0, 2, 1);
-                expect(val).to.be.null;
+
+                let previousHitCount;
+
+                await assertTrueEventually(async () => {
+                    previousHitCount = getNearCacheStats(map1).hitCount;
+                    const val = await map1.get('key1');
+                    // Client should eventually use near cache
```
mdumandag, line 113: I think this comment is not correct. Client should eventually ask the cluster for this key, after seeing that its local copy is invalidated

srknzl: Yes and then it should keep asking from near cache. What about `// Client should eventually fetch invalidated entry from cluster and then start using its near cache `

srknzl: This should "eventually use near cache" refers to that

mdumandag: `// Client should eventually fetch invalidated entry from cluster `
I think this part is enough. In this test, I only care about the invalidation

srknzl: 🤔  Then we should assert that missCount increases in the assertTrueEventually? I am trying to align comment and code

like this:

```js
await assertTrueEventually(async () => {
                    const previousHitCount = getNearCacheStats(map1).missCount;
                    const val = await map1.get('key1');
                    // Client should eventually fetch invalidated entry from cluster due to invalidation
                    expect(getNearCacheStats(map1).missCount).to.be.greaterThan(previousHitCount);
                    expect(val).to.be.null;
                }, 1000);
```

mdumandag: Yes, the code seems correct to me. Or simply
```js
await assertTrueEventually(async () => {
                    const val = await map1.get('key1');
                    // Client should eventually fetch invalidated entry from cluster due to invalidation
                    expect(getNearCacheStats(map1).missCount).to.be.greaterThan(1);
                    expect(val).to.be.null;
                }, 1000);
```

srknzl: ok applying. I was behaving safe, but I think there is no way the missCount to be greater than 1 before we start executing assertTrueEventually.

mdumandag (approved):

PR #1180, opened 2022-01-26
Fix LocalDateTime.fromDate

srknzl:
LocalDateTime.fromDate using Date's month wrongly. This fixes that. It was throwing right away. I think we should backport this and release soon. Also adds fromDate to LocalDate and LocalTime. 

Fixes https://hazelcast.atlassian.net/jira/software/projects/API/boards/40?assignee=604cdbcd2b3f9a006a710ff9&selectedIssue=API-1178

EDIT: 

* Date.UTC is deleted since I realized that they are not necessary. Tests pass without it. new Date() returns the local time of the system which is like LocalDateTime.now() in java.

* The issue happens due to the fact that we don't convert month field of JS Date into our month field. JS Date is 0-11 based and we use 1-12 in LocalDate. Therefore, we should have added 1 to the month field in fromDate constructor. If a user gives a date in january fromDate throws validation error.

### DIFF
--- a/test/unit/core/DatetimeClasses.js
+++ b/test/unit/core/DatetimeClasses.js
@@ -110,21 +110,28 @@ describe('DateTimeClassesTest', function () {
             (() => LocalTime.fromString(null)).should.throw(TypeError, 'String expected');
             (() => LocalTime.fromString()).should.throw(TypeError, 'String expected');
         });
+
+        it('should construct from fromDate correctly', function () {
+            const localTime1 = LocalTime.fromDate(new Date(2000, 2, 29, 0, 0, 0, 0));
+            localTime1.toString().should.be.eq('00:00:00');
+            const localTime2 = LocalTime.fromDate(new Date(2000, 0, 29, 2, 3, 4, 6));
+            localTime2.toString().should.be.eq('02:03:04.006000000');
+        });
     });
     describe('LocalDateTest', function () {
-        it('should throw RangeError if year is not an integer between -999_999_999-999_999_999(inclusive)',
-            function () {
-                (() => new LocalDate(1e9, 1, 1)).should.throw(RangeError, 'Year');
-                (() => new LocalDate(-1e9, 1, 1)).should.throw(RangeError, 'Year');
-                (() => new LocalDate(1.1, 1, 1)).should.throw(RangeError, 'All arguments must be integers');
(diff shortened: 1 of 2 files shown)


yuce:
@srknzl Codecov displays some warnings about coverage at: https://github.com/hazelcast/hazelcast-nodejs-client/pull/1180/files Could you check them?

mdumandag (approved):
I don't have anything extra other than what Yüce wrote

srknzl:
I handled both, thanks for the heads up!

yuce (approved):

yuce:
@mdumandag Could you re-approve this PR?

mdumandag (approved):

test/unit/core/DatetimeClasses.js
```diff
@@ -110,21 +111,43 @@ describe('DateTimeClassesTest', function () {
...
-            function () {
-                (() => new LocalDate(1e9, 1, 1)).should.throw(RangeError, 'Year');
-                (() => new LocalDate(-1e9, 1, 1)).should.throw(RangeError, 'Year');
-                (() => new LocalDate(1.1, 1, 1)).should.throw(RangeError, 'All arguments must be integers');
-                (() => new LocalDate('1', 1, 1)).should.throw(TypeError, 'All arguments must be numbers');
-                (() => new LocalDate({ 1: 1 }, 1, 1)).should.throw(TypeError, 'All arguments must be numbers');
-                (() => new LocalDate([], 1, 1)).should.throw(TypeError, 'All arguments must be numbers');
-                (() => new LocalDate(1e12, 1, 1)).should.throw(RangeError, 'Year');
-            });
+        it('should throw RangeError if year is not an integer between -999_999_999-999_999_999(inclusive)', function () {
```
ihsandemir, line 138: should we also add boundary test, e.g. 
`new LocalDate(-999_999_999, 1, 1)` should not throw. 
Similar for other tests as well. I assume the these ranges are same as Java ranges, right?

srknzl: >  I assume the these ranges are same as Java ranges, right?
Yes. 


>should we also add boundary test, e.g.
new LocalDate(-999_999_999, 1, 1) should not throw.

We did not test all boundary values but I see a boundary test on line 173. We can add them later.

PR #1184, opened 2022-01-31
Add CI workflow for gh actions due to Jenkins's removal

srknzl:
- Adds CI yaml. After we switch to github actions we did not add workflow files to older branches.
- Remove junit reporter, it was making `nyc` fail somehow with no output.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,48 @@
+name: Run tests
+on:
+    push:
+        branches-ignore:
+            - 'gh-pages'
+        tags-ignore:
+            - '*'
+    pull_request:
+
+jobs:
+    run-tests:
+        name: Run Tests on (${{ matrix.os }})
+        runs-on: ${{ matrix.os }}
+
+        strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,48 @@
+name: Run tests
+on:
+    push:
+        branches-ignore:
+            - 'gh-pages'
```
mdumandag, line 5: Will this work on push to master as well?

srknzl: I don't think so. There is a very similar thing in the workflow in master. We have the same `on.push` clause in that workflow and it does not run on pushes to maintenance branches.

srknzl: The push commit check runs on 5.0.x because it has a workflow file with `on.push`, but on 4.2.x there is no check on commit push since there is no workflow file in that branch. So I think every workflow file is local to its branch.

yuce (approved):

mdumandag (approved):

PR #1194, opened 2022-02-18
Compact & generic records reference manual API-1157

srknzl:
Includes compact serialization reference manual update.


zpqrtbnk (dismissed):
Approved, though I am wondering... a large amount of that doc should be in some sort of "common doc" to all clients, really.

srknzl:
> Approved, though I am wondering... a large amount of that doc should be in some sort of "common doc" to all clients, really.

Maybe in the future? As part of the unification of client docs efforts. I could have given a link to reference manual but I don't know I think it is better for users to read it here because compact is an important topic. Giving a link may seem like it is a detailed niche topic.

DOCUMENTATION.md
```diff
@@ -19,12 +19,17 @@
...
-  * [4.2. Portable Serialization](#42-portable-serialization)
-    * [4.2.1. Versioning for Portable Serialization](#421-versioning-for-portable-serialization)
-  * [4.3. Custom Serialization](#43-custom-serialization)
-  * [4.4. Global Serialization](#44-global-serialization)
-  * [4.5. JSON Serialization](#45-json-serialization)
+  * [4.1. Compact Serialization](#41-compact-serialization)
+    * [4.1.1. Schema Evolution](#411-schema-evolution)
+    * [4.1.2. Generic Record Representation](#412-generic-record-representation)
+    * [4.1.3. Compact Configuration and Usage](#413-compact-configuration-and-usage)
+    * [4.1.4. Limitations](#413-compact-configuration-and-usage)
```
yuce, line 26: Link doesn't match to the text.

DOCUMENTATION.md
```diff
@@ -711,7 +717,232 @@ Or, if you want to use your own serialization method, you can use [Custom Serial
 > [API Documentation](http://hazelcast.github.io/hazelcast-nodejs-client) or GitHub repository for a required
 > interface.**
 
-## 4.1. IdentifiedDataSerializable Serialization
+## 4.1. Compact Serialization
```
yuce, line 720: Do all of the documentation in this section apply to the Node.js client? At least "Does not require a class to implement an interface" does not apply, does it?

yuce (changes requested):
A couple of ideas...

DOCUMENTATION.md
srknzl: Actually the markdown link checker now can check these links. I just need to merge upstream master into this branch.

srknzl: Ah sorry it won't work because the destination exists but a wrong one

DOCUMENTATION.md
srknzl: Ok the sentence can be understood for the serializer instead of user class. Our serializers implement an interface. So I removed the `implementing an interface` part for Node.js

yuce (approved):

zpqrtbnk (approved):
Approved.

PR #1247, opened 2022-03-09
Increase timeouts in MapProxyTest [4.0.x] API-1200

### LINKED ISSUE #724: Flaky MapProxyTest [API-1200]
http://jenkins.hazelcast.com/job/NodeJS-12-master/93/testReport/junit/(root)/putTransient_withTTL/MapProxyTest_Near_Cache__true_putTransient_withTTL/

```
putTransient_withTTL.MapProxyTest Near Cache: true putTransient_withTTL (from Near Cache_ true)

Error Message
expected null to equal 'val10'
Stacktrace
AssertionError: expected null to equal 'val10'
    at Context.<anonymous> (test/map/MapProxyTest.js:325:32)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

srknzl:
Increased the timeouts that we increased through https://github.com/hazelcast/hazelcast-nodejs-client/pull/725 and https://github.com/hazelcast/hazelcast-nodejs-client/pull/1104

The test did not fail in master for a long time we must ensure the same happens in maintenance branches too.  Increases ttls to 20 seconds because according to Yuce and Sancar in go and java 20 seconds is used.

fixes #724

### DIFF
--- a/test/map/MapProxyTest.js
+++ b/test/map/MapProxyTest.js
@@ -113,7 +113,7 @@ describe('MapProxyTest', function () {
             });
 
             it('put with ttl puts value to map', function () {
-                return map.put('key-with-ttl', 'val-with-ttl', 3000).then(function () {
+                return map.put('key-with-ttl', 'val-with-ttl', 20000).then(function () {
                     return map.get('key-with-ttl').then(function (val) {
                         return expect(val).to.equal('val-with-ttl');
                     });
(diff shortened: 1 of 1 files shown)


test/map/MapProxyTest.js
```diff
@@ -113,20 +113,20 @@ describe('MapProxyTest', function () {
...
+                return map.put('key-with-ttl', 'val-with-ttl', 20000).then(function () {
                     return map.get('key-with-ttl').then(function (val) {
                         return expect(val).to.equal('val-with-ttl');
                     });
                 });
             });
 
             it('put with ttl removes value after ttl', function () {
-                return map.put('key10', 'val10', 1000).then(function () {
+                return map.put('key10', 'val10', 20000).then(function () {
```
mdumandag, line 124: I think it is safe to use a shorter timeout for this test. We already test the put and get scenario on `put with ttl puts value to map`.  I believe it would be OK to make the TTL shorter and get rid of the map.get assertion

mdumandag: Or, get rid of the `put with ttl puts value to map` altogether, and just have this test, which is similar to others

srknzl: Done. Forwardporting it to other branches

mdumandag (approved):

yuce (approved):

PR #1253, opened 2022-03-10
Decrease job count to 2

srknzl:
There are a lot of remote controller related fails in the past. These happen while creating a member or a cluster such as fail that happen in before all hooks, and `TApplicationError` errors.  I think these are related to parallel execution. I decreased the job count to decrease such errors. I measured the time necessary to spent running tests using with 2, 4 and 8 jobs: 

8: https://github.com/srknzl/hazelcast-nodejs-client/actions/runs/1962142954
4: https://github.com/srknzl/hazelcast-nodejs-client/actions/runs/1962128117
2: https://github.com/srknzl/hazelcast-nodejs-client/actions/runs/1962139938

As can be seen there is no big difference and we wait for windows run anyway. The windows run seems not to be affected by concurrency.

### DIFF
--- a/scripts/test-runner.js
+++ b/scripts/test-runner.js
@@ -178,11 +178,11 @@ if (process.argv.length === 3 || process.argv.length === 4) {
         if (process.argv.length === 4) {
             testCommand = `node node_modules/mocha/bin/mocha -g "${process.argv[3]}" ` +
                           '"test/integration/**/serial/**/*.js" && ';
-            testCommand += `node node_modules/mocha/bin/mocha -j 8 --parallel -g  "${process.argv[3]}" ` +
+            testCommand += `node node_modules/mocha/bin/mocha -j 2 --parallel -g  "${process.argv[3]}" ` +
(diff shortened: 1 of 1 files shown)


yuce (approved):
Approving, but I think these problems maybe related to tests interfering with each other (same map name, etc.) You may need to randomize DDS names, cluster names, etc.

srknzl:
Those are already randomized as far as I know. Every cluster has to have different IDs anyway because RC creates their ids. The DDS names won't matter much in this case.

yuce:
In Go tests, we mostly use a single cluster for all DDS tests. Is a separate cluster created for each test for Node.js client?

srknzl:
For every suite(which is a collection of some tests) a new cluster created, yes. Creating and destroying members actually makes the test duration much longer but it is better for test isolation.

srknzl:
@mdumandag Can you review this

mdumandag (approved):

PR #1277, opened 2022-04-26
Add test for proxy APIs to test if they throw serialization errors other than SchemaNotReplicatedError

srknzl:
Increases coverage. Tests the lines like this in all proxies: https://github.com/hazelcast/hazelcast-nodejs-client/blob/e1123a5ebdeba6d6736874d7e439b7a5b0b336d6/src/proxy/MapProxy.ts#L394

So basically the throw statement where a serialization fails and the failure cause is not SchemaNotReplicatedError. In this case the serialization error is rethrown.


zpqrtbnk (approved):
Approved.

yuce (approved):

PR #1293, opened 2022-05-12
New coverage runner

alisengul53:
This PR enables to run enterprise tests for PRs. Details are already explained here;
https://github.com/hazelcast/hazelcast-python-client/pull/539

Here is the Josef's comment for security concerns
https://github.com/hazelcast/hazelcast-python-client/pull/539#pullrequestreview-945333298

Note for reviewers: It is normal not to run `Run Test` jobs, because the event trigger mechanism is changed with this PR. When this PR is merged, `Run Test` Jobs will be triggered for new PRs.


.github/workflows/coverage_runner.yml
```diff
@@ -0,0 +1,99 @@
...
+        description: Enter PR number coming from community to run coverage on it. Do not run for PR coming from hazelcast organization, they will be run automatically.
+        required: true
+  push:
+    branches:
+      - master
+      - '[45].*.x'
+  pull_request_target:
+    branches:
+      - master
+      - '[45].*.x'
```
srknzl, line 16: In python workflow we have a 
```
types:
      [opened]
```
here. Why don't we have it here and what is it for?

.github/workflows/coverage_runner.yml
```diff
@@ -0,0 +1,99 @@
...
+            uses: codecov/codecov-action@v1
+            with:
+              files: coverage/lcov.info
+              
+          - name: Publish result to Codecov for PR coming from community
+            if: ${{ matrix.os == 'ubuntu-latest' && github.event_name == 'workflow_dispatch' }}
+            uses: codecov/codecov-action@v1
+            with:
+              files: coverage/lcov.info
+              override_pr: ${{ github.event.inputs.pr_number }}
```
srknzl, line 99: Can we also steal the following part from Python 😀 ? I know it is irrelevant to this PR, but we can handle this in this PR. I was planning to add it.

```
      - name: Upload remote controller logs if test run fails
        uses: actions/upload-artifact@v2
        if: failure()
        with:
          name: rc-logs-${{ matrix.os }}
          path: |
            rc_stderr.log
            rc_stdout.log
```
The skips won't trigger `failure()` I guess. So it should work.

srknzl (approved):

srknzl (changes requested):
Reverting the approve, it was accidential.

.github/workflows/coverage_runner.yml
alisengul53: No we don't. We had but I removed it because when setting the `[opened]` workflow is triggered only when it is opened. When you send new commits it is not triggered again.

>Note: By default, only the opened, synchronize, and reopened activity types trigger workflows that run on the pull_request event. To trigger workflows by different activity types, use the types keyword.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

srknzl: Ok thanks.

mdumandag (approved):

Issue #1286, opened 2022-05-06
Typescript compiler complaining in EntryListener.d.ts file


vjgn:
Typescript compiler started complaining today with errors like below:   

```
/node_modules/hazelcast-client/lib/proxy/EntryListener.d.ts:15:5 - error TS2411: Property 'mapCleared' of type 'MapEventListener<K, V> | undefined' is not assignable to 'string' index type 'EntryEventListener<K, V> | MapEventListener<K, V>'.
```

I appears that the `string` index property does not support `undefined` which the interface members can be (all of them are defined as `?:`)   

Doing the below changes got be going, that is:   
Change this:   
```
[event: string]: EntryEventListener<K, V> | MapEventListener<K, V>;
```
To this:   
```
[event: string]: EntryEventListener<K, V> | MapEventListener<K, V> | undefined;
```

Is there a better way to get around this issue?

Thank you.

Regards,
Vijay

srknzl:
Currently the compilation does not fail because this repo does not use strict null checking. You must have enabled it. If you enable it you will get hundreds of compilation errors, not just this one. We are planning to enable it one day, but it will take enourmous amount of effort for a relatively small benefit. The type definition files should not affect your project's compilation. Somehow you are compiling the typescript files inside `node_modules` folder. You need to exclude `node_modules` folder's compilation from your project. It is [by default ](https://www.typescriptlang.org/tsconfig#exclude) excluded. If you overrided the  `exclude` option, make sure to list `node_modules` as well in the `exclude` option.

vjgn:
We have strict null checking turned on. For some reason I only get an error with this one type `EntryListener<K, V>` (about 9 errors all about this Type only).

I have temporarily set `skipLibCheck` to `true` to move forward. That way d.ts files in node_modules folder gets ignored and TypeScript compiler no longer complains.
   
(I believe the types defined in node_modules will be referred even if node_modules folder itself is excluded explicitly in tsconfig.json. This is because if you refer a type elsewhere in your code, then TypeScript needs to figure out the definition for that type and it will check the node_modules folder looking for it)   

Thanks @srknzl for pointing me in the right direction. I still need to find a more permanent solution though.

srknzl:
Ok we can change the string index property to support undefined in the next releases.

srknzl:
@vjgn The fix is merged, it will be available in 5.2.

PR #1294, opened 2022-05-12
Add undefined to EntryListener interface

srknzl:
reference: https://github.com/hazelcast/hazelcast-nodejs-client/issues/1286

### DIFF
--- a/src/proxy/EntryListener.ts
+++ b/src/proxy/EntryListener.ts
@@ -31,7 +31,7 @@ export interface EntryListener<K, V> {
     mapEvicted?: MapEventListener<K, V>;
     mapCleared?: MapEventListener<K, V>;
 
-    [event: string]: EntryEventListener<K, V> | MapEventListener<K, V>;
+    [event: string]: EntryEventListener<K, V> | MapEventListener<K, V> | undefined;
 }
 
 /**


src/proxy/EntryListener.ts
```diff
@@ -31,7 +31,7 @@ export interface EntryListener<K, V> {
     mapEvicted?: MapEventListener<K, V>;
     mapCleared?: MapEventListener<K, V>;
 
-    [event: string]: EntryEventListener<K, V> | MapEventListener<K, V>;
+    [event: string]: EntryEventListener<K, V> | MapEventListener<K, V> | undefined;
```
mdumandag, line 34: Why do we need this line? I mean, aren't all possible events defined as properties above?

srknzl: It is defined but they can be undefined. This line is the general definition. It means any key can be mapped to either EntryEventListener, MapEventListener or undefined. A user mentioned if we don't put undefined here, typescript compilation fails for him.

srknzl: We may not need this line at all. Let me think

srknzl: We need this line in order to be able to index the type like `entryListener[stringVariable]`

srknzl: Yes, removing this line may make the code more complex because we will need more switch/case blocks.

mdumandag (approved):

PR #1319, opened 2022-06-17
Increase wait time for RC [4.0.x]

srknzl:
In 4.0.x and 4.1.x RC fails to start in 10 seconds leading all tests to fail. We increased the wait time in newer branches to be **maximum** 100 seconds. In 4.0.x and 4.1.x, I will make the wait time always 100 seconds. I will not bother with testing the 9701 port is up and running because doing that is more complex and may require the script to be rewritten in Node.js.

On windows, timeout don't work due to this error. https://knowledge.broadcom.com/external/article/29524/the-timeout-command-in-batch-script-job.html. As mentioned in the linked document, I used ping again. 

 Also, waiting extra 100 seconds is not critical because we only run tests of these branches nightly. 

Example fail: https://github.com/hazelcast/hazelcast-nodejs-client/runs/6854367568?check_suite_focus=true#step:9:1957

### DIFF
--- a/scripts/download-rc.bat
+++ b/scripts/download-rc.bat
@@ -94,4 +94,4 @@ if defined HAZELCAST_ENTERPRISE_KEY (
 start /min "hazelcast-remote-controller" cmd /c "java -Dhazelcast.enterprise.license.key=%HAZELCAST_ENTERPRISE_KEY% -cp %CLASSPATH% com.hazelcast.remotecontroller.Main --use-simple-server> rc_stdout.txt 2>rc_stderr.txt"
 
 echo wait for Hazelcast to start ...
-timeout /t 100
+ping -n 150 127.0.0.1 > nul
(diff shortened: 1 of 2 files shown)


scripts/download-rc.bat
```diff
@@ -94,4 +94,4 @@ if defined HAZELCAST_ENTERPRISE_KEY (
 start /min "hazelcast-remote-controller" cmd /c "java -Dhazelcast.enterprise.license.key=%HAZELCAST_ENTERPRISE_KEY% -cp %CLASSPATH% com.hazelcast.remotecontroller.Main --use-simple-server> rc_stdout.txt 2>rc_stderr.txt"
 
 echo wait for Hazelcast to start ...
-ping -n 15 127.0.0.1 > nul
+ping -n 150 127.0.0.1 > nul
```
utku-caglayan, line 97: Does this wait for 150 seconds? What is the default value for "-W" flag?

srknzl: It is 4 seconds but it is the timeout amount not the wait time for each ping. I guess the person who wrote this roughly calculated 15 pings take 10 seconds. Therefore, it should be enough to make it 150 now. Let me try it as well on local.

srknzl: image


15 pings takes at least 15 seconds on my local which is enough. If it takes more, it is not a problem

utku-caglayan: Thanks!

utku-caglayan (approved):

mdumandag (approved):

PR #1321, opened 2022-06-17
Update nightly workflows

srknzl:
Run nightlies in Node 10 and 18 which are the oldest supported and newest Node versions. Running on every versions makes tracking the bug harder since 64 of them are run in maintenance nightly. I hope running on two will be enough though.


yuce (approved):

mdumandag:
I believe we should at least occasionally test the client with all the versions we say we do support. Maybe daily for v10 and v18, weekly for others?

utku-caglayan (approved):

srknzl:
>  I believe we should at least occasionally test the client with all the versions we say we do support. Maybe daily for v10 and v18, weekly for others?


Let's discuss and if we want them I can add them later.

PR #1348, opened 2022-08-16
Fix invalid error message

fatihozer0:


srknzl:
I think first you need to send a fix to master. then we can backport it. also there are unrelated changes in this PR. maybe you chose 5.1.x branch accidentally ?

fatihozer0:
It continues on #1349

PR #1349, opened 2022-08-16
Fix invalid error message

fatihozer0:


src/discovery/HazelcastCloudDiscovery.ts
```diff
@@ -73,6 +73,10 @@ export class HazelcastCloudDiscovery {
 
         let dataAsAString = '';
         const req = get(endpointUrlOptions, (res: IncomingMessage) => {
+            if (res.statusCode != 200) {
+                deferred.reject(new HazelcastError('Your cluster discovery token is invalid.'));
+                return;
```
srknzl, line 78: what data is returned as a response if the token is invalid? Also can you add a test testing your change?  You can use the config from the github issue: `{ network: { hazelcastCloud: { discoveryToken: 'YOUR_DISCOVERY_TOKEN' } }, clusterName: 'YOUR_CLUSTER_NAME' }`

also you can also use some other random `discoveryToken`s for test variety. 

I could not find a integration test file (which are under `test/integration`) about cloud discovery, but you can create a new file. Alternatively, you can write a unit test by mocking the response.

fatihozer0: What should be returned?

srknzl: Two things I wanted here:

1. I wonder what is returned from cloud response when token is invalid. Can you test it to learn?
2. Can you add a unit test testing the logic you added, i.e your new if block. You can use a random token and assert that it throws a error with proper message (`Your cluster discovery token is invalid`). Instead of sending a real request to cloud in the test, you can mock sending a request using sinon.js. If that is too complex for you, let's send a real request.

fatihozer0:
I know unnecessary imports and some quality problems but before fixing them tests are not working I dont know why. Can you check?

test/integration/backward_compatible/serial/CloudDiscoveryTest.js
```diff
@@ -0,0 +1,46 @@
...
+
+        try {
+            testFactory.newHazelcastClientForSerialTests(cfg);
+        } catch (err) {
+            console.log(err);
+            expect(err).to.be.instanceof(HazelcastError);
+        }
+    });
+
+    it('invalid_discovery_token_with_anycluster', async function () {
```
srknzl, line 29: What do you mean by any cluster? 🤔 There's no need to write such a test. If you are writing a serial test, you should ideally start clients only with `testFactory.newHazelcastClientForSerialTests`. Also, this test can be a parallel test. Can you move it?

srknzl: @fatihozer0 

Forget about all my comments about this test file. It is not easy to test our change via starting a client. So we will use a unit test.

I have written the following test:

```js
'use strict';

const { expect } = require('chai');
const { HazelcastError } = require('../../../../lib');
const { ClientConfigImpl } = require('../../../../lib/config/Config');
const { HazelcastCloudDiscovery } = require('../../../../lib/discovery/HazelcastCloudDiscovery');

describe('Cloud Discovery Token Test', function () {
    it('should throw proper error with an invalid token', async function () {
        const clientConfig = new ClientConfigImpl();
        const cloudDiscovery = new HazelcastCloudDiscovery(
            clientConfig.properties['hazelcast.client.cloud.url'], 2 << 31
        );

        try {
            await cloudDiscovery.callService();
        } catch (err) {
            expect(err).to.be.instanceof(HazelcastError);
            expect(err.message).to.include('discovery token is invalid');
        }
    });
});
```

You can add it under `test/unit` folder maybe with a name like `CloudDiscoveryInvalidTokenTest.js`. This test is passing. We can walk through it via huddle.

src/discovery/HazelcastCloudDiscovery.ts
srknzl: I have learned 1 by myself. Returned response 

```
Caused by: java.io.IOException: {"application":"orchestrator-service","message":"Cluster token not found for token.","status":404,"error":"Not Found","errorCode":"ClusterTokenNotFound","timestamp":1663100479625,"path":"https://coordinator.hazelcast.cloud/cluster/discovery","method":"GET"}
	at com.hazelcast.client.impl.spi.impl.discovery.HazelcastCloudDiscovery.checkError(HazelcastCloudDiscovery.java:141)
	at com.hazelcast.client.impl.spi.impl.discovery.HazelcastCloudDiscovery.callService(HazelcastCloudDiscovery.java:85)
	at com.hazelcast.client.impl.spi.impl.discovery.HazelcastCloudDiscovery.discoverNodes(HazelcastCloudDiscovery.java:70)
	... 10 more

```

test/unit/discovery/CloudDiscoveryInvalidTokenTest.js
```diff
@@ -0,0 +1,29 @@
...
+describe('Cloud Discovery Invalid Token Test', function () {
+    it('Cloud Discovery should throw proper error with an invalid token', async function () {
+        const clientConfig = new ClientConfigImpl();
+        const cloudDiscovery = new HazelcastCloudDiscovery(
+            clientConfig.properties['hazelcast.client.cloud.url'], 2 << 31
+        );
+
+        try {
+
+            const error = getRejectionReasonOrThrow(cloudDiscovery.callService);
```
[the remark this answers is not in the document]
srknzl: You need to await it.

test/unit/discovery/CloudDiscoveryInvalidTokenTest.js
```diff
@@ -0,0 +1,25 @@
...
+const {getRejectionReasonOrThrow} = require("../../TestUtil");
+
+describe('Cloud Discovery Invalid Token Test', function () {
+    it('Cloud Discovery should throw proper error with an invalid token', async function () {
+        const clientConfig = new ClientConfigImpl();
+        const cloudDiscovery = new HazelcastCloudDiscovery(
+            clientConfig.properties['hazelcast.client.cloud.url'], 2 << 31
+        );
+
+        try {
```
fatihozer0, line 16: This part returns `TypeError: Cannot read properties of undefined (reading 'endpointUrl')`

test/integration/backward_compatible/serial/CloudDiscoveryTest.js
srknzl: @fatihozer0 You can delete this file now.

fatihozer0: Done.

fatihozer0:
All the comments are done. I think we can approve

srknzl (approved):

PR #1350, opened 2022-08-16
Port nested compact serialization sample [API-1332]

fatihozer0:


srknzl:
There are some warnings in eslint. Can you run `npm run lint` in your local and fix them?

code_samples/compact_serialization_nested_sample.js
```diff
@@ -0,0 +1,98 @@
...
+                )
+            )
+        );
+
+        const result = await map.get(0);
+        console.log(result);
+
+        await client.shutdown();
+    } catch (err) {
+        console.error('Error occurred:', err);
```
srknzl, line 96: can you add `process.exit(1);` here? We add it to avoid script running forever if client stays open. In Node.js, if there is a socket running in the background process does not end, unless you end it yourself.

srknzl (changes requested):
one change requested

srknzl (approved):

PR #1364, opened 2022-09-05
Add map.RemoveAll() method [API-15]

fatihozer0:


test/integration/backward_compatible/parallel/map/NearCachedMapTest.js
```diff
@@ -179,6 +180,11 @@ describe('NearCachedMapTest', function () {
                 expectStats(map1, 0, 1, 0);
             });
 
+            it('removeAll clears nearcache', async function () {
+                this.map.removeAll(true);
+                this.assertEqual(0, this.map.wrapped.nearCache);
+            });
```
srknzl, line 186: `this.map` exists due to test runtime, mocha but you should not be using that. Our map proxies are map1 and map2.

let's think about what to test about removeAll on nearcachedproxy. There is one thing extra in nearcachedmapproxy compared to mapproxy, and that is clearing the near cache. So, let's test that.


 The test I wrote is inspired from `clear clears nearcache`:

```suggestion
            it('removeAll clears nearcache', async function () {
                await map1.get('key1');
                await map1.removeAll(Predicates.alwaysTrue());
                expectStats(map1, 0, 1, 0);
            });
```
Initially near cache is empty

1. With `await map1.get('key1');`, a cache miss happens. Also, entry key1 -> value1 is cached. So entry count is 1 here. 
2. we call removeAll, near cache is cleared.
3. with `expectStats` we assert that entry count in near cache is 0, meaning it is cleared.

srknzl (changes requested):

fatihozer0:
Commented test parts are not working. I will fix with your lead.

src/proxy/IMap.ts
```diff
@@ -269,6 +269,13 @@ export interface IMap<K, V> extends DistributedObject {
      */
     remove(key: K, value?: V): Promise<V | boolean>;
 
+    /**
+     * Removes all entries which match with the supplied predicate.
```
srknzl, line 273: You can add the following sentence as well:

- If this map has index, matching entries will be found via index search, otherwise they will be found by full-scan.

src/proxy/MapProxy.ts
```diff
@@ -838,8 +829,16 @@ export class MapProxy<K, V> extends BaseProxy implements IMap<K, V> {
...
 
     protected removeAllInternal(predicate: Predicate): Promise<void> {
-        const request = MapRemoveAllCodec.encodeRequest(this.name, this.toData(predicate));
-        return this.encodeInvokeOnRandomTarget(MapRemoveAllCodec, () => {}, request);
+        let predicateData: Data;
+        try {
+            predicateData = this.toData(predicate);
+        } catch (e) {
+            if (e instanceof SchemaNotReplicatedError) {
+                return this.registerSchema(e.schema, e.clazz).then(() => this.removeAllInternal(predicate));
```
srknzl, line 837: There is a warning about coverage on this line. This is because you did not test removeAll method with compact. You can add removeAll method to `CompactPublicAPIsTest`. Note that, you need to use `CompactPredicate()` as the predicate.

srknzl: You can add new test here, under "remove"

 
image

src/proxy/NearCachedMapProxy.ts
```diff
@@ -194,7 +194,14 @@ export class NearCachedMapProxy<K, V> extends MapProxy<K, V> {
     }
 
     protected removeAllInternal(predicate: Predicate): Promise<void> {
-        this.nearCache.clear();
+        try {
+            this.nearCache.clear();
+        } catch (e) {
+            if (e instanceof SchemaNotReplicatedError) {
+                return this.registerSchema(e.schema, e.clazz).then(() => this.removeAllInternal(predicate));
```
srknzl, line 201: After you add a test to CompactPublicAPIsTest as I mentioned, coverage warnings for this method will also disappear

test/integration/backward_compatible/serial/map/MapProxyTest.js
```diff
@@ -153,6 +153,42 @@ describe('MapProxyTest', function () {
...
+
+            // it('remove_all_with_null_predicate', async function () {
+            //     const predicate = await Predicates.equal('this', null);
+            //     const respond = await map.removeAll(predicate);
+            //     expect(respond).throws(AssertionError);
+            // });
+
+            it('removes_all_entries_whenPredicateTrue', async function () {
+                const predicate = await Predicates.alwaysTrue();
+                await map.removeAll(predicate);
```
srknzl, line 171: no need for await 

```suggestion
                await map.removeAll(Predicates.alwaysTrue());
```

test/integration/backward_compatible/serial/map/MapProxyTest.js
```diff
@@ -153,6 +153,42 @@ describe('MapProxyTest', function () {
...
+                const size = await map.size();
+                expect(size).to.equal(6);
+            });
+
+            // it('remove_all_with_null_predicate', async function () {
+            //     const predicate = await Predicates.equal('this', null);
+            //     const respond = await map.removeAll(predicate);
+            //     expect(respond).throws(AssertionError);
+            // });
```
srknzl, line 167:
```suggestion
it('removeAll_throws_exception_whenPredicateNull', async function () {
    const rejectionReason = await TestUtil.getRejectionReasonOrThrow(async () => await map.removeAll(null));
    expect(rejectionReason).to.be.instanceOf(AssertionError);
});
```

please note that AssertionError needs to be imported from 'assert' package:

`const { AssertionError } = require('assert');`

src/proxy/NearCachedMapProxy.ts
```diff
@@ -194,7 +194,14 @@ export class NearCachedMapProxy<K, V> extends MapProxy<K, V> {
     }
 
     protected removeAllInternal(predicate: Predicate): Promise<void> {
```
srknzl, line 196: Can you change removeAllInternal to be as follows logicwise:

```java
try {
    super.removeAllInternal(predicate);
} finally {
    nearCache.clear();
}
```

srknzl: https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/main/java/com/hazelcast/client/map/impl/nearcache/NearCachedClientMapProxy.java#L204

src/proxy/NearCachedMapProxy.ts
srknzl: not important, because you should remove this catch block.

fatihozer0:
I know some tests are not working but not sure why

src/proxy/NearCachedMapProxy.ts
srknzl: @fatihozer0 It should be as follows:


```js
try {
    return super.removeAllInternal(predicate);
} finally {
    this.nearCache.clear();
}
```

srknzl:
> I know some tests are not working but not sure why

possibly due to https://github.com/hazelcast/hazelcast-nodejs-client/pull/1364#discussion_r965768655

srknzl (approved):

PR #1367, opened 2022-09-14
Implement TopicProxy [API-1521]

fatihozer0:
This PR will add TopicProxy and edit ITopic interface. See the proposed changes in the [TDD](https://hazelcast.atlassian.net/wiki/spaces/HZC/pages/4239163410/Node.js+Topic+Sync+Interface+Solution+Options).

There are many problems. All in the comments of the files. They will be deleted. Lint problems will be fixed too.


src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+import {MessageListener} from "../MessageListener";
+import {UUID} from "../../core";
+import {TopicAddMessageListenerCodec} from "../../codec/TopicAddMessageListenerCodec";
+import {TopicRemoveMessageListenerCodec} from "../../codec/TopicRemoveMessageListenerCodec";
+import {TopicPublishCodec} from "../../codec/TopicPublishCodec";
+import { ListenerMessageCodec} from "../../listener/ListenerMessageCodec";
+import {ClientMessage, ClientMessageHandler} from "../../protocol/ClientMessage";
+import {assertNotNull} from "../../util/Util";
+
+export class TopicProxy<E> extends BaseProxy implements ITopic<E> {
```
srknzl, line 12:
```suggestion
export class TopicProxy<E> extends PartitionSpecificProxy implements ITopic<E> {
```

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    publish(message: E): Promise<void> {
+        assertNotNull(message);
+        const data = this.toData(message);
+        const request = TopicPublishCodec.encodeRequest(this.name, data);
+        const partitionId = request.getPartitionId();
+        return this.encodeInvokeOnPartition<void>(TopicAddMessageListenerCodec,  partitionId, TopicAddMessageListenerCodec.decodeResponse);
```
srknzl, line 19: After extending `PartitionSpecificProxy` you can just use `encodeInvoke` method of it.

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    addMessageListener(listener: MessageListener<E>): string {
```
srknzl, line 22: can you update MessageListener's (and also Message's) documentation? 
```
/**
 * Message listener for Reliable Topic.
 */
```

We should say ITopic to include both proxies.

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    removeMessageListener(listenerId: string): boolean {
+        if (this.listenerService.deregisterListener(listenerId)){
+            return true;
+        }
+        return false;
+    }
+
+    private createItemListener(name: string, includeValue: boolean): ListenerMessageCodec {
```
srknzl, line 40: `createListenerMessageCodec()` is a better name

srknzl, line 40: includeValue is not used

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+        return {
+            encodeAddRequest(localOnly: boolean): ClientMessage {
+                return TopicAddMessageListenerCodec.encodeRequest(name, localOnly);
```
srknzl, line 43: you can use `this.name` instead of getting it from parameter.

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+            },
+            decodeAddResponse(msg: ClientMessage): UUID {
+                return TopicAddMessageListenerCodec.decodeResponse(msg);
+            },
+            encodeRemoveRequest(listenerId: UUID): ClientMessage {
+                return TopicAddMessageListenerCodec.encodeRequest(name, false);
```
srknzl, line 49: this should use TopicRemove..Codec

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+        const request = TopicPublishCodec.encodeRequest(this.name, data);
+        const partitionId = request.getPartitionId();
+        return this.encodeInvokeOnPartition<void>(TopicAddMessageListenerCodec,  partitionId, TopicAddMessageListenerCodec.decodeResponse);
+    }
+
+    addMessageListener(listener: MessageListener<E>): string {
+        assertNotNull(listener);
+        const message = TopicAddMessageListenerCodec.encodeRequest(this.name, false);
+        const codec = this.createItemListener(this.name, false)
+        TopicAddMessageListenerCodec.handle(message, () => {});
```
srknzl, line 26: this line should be removed.

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+        //I am unable to fix this part. Will be trying tomorrow
+        return this.listenerService.registerListener(codec,  (message => {
+            TopicAddMessageListenerCodec.handle(message, ()=> {})
+        }));
```
srknzl, line 30:
```suggestion
        return this.listenerService.registerListener(codec, listener);
```

It should be this way IMO

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    removeMessageListener(listenerId: string): boolean {
+        if (this.listenerService.deregisterListener(listenerId)){
```
srknzl, line 34: you can simplify this by removing if statement.

src/proxy/topic/TopicProxy.ts
fatihozer0: `<html>TS2345: Argument of type 'MessageListener<E>' is not assignable to parameter of type 'ClientMessageHandler'.<br/>Types of parameters 'message' and 'message' are incompatible.<br/>Type 'ClientMessage' is missing the following properties from type 'Message<E>': messageObject, publisher, publishingTime` the reason I tried this way is this

src/proxy/topic/TopicProxy.ts
fatihozer0: `TS2322: Type 'Promise<boolean>' is not assignable to type 'boolean'.` I did in this way. Any other way to solve?

src/proxy/topic/TopicProxy.ts
srknzl: yeah, you need to convert the `ClientMessage` to a topic `Message`. `TopicItemHandler` does that in java. 

Instead of it, you can pass something like the following to `registerListener` method:
```js
(clientMessage: ClientMessage) => {
    TopicAddMessageListenerCodec.handle(clientMessage, (item: Data, publishTime: Long, uuid: UUID) => {
    // And then write the equivalent of the java code here:
    /*

            Member member = getContext().getClusterService().getMember(uuid);
            Message message = new DataAwareMessage(name, item, publishTime, member, getSerializationService());
            listener.onMessage(message);

    */
    });
} 
```

src/proxy/topic/TopicProxy.ts
srknzl: Ok this is some real problematic thing. ITopic.removeMessageListener is added synchronously because in reliable topic it can be synchronous but in ITopic it cannot be. Changing ITopic is a breaking change. I don't know what we should do. Let us talk in the standup tomorrow.

src/proxy/topic/TopicProxy.ts
```diff
@@ -13,22 +12,23 @@ export class TopicProxy<E> extends PartitionSpecificProxy implements ITopic<E> {
 
     publish(message: E): Promise<void> {
         assertNotNull(message);
-        const data = this.toData(message);
-        const request = TopicPublishCodec.encodeRequest(this.name, data);
         return this.encodeInvoke<void>(TopicAddMessageListenerCodec,
                                        TopicAddMessageListenerCodec.decodeResponse);
```
srknzl, line 16: You should use publish codec

srknzl: I recomment you to check other partitionspecificproxies and how they use encodeInvoke

src/proxy/topic/TopicProxy.ts
srknzl: You can just look for the string `reliable` in client repo and change the places you should change. Things that are common should say ITopic instead.

fatihozer0:
I get this error: 
`RangeError: There is no suitable deserializer for data with type 12582913
    at SerializationServiceV1.toObject (/Users/fatihozer/Desktop/hazelcast/client/hazelcast-nodejs-client/lib/serialization/SerializationService.js:103:19)
    at TopicProxy.toObject (/Users/fatihozer/Desktop/hazelcast/client/hazelcast-nodejs-client/lib/proxy/BaseProxy.js:135:42)
    at /Users/fatihozer/Desktop/hazelcast/client/hazelcast-nodejs-client/lib/proxy/topic/TopicProxy.js:40:42
    at Function.handle (/Users/fatihozer/Desktop/hazelcast/client/hazelcast-nodejs-client/lib/codec/TopicAddMessageListenerCodec.js:60:13)
    at Invocation.handler [as eventHandler] (/Users/fatihozer/Desktop/hazelcast/client/hazelcast-nodejs-client/lib/proxy/topic/TopicProxy.js:37:73)
    at InvocationService.callEventHandlerWithMessage (/Users/fatihozer/Desktop/hazelcast/client/hazelcast-nodejs-client/lib/invocation/InvocationService.js:337:24)
    at /Users/fatihozer/Desktop/hazelcast/client/hazelcast-nodejs-client/lib/invocation/InvocationService.js:369:26
    at processTicksAndRejections (node:internal/process/task_queues:78:11)

` while trying to run [this code](https://gist.github.com/fatihozer0/5b2f82ffebccff489497c43dca3905fe)

srknzl:
it seems totally weird. Let me try the code sample. I will comment about it

srknzl:
```diff
diff --git a/src/proxy/topic/TopicProxy.ts b/src/proxy/topic/TopicProxy.ts
index d6cfc637..5a2224f0 100644
--- a/src/proxy/topic/TopicProxy.ts
+++ b/src/proxy/topic/TopicProxy.ts
@@ -64,19 +64,16 @@ export class TopicProxy<E> extends PartitionSpecificProxy implements ITopic<E> {
         assertNotNull(message);
 
         const messageData = this.toData(message);
-        const request = TopicPublishCodec.encodeRequest(this.name, messageData);
         const partitionId = this.partitionService.getPartitionId(messageData);
-        return this.encodeInvokeOnPartition(TopicPublishCodec, partitionId, () => {
-        }, request);
+        return this.encodeInvokeOnPartition(TopicPublishCodec, partitionId, () => {}, messageData);
     }
 
     publishAll(messages: any[]): Promise<void> {
         assertNotNull(messages);
         const messageDataList = this.toData(messages);
-        const request = TopicPublishCodec.encodeRequest(this.name, messageDataList);
         const partitionId = this.partitionService.getPartitionId(messageDataList);
         return this.encodeInvokeOnPartition(TopicPublishCodec, partitionId, () => {
-        }, request);
+        }, messageDataList);
     }
 
     addListener(listener: MessageListener<E>): Promise<string> {
```

srknzl:
@fatihozer0 If you apply this diff, it will be fixed. There is an error in your encodeInvokeOnPartition method usage.

fatihozer0:
> @fatihozer0 If you apply this diff, it will be fixed. There is an error in your encodeInvokeOnPartition method usage.

Now it looks like the sample is working. If everything looks good I am going to fix minor problems, lints etc. @srknzl

srknzl:
@fatihozer0 Checking the PR. I will let you know how the PR looks like right now.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +151,61 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+                    }
+                    break;
+                default:
+                    throw new RangeError('Unknown overload policy');
+            }
+        } catch (e) {
+            if (e instanceof SchemaNotReplicatedError) {
+                return this.registerSchema(e.schema, e.clazz).then(() => this.publishAll(messages));
+            }
+            throw e;
```
srknzl, line 194: Java wraps the error:

`throw (RuntimeException) peel(e, null,
                    "Failed to publish messages: " + payload + " to topic:" + getName());`


This converts exceptions to RuntimeException basically. We can throw a HazelcastError IMO instead of that since RuntimeException does not exist in node.js. Let's say we throw HazelcastError, we should add it to the documentation of ITopic.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +151,61 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+            for (const message of messages) {
+                const reliableTopicMessage = new ReliableTopicMessage();
+                reliableTopicMessage.payload = this.serializationService.toData(message);
+                reliableTopicMessage.publishTime = Long.fromNumber(Date.now());
+                reliableTopicMessage.publisherAddress = this.localAddress;
+                reliableTopicMessages.push(reliableTopicMessage);
+            }
+            switch (this.overloadPolicy) {
+                case TopicOverloadPolicy.ERROR:
+                    for (const reliableTopicMessage of reliableTopicMessages) {
```
srknzl, line 168: Instead of for loops and add() methods of ringbuffer, you should use addAll methods like Java. You also need to await these promises.

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+        this.localAddress = connection != null ? connection.getLocalAddress() : null;
+        const config = (clientConfig as ClientConfigImpl).getReliableTopicConfig(name);
+        this.overloadPolicy = config.overloadPolicy;
+    }
+
+    publish(message: E): Promise<void> {
+        assertNotNull(message);
+
+        const messageData = this.toData(message);
+        const partitionId = this.partitionService.getPartitionId(messageData);return this.encodeInvokeOnPartition(TopicPublishCodec, partitionId, () => {}, messageData);
```
srknzl, line 67: Line break needed

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+        };
+    }
+
+    removeListener(listenerId: string): Promise<boolean> {
+        return this.listenerService.deregisterListener(listenerId);
+    }
+
+    addMessageListener(listener: MessageListener<E>): string {
+        throw new HazelcastError('This method is not supported for Topic. ' +
+            'Try to use addListener instead.');
```
srknzl, line 116:
```suggestion
            'Use addListener instead.');
```

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    removeMessageListener(listenerId: string): boolean {
+        throw new HazelcastError('This method is not supported for Topic. ' +
+            'Try to use removeListener instead.');
```
srknzl, line 121:
```suggestion
            'Use removeListener instead.');
```

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+import {ConnectionRegistry} from '../../network/ConnectionRegistry';
+import {SchemaService} from '../../serialization/compact/SchemaService';
+import {Connection} from '../../network/Connection';
+import {TopicOverloadPolicy} from '../TopicOverloadPolicy';
+
+export class TopicProxy<E> extends PartitionSpecificProxy implements ITopic<E> {
+
+
+    private readonly overloadPolicy: TopicOverloadPolicy;
+    private readonly localAddress: AddressImpl;
```
srknzl, line 29: Why do we have them if they are not used?

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+            serializationService,
+            listenerService,
+            clusterService,
+            connectionRegistry,
+            schemaService
+        );
+        const connection: Connection = this.connectionRegistry.getRandomConnection();
+        this.localAddress = connection != null ? connection.getLocalAddress() : null;
+        const config = (clientConfig as ClientConfigImpl).getReliableTopicConfig(name);
+        this.overloadPolicy = config.overloadPolicy;
```
srknzl, line 60:
```suggestion
```

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+
+
+    private readonly overloadPolicy: TopicOverloadPolicy;
+    private readonly localAddress: AddressImpl;
+
+
+    constructor(
+        serviceName: string,
+        name: string,
+        clientConfig: ClientConfig,
```
srknzl, line 35:
```suggestion
```

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+            connectionRegistry,
+            schemaService
+        );
+        const connection: Connection = this.connectionRegistry.getRandomConnection();
+        this.localAddress = connection != null ? connection.getLocalAddress() : null;
+        const config = (clientConfig as ClientConfigImpl).getReliableTopicConfig(name);
+        this.overloadPolicy = config.overloadPolicy;
+    }
+
+    publish(message: E): Promise<void> {
```
srknzl, line 63: This method should be like this: 

```js
publish(message: E): Promise<void> {
        assertNotNull(message);

        const messageData = this.toData(message);
        return this.encodeInvoke(TopicPublishCodec, () => {}, messageData);
    }
```

Can you write some tests and try this if the tests pass? 

The explanation: You were using the partition id of the message. The idea is that we need to use partition id of the topic itself calculated using the proxy name because Topic is partitioned according to the name of the topic, not the data. You can see the details of how it is done in PartitionSpecificProxy class. That's why we use encodeInvoke now.

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+        assertNotNull(message);
+
+        const messageData = this.toData(message);
+        const partitionId = this.partitionService.getPartitionId(messageData);return this.encodeInvokeOnPartition(TopicPublishCodec, partitionId, () => {}, messageData);
+    }
+
+    publishAll(messages: any[]): Promise<void> {
+        assertNotNull(messages);
+        const messageDataList = this.toData(messages);
+        const partitionId = this.partitionService.getPartitionId(messageDataList);
```
srknzl, line 73: similar comment for this one about partition id.

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    publish(message: E): Promise<void> {
+        assertNotNull(message);
+
+        const messageData = this.toData(message);
+        const partitionId = this.partitionService.getPartitionId(messageData);return this.encodeInvokeOnPartition(TopicPublishCodec, partitionId, () => {}, messageData);
+    }
+
+    publishAll(messages: any[]): Promise<void> {
+        assertNotNull(messages);
```
srknzl, line 71: You need to check all messages for nullness.

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,123 @@
...
+        const messageDataList = this.toData(messages);
+        const partitionId = this.partitionService.getPartitionId(messageDataList);
+        return this.encodeInvokeOnPartition(TopicPublishCodec, partitionId, () => {
```
srknzl, line 74: You are using the wrong codec.

src/proxy/topic/TopicProxy.ts
srknzl: You did not do this still as I see.

srknzl:
@fatihozer0 I reviewed your PR but the most important thing is missing: tests

src/proxy/ITopic.ts
```diff
@@ -38,8 +38,13 @@ export interface ITopic<E> extends DistributedObject {
      *
```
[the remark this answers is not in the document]
fatihozer0: I see now. Got what you wanted to tell. Gonna fix them

src/proxy/topic/ReliableTopicProxy.ts
fatihozer0: The thing that confuses me is should I implement publishAll or publishAllAsync?

src/proxy/topic/TopicProxy.ts
fatihozer0: I would fix them after major things are done. Got them.

fatihozer0:
> @fatihozer0 I reviewed your PR but the most important thing is missing: tests

Added but last 2 tests are timed out. Where do you think the problem is?

src/proxy/topic/TopicProxy.ts
fatihozer0: Done

src/proxy/topic/TopicProxy.ts
fatihozer0: Done

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +151,61 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
         }
     }
 
+    publishAll(messages: any[]): Promise<void> {
+        const reliableTopicMessages: ReliableTopicMessage[] = [];
+        assertNotNull(messages);
+
```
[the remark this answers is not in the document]
fatihozer0: Done

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +151,61 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+    addListener(listener: MessageListener<E>): Promise<string> {
+        throw new HazelcastError('This method is not supported for Reliable Topic.' +
+                                        ' Use addMessageListener instead.');
+    }
+
+    removeListener(listenerId: string): Promise<boolean> {
+        throw new HazelcastError('This method is not supported for Reliable Topic.' +
+                                        ' Use removeMessageListener instead.');
+    }
```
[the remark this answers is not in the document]
fatihozer0: Done

srknzl:
> Added but last 2 tests are timed out. Where do you think the problem is?

No idea, did you try debugging? I will check this PR again when you resolve all my comments.

src/proxy/ITopic.ts
```diff
@@ -38,8 +38,13 @@ export interface ITopic<E> extends DistributedObject {
...
      * More than one message listener can be added on one instance.
      *
+     * This method is for only Reliable Topic since this is a sync method.
```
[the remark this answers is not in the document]
fatihozer0: Done.

src/proxy/ITopic.ts
```diff
@@ -60,4 +70,42 @@ export interface ITopic<E> extends DistributedObject {
...
+
+    /**
+     * Subscribes to this topic. When a message is published,
+     * the given MessageListener is called.
+     *
+     * More than one message listener can be added on one instance.
+     *
+     * This method is for only Topic since this is an async method.
+     * When this method is called from Reliable Topic, it throws HazelcastError.
+     * Check addMessageListener method for using with Reliable Topic.
```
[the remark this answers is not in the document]
fatihozer0: Done.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +151,61 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+    publishAll(messages: any[]): Promise<void> {
+        const reliableTopicMessages: ReliableTopicMessage[] = [];
+        assertNotNull(messages);
+
+        try {
+            for (const message of messages) {
+                const reliableTopicMessage = new ReliableTopicMessage();
+                reliableTopicMessage.payload = this.serializationService.toData(message);
+                reliableTopicMessage.publishTime = Long.fromNumber(Date.now());
+                reliableTopicMessage.publisherAddress = this.localAddress;
```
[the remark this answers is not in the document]
fatihozer0: Done.

src/proxy/topic/ReliableTopicProxy.ts
fatihozer0: I see, I think done now.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +151,61 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+                        this.addOrOverwrite(reliableTopicMessage);
+                    }
+                    break;
+                case TopicOverloadPolicy.BLOCK:
+                    for (const reliableTopicMessage of reliableTopicMessages) {
+                        this.addWithBackoff(reliableTopicMessage);
+                    }
+                    break;
+                default:
+                    throw new RangeError('Unknown overload policy');
```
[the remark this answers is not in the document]
fatihozer0: Done.

src/proxy/topic/ReliableTopicProxy.ts
fatihozer0: Done.

src/proxy/topic/TopicProxy.ts
fatihozer0: Done.

src/proxy/topic/TopicProxy.ts
fatihozer0: Done.

src/proxy/topic/TopicProxy.ts
fatihozer0: I commented below about tests

src/proxy/topic/TopicProxy.ts
fatihozer0: Done.

src/proxy/topic/TopicProxy.ts
fatihozer0: Done.

src/proxy/topic/TopicProxy.ts
fatihozer0: ListenerMessageCodec does not have the property of name, so I can not use it.

src/proxy/topic/TopicProxy.ts
fatihozer0: Done

src/proxy/topic/TopicProxy.ts
fatihozer0: Done.

src/proxy/topic/TopicProxy.ts
srknzl:
```js
private createListenerCodec(): ListenerMessageCodec {
        const name = this.name;
        return {
            encodeAddRequest(localOnly: boolean): ClientMessage {
                return TopicAddMessageListenerCodec.encodeRequest(name, localOnly);
            },
            decodeAddResponse(msg: ClientMessage): UUID {
                return TopicAddMessageListenerCodec.decodeResponse(msg);
            },
            encodeRemoveRequest(): ClientMessage {
                return TopicAddMessageListenerCodec.encodeRequest(name, super.localOnly);
            },
        };
    }
```

This is how you can do it

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,112 @@
+import {ITopic} from '../ITopic';
```
srknzl, line 1: You need to add copyright here. Also 

`/** @ignore *//** */` under the copyright. See other proxy class files. Also you need to add `/** @internal */` to all exported classes.

src/proxy/ITopic.ts
```diff
@@ -38,8 +38,15 @@ export interface ITopic<E> extends DistributedObject {
...
+     * When this method is called from Topic, it throws {@link HazelcastError}.
+     * Check addListener method for using with Topic.
+     *
+     * `addMessageListener` is @deprecated since 5.3. Use {@link addListener} instead.
+     *
      * @param listener the MessageListener to add
      * @return registration ID
+     * @throws {HazelcastError} if it is used from Topic
      */
     addMessageListener(listener: MessageListener<E>): string;
```
srknzl, line 51:
```suggestion
     * This method is for only Reliable Topic since this is a sync method.
     * When this method is called from Topic, it throws {@link HazelcastError}.
     * Check addListener method for using with Topic.
     *
     * @deprecated `addMessageListener` is deprecated since 5.3. Use {@link addListener} instead.
     *
     * @param listener the MessageListener to add
     * @return registration ID
     * @throws {@link HazelcastError} if it is used from Topic
     */
    addMessageListener(listener: MessageListener<E>): string;
```

srknzl: I suggest you to check the generated documentation as well.

src/proxy/ITopic.ts
```diff
@@ -48,8 +55,15 @@ export interface ITopic<E> extends DistributedObject {
...
      *
+     * This method is for only Reliable Topic.
+     * When this method is called from Topic, it throws {@link HazelcastError}.
+     * Check removeListener method for using with Topic.
+     *
+     * `removeMessageListener` is @deprecated since 5.3. Use {@link removeListener} instead.
+     *
      * @param listenerId listener registration ID
      * @return `true` if registration is removed, `false` otherwise
+     * @throws {HazelcastError} if it is used from Topic
```
srknzl, line 66:
```suggestion
     * This method is for only Reliable Topic.
     * When this method is called from Topic, it throws {@link HazelcastError}.
     * Check removeListener method for using with Topic.
     *
     * @deprecated `removeMessageListener` is deprecated since 5.3. Use {@link removeListener} instead.
     *
     * @param listenerId listener registration ID
     * @return `true` if registration is removed, `false` otherwise
     * @throws {@link HazelcastError} if it is used from Topic
```

src/proxy/ITopic.ts
```diff
@@ -22,7 +22,7 @@ import {MessageListener} from './MessageListener';
  * delivered to multiple subscribers, which is also known as a publish/subscribe
  * (pub/sub) messaging model. Publish and subscriptions are cluster-wide.
  *
- * This interface stand for reliable topic, i.e. it uses a Ringbuffer to store
+ * This interface stand for topic and reliable topic. Reliable uses a Ringbuffer to store
```
srknzl, line 25:
```suggestion
 * This interface stand for topic and reliable topic. Reliable topic uses a Ringbuffer to store
```

src/proxy/ITopic.ts
```diff
@@ -60,4 +74,33 @@ export interface ITopic<E> extends DistributedObject {
      */
     publish(message: E): Promise<void>;
 
+    /**
+     * Publishes all messages to all subscribers of this topic.
+     *
+     * @param messages the messages to publish to all subscribers of this topic
+     * @throws {TopicOverloadError} if the consumer is too slow
```
srknzl, line 81:
```suggestion
     * @throws {@link TopicOverloadError} if the consumer is too slow
```

srknzl:
@fatihozer0 Can you update https://github.com/srknzl/hazelcast-nodejs-client/blob/9ad6b4847b3b7edce3357a6dfbdfb027c0c667c0/src/proxy/MessageListener.ts#L21. You should say ITopic not reliable topic

srknzl:
@fatihozer0 Same for https://github.com/srknzl/hazelcast-nodejs-client/blob/9ad6b4847b3b7edce3357a6dfbdfb027c0c667c0/src/proxy/MessageListener.ts#L26

src/proxy/ITopic.ts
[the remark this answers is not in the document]
srknzl: I don't think you fixed all the points in my patch. I sent more comment below about the same things accidentally but won't resolve this comment either until you apply them. I will resolve all ones including duplicates after you do.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +157,63 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+                    throw new IllegalArgumentError('Unknown overload policy');
+            }
+        } catch (e) {
+            if (e instanceof SchemaNotReplicatedError) {
+                return this.registerSchema(e.schema, e.clazz).then(() => this.publishAll(messages));
+            }
+            throw new HazelcastError(e + 'Failed to publish these messages' + messages + 'to topic' + this.name);
+        }
+        return Promise.resolve();
+    }
```
srknzl, line 207:
```suggestion
    publishAll(messages: any[]): Promise<void> {
        const capacity = messages.length;
        const reliableTopicMessages: Array<ReliableTopicMessage> = Array<ReliableTopicMessage>(capacity);

        assertNotNull(messages);
        for (const message in messages) {
            assertNotNull(message);
        }

        try {
            for (const message of messages) {
                const reliableTopicMessage = new ReliableTopicMessage();
                reliableTopicMessage.payload = this.serializationService.toData(message);
                reliableTopicMessage.publishTime = Long.fromNumber(Date.now());
                reliableTopicMessages.push(reliableTopicMessage);
            }
            switch (this.overloadPolicy) {
                case TopicOverloadPolicy.ERROR:
                    return this.addMessagesOrFail(reliableTopicMessages);
                case TopicOverloadPolicy.DISCARD_NEWEST:
                    return this.ringbuffer.addAll(reliableTopicMessages, OverflowPolicy.OVERWRITE).then(() => {});
                case TopicOverloadPolicy.DISCARD_OLDEST:
                    return this.ringbuffer.addAll(reliableTopicMessages, OverflowPolicy.FAIL).then(() => {});
                case TopicOverloadPolicy.BLOCK:
                    return this.addAndBlock(reliableTopicMessages, 100);
                default:
                    return Promise.reject(new IllegalArgumentError('Unknown overload policy'));
            }
        } catch (e) {
            if (e instanceof SchemaNotReplicatedError) {
                return this.registerSchema(e.schema, e.clazz).then(() => this.publishAll(messages));
            }
            return Promise.reject(new HazelcastError(e + 'Failed to publish these messages' + messages + 'to topic' + this.name));
        }
    }
```

Handling of promises should be like this.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -204,4 +268,22 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+                throw new TopicOverloadError(`Failed to publish messages on topic ${this.name}.`);
+            }
+        });
+    }
+
+    private addAndBlock(messages: ReliableTopicMessage[], pauseMillis: number): Promise<void> {
+        this.ringbuffer.addAll(messages, OverflowPolicy.FAIL).then((id: Long) => {
+            if (id.toNumber() === -1) {
+                setTimeout(() => {
+                    this.addAndBlock(messages, Math.min(pauseMillis * 2, 2000));
```
srknzl, line 283:
```suggestion
                    this.addAndBlock(messages, Math.min(pauseMillis * 2, MAX_BACKOFF));
```

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -204,4 +268,22 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+                }, pauseMillis);
+            }
+        });
+        return Promise.resolve();
```
srknzl, line 287:
```suggestion
   private addAndBlock(returnPromise: DeferredPromise<void>, messages: ReliableTopicMessage[], pauseMillis: number): void {
        this.ringbuffer.addAll(messages, OverflowPolicy.FAIL).then((id: Long) => {
            if (id.toNumber() === -1) {
                setTimeout(() => {
                    this.addAndBlock(returnPromise, messages, Math.min(pauseMillis * 2, 2000));
                }, pauseMillis);
            } else {
                returnPromise.resolve();
            }
        }).catch(returnPromise.reject);
    }
```


Then in the publishAll: 

```ts
publishAll(messages: any[]): Promise<void> {
        const capacity = messages.length;
        const reliableTopicMessages: Array<ReliableTopicMessage> = Array<ReliableTopicMessage>(capacity);

        assertNotNull(messages);
        for (const message of messages) {
            assertNotNull(message);
        }

        const deferred = deferredPromise<void>();

        try {
            for (const message of messages) {
                const reliableTopicMessage = new ReliableTopicMessage();
                reliableTopicMessage.payload = this.serializationService.toData(message);
                reliableTopicMessage.publishTime = Long.fromNumber(Date.now());
                reliableTopicMessages.push(reliableTopicMessage);
            }
            switch (this.overloadPolicy) {
                case TopicOverloadPolicy.ERROR:
                    return this.addMessagesOrFail(reliableTopicMessages);
                case TopicOverloadPolicy.DISCARD_NEWEST:
                    return this.ringbuffer.addAll(reliableTopicMessages, OverflowPolicy.OVERWRITE).then(() => {});
                case TopicOverloadPolicy.DISCARD_OLDEST:
                    return this.ringbuffer.addAll(reliableTopicMessages, OverflowPolicy.FAIL).then(() => {});
                case TopicOverloadPolicy.BLOCK:
                    this.addAndBlock(deferred, reliableTopicMessages, 100);
                    break;
                default:
                    return Promise.reject(new IllegalArgumentError('Unknown overload policy'));
            }
        } catch (e) {
            if (e instanceof SchemaNotReplicatedError) {
                return this.registerSchema(e.schema, e.clazz).then(() => this.publishAll(messages));
            }
            return Promise.reject(new HazelcastError(e + 'Failed to publish these messages' + messages + 'to topic' + this.name));
        }
        return deferred.promise;
    }
```

src/proxy/topic/ReliableTopicProxy.ts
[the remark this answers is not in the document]
srknzl: java uses null for all cases, but we use connection address from a random connection.  Can you remove localAddress variable and use null all the time. Not undefined, but null please.

src/proxy/topic/ReliableTopicProxy.ts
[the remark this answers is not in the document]
srknzl: You printed the overload policy in your latest change but it is just a number. I don't think we should do it as java did not either.

src/proxy/topic/ReliableTopicProxy.ts
srknzl: Can you add the HazelcastError that can be thrown from ReliableTopic to ITopic interface? You can say that HazelcastError may be thrown (you should give a link to HazelcastError) when the method is invoked from a reliable topic and the messages cannot be published.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +157,63 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+                    });
+                    break;
+                default:
+                    throw new IllegalArgumentError('Unknown overload policy');
+            }
+        } catch (e) {
+            if (e instanceof SchemaNotReplicatedError) {
+                return this.registerSchema(e.schema, e.clazz).then(() => this.publishAll(messages));
+            }
+            throw new HazelcastError(e + 'Failed to publish these messages' + messages + 'to topic' + this.name);
```
srknzl, line 204:
```suggestion
            throw new HazelcastError(e + 'Failed to publish these messages ' + messages + ' to topic ' + this.name);
```

src/proxy/topic/TopicProxy.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    publishAll(messages: any[]): Promise<void> {
+        assertNotNull(messages);
+        for (const message of messages) {
+            assertNotNull(message);
+        }
+        const messageDataList = this.toData(messages);
+        return this.encodeInvoke(TopicPublishAllCodec, () => {}, () => {
+        }, messageDataList);
```
srknzl, line 64:
```suggestion
        return this.encodeInvoke(TopicPublishAllCodec, () => {}, () => {}, messageDataList);
```

srknzl:
General idea about how many tests you should add: 

You added publishAll to both proxies. You added addListener and removeListener to all proxies as well. Basic tests about their basic functionality would be good. 


Also, we disallow addMessageListener and removeMessageListener methods from TopicProxy. You need to write a test about the exception throw behavior when these methods are called from TopicProxy.

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
```diff
@@ -189,4 +189,38 @@ describe('ReliableTopicTest', function () {
         const obj = clientOne.getSerializationService().toObject(item.payload);
         expect(obj).to.equal(11);
     });
+
+    it('writes and reads messages using async methods', async function () {
+        const topic = await clientOne.getReliableTopic('async');
+
+        await topic.addListener(() => {});
```
srknzl, line 196: why add a listener here that is not used?

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
```diff
@@ -189,4 +189,38 @@ describe('ReliableTopicTest', function () {
...
+
+        await topic.publish('test');
+
+        const ringbuffer = topic.getRingbuffer();
+        const seq = await ringbuffer.tailSequence();
+        const item = await ringbuffer.readOne(seq);
```
srknzl, line 202: This test logic exists with name `writes and reads messages`

test/integration/backward_compatible/parallel/topic/TopicTest.js
```diff
@@ -0,0 +1,163 @@
...
+const RC = require('../../../RC');
+
+describe('TopicTest', function () {
+    let client;
+    let topic;
+
+    const testFactory = new TestUtil.TestFactory();
+
+    before(async function () {
+        const cluster = await testFactory.createClusterForParallelTests();
```
srknzl, line 29: We need to skip this test if the client version is not more than or equal to 5.3 because this test is a backward compatible test. See TestUtil.markClientVersionAtLeast method. Use "5.3" version.

test/integration/backward_compatible/parallel/topic/TopicTest.js
```diff
@@ -0,0 +1,163 @@
...
+            onMessage(event) {
+                event.publish_time = Date.now();
+                this.events.push(event);
+            }
+        }
+
+        const collector = new Collector();
+
+        setTimeout(() => {
+            expect(collector.events.length).to.be.equal(1);
```
srknzl, line 59: What will add events to the collector?

test/integration/backward_compatible/parallel/topic/TopicTest.js
```diff
@@ -0,0 +1,163 @@
...
+        await topic.publish(message);
+
+        await new Promise((resolve) => {
+            const interval = setInterval(() => {
+                if (count === 1) {
+                    clearInterval(interval);
+                    expect(receivedValues).to.deep.equal([message]);
+                    resolve();
+                }
+            }, 100);
```
srknzl, line 106: TestUtil.assertTrueEventually maybe?

test/integration/backward_compatible/parallel/topic/TopicTest.js
```diff
@@ -0,0 +1,163 @@
...
+                    clearInterval(interval);
+                    expect(receivedValues).to.have.members(messages);
+                    resolve();
+                }
+            }, 100);
+            setTimeout(() => {
+                clearInterval(interval);
+                reject(new Error('Timed out while waiting for messages to be published'));
+            }, 5000);
+        });
```
srknzl, line 134: Again we have assertTrueEventually methods for this kind of logic.

test/integration/backward_compatible/parallel/topic/TopicTest.js
```diff
@@ -0,0 +1,163 @@
...
+
+        await topic.addListener((message) => {
+            count++;
+            receivedValues.push(message.messageObject);
+        });
+
+        const messages = [1, null, 3];
+        await topic.publishAll(messages);
+
+        await new Promise((resolve, reject) => {
```
srknzl, line 149: same here.

test/integration/backward_compatible/parallel/topic/TopicTest.js
```diff
@@ -0,0 +1,163 @@
...
+                }
+            }, 100);
+            setTimeout(() => {
+                clearInterval(interval);
+                reject(new Error('Timed out while waiting for messages to be published'));
+            }, 5000);
+        });
+    });
+
+    it('tests publishAll with null', async function() {
```
srknzl, line 137: Can you try publishAll with null value array as well?

test/integration/backward_compatible/parallel/topic/TopicTest.js
```diff
@@ -0,0 +1,163 @@
...
+        });
+
+        const messages = [1, null, 3];
+        await topic.publishAll(messages);
+
+        await new Promise((resolve, reject) => {
+            const interval = setInterval(() => {
+                if (count === messages.length) {
+                    clearInterval(interval);
+                    expect(receivedValues).to.have.members(messages);
```
srknzl, line 153: This test should throw as I know

srknzl (commented):
We are getting there, thanks for the changes but we need more changes. Keep up the good work

src/proxy/ITopic.ts
fatihozer0: Done.

src/proxy/ITopic.ts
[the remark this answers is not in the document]
fatihozer0: Okay.

src/proxy/ITopic.ts
fatihozer0: `{@link HazelcastError}` does not work like this. This is why I implemented it as `{HazelcastError}` . Others are done.

src/proxy/ITopic.ts
fatihozer0: Same as above. deprecated done but links are the same.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +157,63 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
         }
     }
 
+    publishAll(messages: any[]): Promise<void> {
+        const capacity = messages.length;
+        const reliableTopicMessages: Array<ReliableTopicMessage> = Array<ReliableTopicMessage>(capacity);
+
+        assertNotNull(messages);
+        for (const message in messages) {
```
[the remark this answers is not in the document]
fatihozer0: I see. Just missed it sorry.

src/proxy/topic/ReliableTopicProxy.ts
```diff
@@ -150,6 +157,63 @@ export class ReliableTopicProxy<E> extends BaseProxy implements ITopic<E> {
...
+            assertNotNull(message);
+        }
+
+        try {
+            for (const message of messages) {
+                const reliableTopicMessage = new ReliableTopicMessage();
+                reliableTopicMessage.payload = this.serializationService.toData(message);
+                reliableTopicMessage.publishTime = Long.fromNumber(Date.now());
+                reliableTopicMessages.push(reliableTopicMessage);
```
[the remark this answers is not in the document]
fatihozer0: Done.

src/proxy/topic/ReliableTopicProxy.ts
fatihozer0: I see. Done

src/proxy/topic/ReliableTopicProxy.ts
fatihozer0: Done.

src/proxy/topic/ReliableTopicProxy.ts
fatihozer0: Done.

src/proxy/topic/ReliableTopicProxy.ts
fatihozer0: I have to improve myself about promises.

src/proxy/topic/TopicProxy.ts
fatihozer0: I see. Done.

src/proxy/topic/TopicProxy.ts
fatihozer0: Done

src/proxy/topic/TopicProxy.ts
fatihozer0: Okay, done. Why is this better

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
fatihozer0: There was something in my head but not implemented and missed. Done.

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
fatihozer0: The reason is same as above.

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: Okay, done.

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
```diff
@@ -189,4 +189,38 @@ describe('ReliableTopicTest', function () {
...
+        const obj = clientOne.getSerializationService().toObject(item.payload);
+        expect(obj).to.equal('test');
+    });
+
+    it('should not receive messages after listener is removed by async removeListener', async function () {
```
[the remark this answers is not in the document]
fatihozer0: Implemented before deprecating old methods (while both exists). Removed, done

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: How does it help?

fatihozer0: I was not aware of it. Converted to it.

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: Okay, I see

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: Okay.

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: Okay.

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: Yes, fixed.

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: Is it better to do this using an array function in expect or using try/catch block

src/proxy/topic/TopicProxy.ts
srknzl: as far as I remember, this was a method of an instance which has name field. and we were passing it as a parameter. We don't need it, therefore it is better.

srknzl: Do you wonder why we assign this.name to name? Because this refers to the object we return inside the object we return.

src/proxy/ITopic.ts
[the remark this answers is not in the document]
srknzl: I applied my suggestions in a commit separately.

src/proxy/ITopic.ts
srknzl: No, it works.

src/proxy/topic/ReliableTopicProxy.ts
srknzl: 🤔 are you sure? I will do this myself as well.

test/integration/backward_compatible/parallel/topic/TopicTest.js
srknzl: It is currently commented out. Changing it.

test/integration/backward_compatible/parallel/topic/TopicTest.js
srknzl: It's fine, you could have used Long.js methods as well.

test/integration/backward_compatible/parallel/topic/TopicTest.js
srknzl: if you don't need a try catch, don't use it

src/proxy/topic/TopicProxy.ts
srknzl, line 1: Where are the tests mentioned in TDD ? 

- testListener()
- testRemoveListener()
- testPublishAsync()
- testPublishAllAsync()
- testPublishAllException()
- ClientTopicNullTest()

They should have the exact suite and test names.

test/integration/backward_compatible/parallel/topic/TopicTest.js
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    it('tests publishAll with null array', async function() {
+        const messages = null;
+        await expect(() => topic.publishAll(messages)).to.throw('Non null value expected.');
+    });
+
+    it('tests publishAll with null elements array', async function() {
+        const messages = [null, null, null];
+        await expect(() => topic.publishAll(messages)).to.throw('Non null value expected.');
```
srknzl, line 144: You should not need this async/await in these tests.

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
srknzl, line 1: I wonder where you got these tests from. Also, addListener, removeListener methods are new in ReliableTopic. You should test them as well.

test/integration/backward_compatible/parallel/topic/TopicTest.js
srknzl, line 1: Where are the tests mentioned in TDD ? 

- testListener()
- testRemoveListener()
- testPublishAsync()
- testPublishAllAsync()
- testPublishAllException()
- ClientTopicNullTest()

They should have the exact suite and test names. 


which means there should be describe block called `ClientTopicTest` and tests with names `testListener`, `testRemoveListener` etc.

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: Right, fixed

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
fatihozer0: I got them from Python Client. Is it problematic?

test/integration/backward_compatible/parallel/topic/TopicTest.js
fatihozer0: But we have some differences from Java Client in other tests. This is why I named them different. Is the only problem about names or about logic? I think I implemented all of them.

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
fatihozer0: I added tests with new methods by just changing deprecated methods name. Is it okay?

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
```diff
@@ -93,6 +99,20 @@ describe('ReliableTopicTest', function () {
         }).catch(done);
     });
 
+    it('writes and reads messages with addListener method', async function () {
+        const topicName = 't' + Math.random();
```
srknzl, line 103:
```suggestion
        const topicName = TestUtil.randomString(8);
```

srknzl: even other tests use Math.random() I don't think it is elegant.

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
```diff
@@ -93,6 +99,20 @@ describe('ReliableTopicTest', function () {
...
+        const topicOne = await clientOne.getReliableTopic(topicName);
+        const topicTwo = await clientTwo.getReliableTopic(topicName);
+        await topicTwo.addListener(async (msg) => {
+            if (msg.messageObject['value'] !== 'foo') {
+                throw new Error('Message received does not match expected value.');
+            }
+        });
```
srknzl, line 110: This test does not wait for the listener to run. It is not a valid test. You can create a deferred promise and await it in the test while resolving it in the addListener callback. Alternatively, you can use done syntax(which I don't recommend)

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
```diff
@@ -123,8 +143,35 @@ describe('ReliableTopicTest', function () {
...
+        setTimeout(() => {
+            topicTwo.removeListener(id);
+            topicOne.publish({ 'value2': 'foo2' });
+            topicOne.publish({ 'value3': 'foo3' });
+            topicOne.publish({ 'value4': 'foo4' });
+            topicOne.publish({ 'value5': 'foo5' });
+            setTimeout(() => {
+                // Test passes
+            }, 500);
+        }, 500);
```
srknzl, line 170: Instead of this you should do `await TestUtil.promiseWaitMilliseconds(500);`. With setTimeout, it is async and we don't wait for it.

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
```diff
@@ -123,8 +143,35 @@ describe('ReliableTopicTest', function () {
         }).catch(done);
     });
 
+    it('removed message listener does not receive items after removal with new added methods', async function() {
```
srknzl, line 146:
```suggestion
    it('removed message listener does not receive items after removal with removeListener', async function() {
```

test/integration/backward_compatible/parallel/topic/ReliableTopicTest.js
srknzl: You should implement the tests using the Java client tests. You can search for the tests in ClientReliableTopicOverloadTest that uses publishAll method and implement them.

PR #1369, opened 2022-09-19
Add an example code for `map.removeAll` method [API-15]

fatihozer0:


code_samples/map_removeAll_sample.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
```
srknzl, line 1: Can you add copyright? See other files

code_samples/map_removeAll_sample.js
```diff
@@ -0,0 +1,24 @@
...
+
+const { Client, Predicates } = require('../lib/index');
```
srknzl, line 3:
```suggestion
const { Client, Predicates } = require('hazelcast-client');
```

srknzl (changes requested):
some minor commetns are added

code_samples/map_removeAll_sample.js
fatihozer0: I would add it but could not be sure should I copy/paste the copyright or is there a way to do it?

code_samples/map_removeAll_sample.js
fatihozer0: Yes, my bad. Just wanted to see if it works

code_samples/map_removeAll_sample.js
srknzl: Just copy paste

srknzl (approved):

PR #1393, opened 2022-10-08
Add filter argument to Ringbuffer interface [API-1294]

fatihozer0:
I am not sure if anything is right but can not say why it is wrong. I tried this on some other codes to try if it works and worked but not sure.

### DIFF
--- a/src/proxy/Ringbuffer.ts
+++ b/src/proxy/Ringbuffer.ts
@@ -156,9 +156,18 @@ export interface Ringbuffer<E> extends DistributedObject {
      * until further items become available, and it can read at least the
      * minimum number of items.
      *
+     * `filter` argument is provided to select only the items that are needed
+     * to be read. If the filter is not provided, all items will be read.
+     * Otherwise, only the items where the filter function returns true are returned.
+     * Using filters is a good way to prevent getting items that are of no value
+     * to the receiver. amount of IO and the number of operations being executed
+     * and can result in a significant performance improvement. Note that,
+     * filtering logic must be defined on the server-side.
+     *
      * @param startSequence sequence number of the first item to be read.
      * @param minCount minimum number of items to be read.
      * @param maxCount maximum number of items to be read.
+     * @param filter  optional filter to be applied to the items
      * @throws RangeError if `startSequence` is smaller than `0`,
      *                      or if `minCount` smaller than `0`,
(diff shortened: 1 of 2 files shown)


srknzl:
> I am not sure if anything is right but can not say why it is wrong. 

😄 What? 

Looks good, thanks

srknzl (dismissed):

yuce (changes requested):
Could you port Java client's tests and add documentation and a code sample?

srknzl:
Adding a test may not be possible because server side code is needed. I don't know if we have a filter in the test jar. A code sample requires users to write java code in the server side. Not good.

You are right about the documentation. API documentation can be added.  Also there can be a small section about the filter in section https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/DOCUMENTATION.md#847-using-ringbuffer

yuce:
@srknzl The python client has two tests which uses the filter: https://github.com/hazelcast/hazelcast-python-client/blob/b8dd0d737ef976922311a8241e2a51b50e7ab3fa/tests/integration/backward_compatible/proxy/ringbuffer_test.py#L230 Could you guys check out them to see whether they can be ported to the Node.js client?

srknzl:
We can do and we should. It is in the test jar. 

https://github.com/hazelcast/hazelcast/blob/14ef0dbefa4a6fa00a50a8873aeea02ec6c3f5e4/hazelcast/src/test/java/com/hazelcast/client/test/PrefixFilter.java#L26

fatihozer0:
Added a test which is not ported. I checked the links you shared. Gonna add documentations now

src/proxy/Ringbuffer.ts
```diff
@@ -164,5 +164,5 @@ export interface Ringbuffer<E> extends DistributedObject {
      *                      or if `minCount` larger than `maxCount`,
      *                      or if `maxCount` larger than `1000` (to prevent overloading)
      */
-    readMany(startSequence: number | Long, minCount: number, maxCount: number): Promise<ReadResultSet<E>>;
```
srknzl, line 167: Can you add the API documentation?

fatihozer0: Done.

srknzl (approved):

yuce (approved):

fatihozer0:
Continues on #1413

PR #1398, opened 2022-10-11
Add skipped check before asserting in CompactPublicAPIsTest.afterEach

srknzl:


test/integration/backward_compatible/parallel/serialization/compact/CompactPublicAPIsTest.js
```diff
@@ -405,6 +409,7 @@ describe('CompactPublicAPIsTest', function () {
 
         it('removeAll', async function () {
             TestUtil.markClientVersionAtLeast(this, '5.2.0');
+            skipped = true;
```
mdumandag, line 412: should we set it no matter what?

mdumandag (commented):
Also, there is one more conditional test here in this file

test/integration/backward_compatible/parallel/serialization/compact/CompactPublicAPIsTest.js
srknzl: right, sorry

srknzl:
> Also, there is one more conditional test here in this file

The other checks are in `before`. They won't cause problems because `afterEach` won't run if whole test suite is skipped. Anyway, assigning `skipped` in those as well.

test/integration/backward_compatible/parallel/serialization/compact/CompactPublicAPIsTest.js
mdumandag: I meant, shouldn't we do something similar to others and set this in an if block, if the client version is older and skip the test?

mdumandag: Also, we are using a single variable here, I don't know if it is shared or not. Can this be a problem in parallel tests?

srknzl: No it won't be a problem because tests that are in each file are run in a individual processes.

srknzl: Changed to if block

mdumandag (approved):

mdumandag:
Merging this to restart release tests

PR #1399, opened 2022-10-13
Make ReadResultSet iterable [API-1315]

harunalpak:
Following changes done:
1. ReadResultSet interface was changed.
2. Implemented LazyReadResultSet as iterable
3. Added new test for testing LazyReadResultSet is iterable.
4. Changed documentation of Ringbuffer by adding readMany sample.

### DIFF
--- a/src/core/ReadResultSet.ts
+++ b/src/core/ReadResultSet.ts
@@ -62,6 +62,20 @@ export interface ReadResultSet<T> {
     size(): number;
 
     /**
+     * Returns an iterator for elements in the list.
+     * 
+     * @returns the iterator for elements in the list.
+     */
+     values(): Iterator<T>;
+
+     /**
+     * Returns an iterator for elements in the list.
+     * 
+     * @returns the iterator for elements in the list.
+     */
+    [Symbol.iterator](): Iterator<T>; 
+
+    /**
      * Returns the sequence of the item following the last read item. This
      * sequence can then be used to read items following the ones returned by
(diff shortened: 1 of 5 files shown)


src/core/ReadResultSet.ts
```diff
@@ -61,6 +61,20 @@ export interface ReadResultSet<T> {
      */
     size(): number;
 
+    /**
+     * Returns an iterator for elements in the list.
+     * 
+     * @returns the iterator for elements in the list.
+     */
+     values(): Iterator<T>;
```
srknzl, line 69: Why do we need this?

test/unit/ringbuffer/LazyReadResultSetTest.js
```diff
@@ -42,4 +42,14 @@ describe('LazyReadResultSetTest', function () {
         const set = new LazyReadResultSet(mockSerializationService, 4, [1, 2, 3, 4], [11, 12, 13, 14], 15);
         expect(set.get(4)).to.be.undefined;
     });
+
+    it('object can be iteratable', function () {
```
srknzl, line 46:
```suggestion
    it('should be iterable', function () {
```

"it should be iterable" is fine I think.

src/core/ReadResultSet.ts
harunalpak: You are right. We can remove it.

srknzl (approved):

PR #1401, opened 2022-10-14
Add `filter` argument information to DOCUMENTATION.md [API-1294]

fatihozer0:
Documentation update on Master Branch about RingBuffer Filter argument.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1902,6 +1902,12 @@ added and the head is where the items are overwritten or expired. You can reach 
 which is mapped to the elements between the head and tail (inclusive) of the Ringbuffer. For details, see the
 [Ringbuffer section](https://docs.hazelcast.com/hazelcast/latest/data-structures/ringbuffer) in the Hazelcast Reference Manual.
 
+`filter` argument is provided for to only select items that need to be read. If
+the filter is not provided, all items will be read. Otherwise, only items where the filter function returns true are
+returned. Using  filters is a good way to prevent getting items that are of no value to the receiver. This reduces the
+amount of IO and the number of operations being executed, and can result in a significant performance improvement. Note that,
+filtering logic must be defined on the server-side.
+
 A Ringbuffer usage example is shown below.
 
 ```javascript


DOCUMENTATION.md
```diff
@@ -1902,6 +1902,12 @@ added and the head is where the items are overwritten or expired. You can reach
 which is mapped to the elements between the head and tail (inclusive) of the Ringbuffer. For details, see the
 [Ringbuffer section](https://docs.hazelcast.com/hazelcast/latest/data-structures/ringbuffer) in the Hazelcast Reference Manual.
 
+`filter` argument is provided for to only select items that need to be read. If
```
srknzl, line 1905: I don't remember if I said to add some documentation here but I think this type of documentation will be better in the API documentation.

DOCUMENTATION.md
```diff
@@ -1902,6 +1902,12 @@ added and the head is where the items are overwritten or expired. You can reach
...
+the filter is not provided, all items will be read. Otherwise, only items where the filter function returns true are
```
srknzl, line 1906:
```suggestion
the filter is not provided, all items will be read. Otherwise, only the items where the filter function returns true are
```

DOCUMENTATION.md
```diff
@@ -1902,6 +1902,12 @@ added and the head is where the items are overwritten or expired. You can reach
...
+returned. Using  filters is a good way to prevent getting items that are of no value to the receiver. This reduces the
```
srknzl, line 1907:
```suggestion
returned. Using filters is a good way to prevent getting items that are of no value to the receiver. This reduces the
```

DOCUMENTATION.md
```diff
@@ -1902,6 +1902,12 @@ added and the head is where the items are overwritten or expired. You can reach
...
+amount of IO and the number of operations being executed, and can result in a significant performance improvement. Note that,
```
srknzl, line 1908:
```suggestion
amount of IO and the number of operations being executed and can result in a significant performance improvement. Note that,
```

DOCUMENTATION.md
srknzl: You can add a short usage here or to the code samples and give a link to it.



something like:

You can use `Ringbuffer.readMany` function to read more than one item. You can use the `filter` method to filter out some items if you want. The filter class have to be in the class path of members. For an example usage check out this(add a link to `this`) code sample.

fatihozer0:
PR is closed due to misunderstanding the comments of #1393

PR #1403, opened 2022-10-24
Change doTryNextCluster logic to make it the same. [API-1354]

harunalpak:
Before changes, client try only maxTryCount.
After code changes, client will try (candidateClusters.length * maxTryCount). For each cluster we will try maxTryCount.


srknzl (approved):
looking good

PR #1415, opened 2022-11-17
Send client state to cluster in re-connections[API-1644]

harunalpak:
Related PR: https://github.com/hazelcast/hazelcast/pull/22164


srknzl (changes requested):
Should we add a test? We cannot add split brain tests for obvious reasons but, can we add a test that will check if state is sent to the cluster after a reconnect? (could be in ClientReconnectTest, via using sinon spies/fakes)

harunalpak:
> Should we add a test? We cannot add split brain tests for obvious reasons but, can we add a test that will check if state is sent to the cluster after a reconnect? (could be in ClientReconnectTest, via using sinon spies/fakes)

Yes we should add this test to ensure that whether the state sent to cluster or not?

test/integration/backward_compatible/serial/ClientReconnectTest.js
```diff
@@ -59,6 +40,37 @@ describe('ClientReconnectTest', function () {
...
+        cluster = await testFactory.createClusterForSerialTests();
+        const member = await RC.startMember(cluster.id);
+        client = await testFactory.newHazelcastClientForSerialTests({
+            clusterName: cluster.id,
+            properties: {
+                'hazelcast.client.heartbeat.interval': 1000,
+                'hazelcast.client.heartbeat.timeout': 3000
+            }
+        });
+        await RC.terminateMember(cluster.id, member.uuid);
```
srknzl, line 59:
```suggestion
        await TestUtil.waitForConnectionCount(client, 1);
        await RC.terminateMember(cluster.id, member.uuid);
```

srknzl (approved):

test/integration/backward_compatible/serial/ClientReconnectTest.js
```diff
@@ -59,6 +40,37 @@ describe('ClientReconnectTest', function () {
         await testFactory.shutdownAll();
     });
 
+    it('should send the client state to the cluster after reconnections, ' +
+        +'regardless it is connected back to possibly the same cluster with the same id or not.', async function () {
+        const fakeInitializeClientOnCluster = sandbox.replace(
```
mdumandag, line 45: Just a note: This test will probably fail in backward compatibility tests as there is no client version filter

srknzl: @harunalpak I overlooked this, can you add a version check?

PR #1417, opened 2022-11-17
Fix pipelined writer API-1283

### LINKED ISSUE #1256: The error: `Cannot add property 0, object is not extensible`
This happened in several tests, the main cause is writing to a connection's writer after it is closed. I could not reproduced the issue, so we don't know when it happens. As the fix we can add a try catch and ignore the writes after `connection.close()`.

See #1199 and #1230

srknzl:
fixes #1256 

The issue was writing some data to the connection after it being closed. I changed and refactored some logic around writer and connection to avoid that.

### DIFF
--- a/src/network/Connection.ts
+++ b/src/network/Connection.ts
@@ -96,7 +96,8 @@ export class PipelinedWriter extends Writer {
         this.closeReason = this.makeIOError(closeReason);
         this.closed = true;
         this.canWrite = false;
-        this.socket.destroy(closeReason);
+        // If we pass a error to destroy, an unhandled error will be thrown because we don't handle the error event
+        this.socket.destroy();
         // no more items can be added now
         this.queue = FROZEN_ARRAY;
     }
--- a/src/network/Connection.ts
+++ b/src/network/Connection.ts
@@ -41,7 +41,7 @@ abstract class Writer extends EventEmitter {
 
     abstract write(message: ClientMessage, resolver: DeferredPromise<void>): void;
 
-    abstract close(): void;
+    abstract close(cause?: Error): void;
 
 }
 
@@ -57,9 +57,10 @@ interface OutputQueueItem {
 export class PipelinedWriter extends Writer {
 
     private queue: OutputQueueItem[] = [];
-    private error: Error;
     private scheduled = false;
     private canWrite = true;
+    private closed = false;
+    private closeReason: Error;
     // reusable buffer for coalescing
     private readonly coalesceBuf: Buffer;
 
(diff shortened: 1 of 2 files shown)


test/unit/connection/PipelinedWriterTest.js
```diff
@@ -199,16 +201,17 @@ describe('PipelinedWriterTest', function () {
...
 
-    it('rejects single promise on write failure', function(done) {
+    it('rejects single promise on write failure', async function() {
         const err = new Error();
         setUpWriteFailure(err);
 
         const resolver = deferredPromise();
         writer.write(createMessageFromString('test'), resolver);
-        resolver.promise.catch((err) => {
-            expect(err).to.be.equal(err);
```
srknzl, line 209: This was silly, basically making 1 === 1 assertion. It does not work if I rename one of the variables because we wrap the err in IOError. (in a defected way we pass error as message, fixed that as well.)

src/network/Connection.ts
```diff
@@ -164,15 +175,22 @@ export class PipelinedWriter extends Writer {
...
-        this.error = new IOError(err);
+    private handleError(err: Error, sentResolvers: OutputQueueItem[]): void {
+        const error = this.makeIOError(err);
         for (const item of sentResolvers) {
-            item.resolver.reject(this.error);
+            item.resolver.reject(error);
         }
         for (const item of this.queue) {
-            item.resolver.reject(this.error);
+            item.resolver.reject(error);
```
srknzl, line 184: This line is added but not covered, and this is normal. Actually there was no test before my change as well. I did not change the logic of here. Just added this.close() to the end of this function. I don't want to add a test for this because it is not trivial and requires some magic with sinon.

src/network/Connection.ts
```diff
@@ -41,7 +41,7 @@ abstract class Writer extends EventEmitter {
 
     abstract write(message: ClientMessage, resolver: DeferredPromise<void>): void;
 
-    abstract close(): void;
+    abstract close(cause: Error | null): void;
```
mdumandag, line 44: When can this be null?

mdumandag: And, if this can't be null, then I guess we won't need the `closed` boolean anymore

srknzl: Right, we check if it is null and create a new error if it is null in `Connection.close`. Let me try to get rid of `closed` boolean as well.

srknzl: Removed `closed` boolean. Add `isClosed()` method just for readability.

src/network/Connection.ts
```diff
@@ -79,17 +79,28 @@ export class PipelinedWriter extends Writer {
         });
     }
 
+    private isClosed(): boolean {
```
mdumandag, line 82: Not sure how things are optimized in Node.js but are method calls expensive here? I mean, if they are, I think the old way of doing things (with the name `error` instead of `closeCause`) seems OK to me

srknzl: Ok removed the method and renamed the variable

mdumandag (dismissed):

srknzl:
The coverage drop is shown as 0.16% but I see that the drop is due to SSL tests because the drop is in `connectTLSSocket`. I think it should be related to the ssl tests being disabled.

src/network/Connection.ts
```diff
@@ -200,7 +214,7 @@ export class DirectWriter extends Writer {
     }
 
     close(): void {
```
harunalpak, line 216: I am not sure but this method different syntax than abstract method? Do we need to change this method also?

harunalpak (commented):
I reviewed all codes and tests, it seems good. I could not find any problem. I have only one question that I am not sure. After response I can approve.

src/network/Connection.ts
srknzl: we don't need to, because the parameter will be unused. It does not matter. I added it.

harunalpak (approved):

PR #1418, opened 2022-11-23
Check the member UUIDs that the schema is replicated in the client schema service [API-1646]

harunalpak:
Related java issue: https://github.com/hazelcast/hazelcast/pull/22197

One of the test failed after code changes. I skipped this test. I will find a solution for test.


src/HazelcastClient.ts
```diff
@@ -146,7 +146,12 @@ export class HazelcastClient {
         this.loggingService = new LoggingService(this.config.customLogger,
             this.config.properties['hazelcast.logging.level'] as string);
         const logger = this.loggingService.getLogger();
-        this.schemaService = new SchemaService(() => this.invocationService, logger);
+        this.schemaService = new SchemaService(
```
srknzl, line 149: You can move clusterService and clusterFailoverService construction above and use clusterService directly: 

```diff
diff --git a/src/HazelcastClient.ts b/src/HazelcastClient.ts
index a39d6eb5..9d35a447 100644
--- a/src/HazelcastClient.ts
+++ b/src/HazelcastClient.ts
@@ -146,10 +146,16 @@ export class HazelcastClient {
         this.loggingService = new LoggingService(this.config.customLogger,
             this.config.properties['hazelcast.logging.level'] as string);
         const logger = this.loggingService.getLogger();
+        this.clusterFailoverService = this.initClusterFailoverService();
+        this.clusterService = new ClusterService(
+            this.config,
+            logger,
+            this.clusterFailoverService
+        );
         this.schemaService = new SchemaService(
-            this.config, 
-            () => this.clusterService, 
-            () => this.invocationService, 
+            this.config,
+            this.clusterService,
+            () => this.invocationService,
             logger
         );
         this.serializationService = new SerializationServiceV1(this.config.serialization, this.schemaService);
@@ -166,12 +172,6 @@ export class HazelcastClient {
             this.config.lifecycleListeners,
             logger
         );
-        this.clusterFailoverService = this.initClusterFailoverService();
-        this.clusterService = new ClusterService(
-            this.config,
-            logger,
-            this.clusterFailoverService
-        );
         this.connectionRegistry = new ConnectionRegistryImpl(
             this.config.connectionStrategy.asyncStart,
             this.config.connectionStrategy.reconnectMode,

```

src/config/Config.ts
```diff
@@ -178,6 +178,7 @@ const DEFAULT_PROPERTIES: Properties = {
     'hazelcast.client.heartbeat.interval': 5000,
     'hazelcast.client.heartbeat.timeout': 60000,
     'hazelcast.client.invocation.retry.pause.millis': 1000,
+    'hazelcast.client.schema.max.put.retry.count': 100,
```
srknzl, line 181: Can you update ConfigBuilder.validateProperty and ConfigBuilderTest as well? We should add validation and test for validation. 

@mdumandag Should we document this property? Is it public?

src/proxy/Ringbuffer.ts
```diff
@@ -156,13 +156,22 @@ export interface Ringbuffer<E> extends DistributedObject {
...
      * @param minCount minimum number of items to be read.
      * @param maxCount maximum number of items to be read.
+     * @param filter optional filter to be applied to the items
      * @throws RangeError if `startSequence` is smaller than `0`,
      *                      or if `minCount` smaller than `0`,
      *                      or if `minCount` larger than `maxCount`,
      *                      or if `maxCount` larger than `1000` (to prevent overloading)
      */
-    readMany(startSequence: number | Long, minCount: number, maxCount: number): Promise<ReadResultSet<E>>;
+    readMany(startSequence: number | Long, minCount: number, maxCount: number, filter?: any): Promise<ReadResultSet<E>>;
```
srknzl, line 176: Is this change related?

test/integration/backward_compatible/parallel/ringbuffer/RingbufferProxyTest.js
```diff
@@ -158,4 +158,13 @@ describe('RingbufferProxyTest', function () {
         const size = await rb.size();
         expect(size.toNumber()).to.equal(2);
     });
+
+    it('readMany correctly works with a filter', async function () {
```
srknzl, line 162: Is this change related?

src/serialization/compact/SchemaService.ts
```diff
@@ -106,4 +129,39 @@ export class SchemaService {
...
         return this.getInvocationService().invokeUrgent(invocation).then(() => {});
     }
+
+    async replicateSchemaInCluster(schema: Schema): Promise<boolean> {
+        let clientMessage = ClientSendSchemaCodec.encodeRequest(schema);
+        outer:
+        for (let index = 0; index < this.maxPutRetryCount; index++) {
+            const invocation = new Invocation(this.getInvocationService(), clientMessage);
+            const response = await this.getInvocationService().invoke(invocation);
+            const replicatedMemberUuids = UuidUtil.convertUUIDSetToStringSet(ClientSendSchemaCodec.decodeResponse(response));
```
srknzl, line 139: Why do we convert UUID set to a string set?

src/serialization/compact/SchemaService.ts
```diff
@@ -67,18 +82,26 @@ export class SchemaService {
...
+                `The schema ${schema} cannot be replicated in the cluster, after ${this.maxPutRetryCount}  retries. 
+                It might be the case that the client is connected to the two halves of the cluster that 
+                is experiencing a split-brain, and continue putting the data associated with that schema might 
+                result in data loss. It might be possible to replicate the schema after some time, when 
+                the cluster is healed.`
+            );
+        } else {
             this.putIfAbsent(schema);
-        });
+        }
```
srknzl, line 104:
```suggestion
        }
```

src/serialization/compact/SchemaService.ts
```diff
@@ -67,18 +82,26 @@ export class SchemaService {
     /**
      * Puts the schema with the given id to the cluster.
      */
-    put(schema: Schema): Promise<void> {
+    async put(schema: Schema): Promise<void> {
```
srknzl, line 85: Please use then/catch as async/await has more overhead.

src/serialization/compact/SchemaService.ts
```diff
@@ -67,18 +82,26 @@ export class SchemaService {
...
             return Promise.resolve();
         }
-        const message = ClientSendSchemaCodec.encodeRequest(schema);
-        const invocation = new Invocation(this.getInvocationService(), message);
-        return this.getInvocationService().invoke(invocation).then(() => {
+
+        const result = await this.replicateSchemaInCluster(schema);
+        if (!result) {
+            throw new IllegalStateError(
+                `The schema ${schema} cannot be replicated in the cluster, after ${this.maxPutRetryCount}  retries.
```
srknzl, line 96: schema.toString does not exist, and this won't look good. What about "the schema with typename schema.typeName"?

src/serialization/compact/SchemaService.ts
```diff
@@ -106,4 +129,39 @@ export class SchemaService {
         const invocation = new Invocation(this.getInvocationService(), message);
         return this.getInvocationService().invokeUrgent(invocation).then(() => {});
     }
+
+    async replicateSchemaInCluster(schema: Schema): Promise<boolean> {
```
srknzl, line 133: Without async/await please

srknzl (changes requested):

src/HazelcastClient.ts
harunalpak: Fixed

test/integration/backward_compatible/parallel/ringbuffer/RingbufferProxyTest.js
harunalpak: Not related. I dont know, how it was included in this PR? I will fix it.

src/config/Config.ts
harunalpak: Fixed

src/proxy/Ringbuffer.ts
harunalpak: Not related. I will fix it.

src/serialization/compact/SchemaService.ts
harunalpak: Because I could not use `Set.has` for objects.  And `ClientSendSchemaCodec.decodeResponse` return `Set<UUID>`.
Because of that I convert from it to `Set<string>`.

srknzl: You can use objects. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has#using_the_has_method

src/serialization/compact/SchemaService.ts
harunalpak: Fixed

src/serialization/compact/SchemaService.ts
harunalpak: Fixed

src/serialization/compact/SchemaService.ts
harunalpak: Fixed

src/serialization/compact/SchemaService.ts
srknzl: Turned out that we can't because SameValue semantics is used and the same object should be passed to the has() method, which is not possible

src/serialization/compact/SchemaService.ts
```diff
@@ -74,11 +89,19 @@ export class SchemaService {
...
+                    `The schema ${schema.typeName} cannot be replicated in the cluster, after ${this.maxPutRetryCount}  retries. 
+                    It might be the case that the client is connected to the two halves of the cluster that 
+                    is experiencing a split-brain, and continue putting the data associated with that schema might 
+                    result in data loss. It might be possible to replicate the schema after some time, when 
+                    the cluster is healed.`
+                );
+            } else {
+                this.putIfAbsent(schema);
+            }
+        }).catch((err) => { console.error(err); });
```
srknzl, line 104: We don't use console.error directly. We can either ignore the error completely, by `catch(() => {})` or if we are sure that the error handling of the promise returned by this method will be handled, we can skip putting a catch(). We can also log the error if we want via the logger.  

Right now, all the register errors are propagated to the user via async APIs. If the user doesn't register an error handler the program may crash if the user is using a new Node.js runtime. I think we can remove the catch block for this promise since we cannot continue without a schema. User should get an error.

src/serialization/compact/SchemaService.ts
```diff
@@ -106,4 +129,42 @@ export class SchemaService {
         const invocation = new Invocation(this.getInvocationService(), message);
         return this.getInvocationService().invokeUrgent(invocation).then(() => {});
     }
+
+    replicateSchemaInCluster(schema: Schema): Promise<boolean> {
+        const clientMessage = ClientSendSchemaCodec.encodeRequest(schema);
+        return this.retryMaxPutRetryCount(clientMessage, 0);
+    }
+
+    private retryMaxPutRetryCount(clientMessage: ClientMessage, index: number) : Promise<boolean> {
```
srknzl, line 138:
```suggestion
    private retryMaxPutRetryCount(clientMessage: ClientMessage, currentRetryCount: number) : Promise<boolean> {
```

or it can be `retryCount`. IMO, index is too general

src/serialization/compact/SchemaService.ts
```diff
@@ -106,4 +129,42 @@ export class SchemaService {
...
+                const replicatedMemberUuids = UuidUtil.convertUUIDSetToStringSet(ClientSendSchemaCodec.decodeResponse(response));
+                const members = this.getClusterService().getMembers();
+                for (const member of members) {
+                    if (!replicatedMemberUuids.has(member.uuid.toString())) {
+                        // There is a member in our member list that the schema
+                        // is not known to be replicated yet. We should retry
+                        // sending it in a random member.
+    
+                        return delayedPromise(this.retryPauseMillis)
+                        .then(() => {
```
srknzl, line 154:
```suggestion
                        return delayedPromise(this.retryPauseMillis).then(() => {
```

src/serialization/compact/SchemaService.ts
```diff
@@ -106,4 +129,42 @@ export class SchemaService {
...
+
+    replicateSchemaInCluster(schema: Schema): Promise<boolean> {
+        const clientMessage = ClientSendSchemaCodec.encodeRequest(schema);
+        return this.retryMaxPutRetryCount(clientMessage, 0);
+    }
+
+    private retryMaxPutRetryCount(clientMessage: ClientMessage, index: number) : Promise<boolean> {
+        if (index === this.maxPutRetryCount) {
+            return Promise.resolve(false);
+        } else {
```
srknzl, line 141: we can remove the else block here, not to indent the long code below.

src/serialization/compact/SchemaService.ts
```diff
@@ -74,11 +89,19 @@ export class SchemaService {
...
+                throw new IllegalStateError(
+                    `The schema ${schema.typeName} cannot be replicated in the cluster, after ${this.maxPutRetryCount}  retries. 
+                    It might be the case that the client is connected to the two halves of the cluster that 
+                    is experiencing a split-brain, and continue putting the data associated with that schema might 
+                    result in data loss. It might be possible to replicate the schema after some time, when 
+                    the cluster is healed.`
+                );
+            } else {
+                this.putIfAbsent(schema);
+            }
```
srknzl, line 103:
```suggestion
            }
            this.putIfAbsent(schema);
```

srknzl (changes requested):

srknzl:
@harunalpak Can you check if we can increase coverage easily by adding some tests? There are warnings regarding coverage

srknzl:
The coverage drop is 0.11%, which is not acceptable

src/serialization/compact/SchemaService.ts
harunalpak: I agree with you. I forgot to remove catch block. I only use for testing. Thanks for attention.

I removed catch block.

src/serialization/compact/SchemaService.ts
harunalpak: Fixed

src/serialization/compact/SchemaService.ts
harunalpak: Fixed

src/serialization/compact/SchemaService.ts
harunalpak: Fixed

src/serialization/compact/SchemaService.ts
harunalpak: Fixed

harunalpak:
> The coverage drop is 0.11%, which is not acceptable

I added a new test. And code coverage increased now.

test/integration/backward_compatible/serial/CompactRestartTest.js
```diff
@@ -98,4 +101,28 @@ describe('CompactRestartTest', function() {
         const values = await map.valuesWithPredicate(Predicates.sql('INT32 == 42'));
         values.size().should.be.eq(1);
     });
+
+    it('should throw IllegalStateError exception, because of schema could not be replicated in the cluster', async function() {
```
srknzl, line 105: I think the test is ok, though we should put it in another file. `test/integration/backward_compatible/parallel/serialization/compact/CompactTest.js` should be better. Don't forget to skip the test if the client version is older than 5.2.0 because we added this behavior in 5.2.0.

srknzl (changes requested):
One last thing:

srknzl (approved):

PR #1432, opened 2022-12-08
Logging does not work if written in lowercase(hazelcast.logging.level) [API-1740]

harunalpak:
This is simple change, I only add toUpperCase().

I am not sure we need extra test for this. I added a sample test for logging.


srknzl (changes requested):

srknzl:
Can you update the documentation as well? https://github.com/srknzl/hazelcast-nodejs-client/blob/192b7be3059916f33e1c31c10612bdd54364498e/DOCUMENTATION.md#L654

code_samples/logging_sample.js
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const { Client } = require('hazelcast-client');
+
+// This sample code demonstrates logging. We don't need to do any operations.
```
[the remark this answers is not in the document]
srknzl: Also can we say that the DEBUG is can be any case. DeBuG or debug etc.

code_samples/logging_level_sample.js
```diff
@@ -17,11 +17,11 @@
 
 const { Client } = require('hazelcast-client');
 
-// This sample code demonstrates logging. We don't need to do any operations.
+// This sample code demonstrates configuring logging level. We don't need to do any operations.
 (async () => {
     const client = await Client.newHazelcastClient({
         properties: {
-            'hazelcast.logging.level': 'DEBUG'
+            'hazelcast.logging.level': 'DeBuG'
```
srknzl, line 24: Can we add a comment saying that this string is case-insensitive? Then it can be 'DEBUG' if you want. I meant to add a comment about it.

harunalpak: I added a comment for this value.

srknzl (dismissed):

srknzl (approved):

PR #1435, opened 2022-12-09
Fix backward compatibility test v5.2.0

harunalpak:
Related to https://github.com/hazelcast/hazelcast-nodejs-client/pull/1418
this PR we added a new Codec named "SetUUIDCodec".

And this is used in SchemaService.put method. 

```
replicateSchemaInCluster(schema: Schema): Promise<boolean> {
        const clientMessage = ClientSendSchemaCodec.encodeRequest(schema);
        return this.retryMaxPutRetryCount(clientMessage, 0);
    }

private retryMaxPutRetryCount(clientMessage: ClientMessage, currentRetryCount: number) : Promise<boolean> {
    if (currentRetryCount === this.maxPutRetryCount) {
        return Promise.resolve(false);
    }
    const invocationService = this.getInvocationService(); 
    const invocation = new Invocation(invocationService, clientMessage);
    return invocationService.invoke(invocation).then((response) => {
        const replicatedMemberUuids = UuidUtil.convertUUIDSetToStringSet(ClientSendSchemaCodec.**decodeResponse**(response));
        const members = this.getClusterService().getMembers();
        for (const member of members) {
            if (!replicatedMemberUuids.has(member.uuid.toString())) {
                // There is a member in our member list that the schema
                // is not known to be replicated yet. We should retry
                // sending it in a random member.
                return delayedPromise(this.retryPauseMillis).then(() => {
                    // correlation id will be set when the invoke method is
                    // called above
                    clientMessage = clientMessage.copyMessageWithSharedNonInitialFrames();

                    return this.retryMaxPutRetryCount(clientMessage, ++currentRetryCount);
                })
            }
        }

        // All members in our member list all known to have the schema
        return Promise.resolve(true);
    })
    
}
```

And I check that it is also added to server version 5.2.0. (https://github.com/hazelcast/hazelcast/pull/22197)

IMO 'CompactTest' should run greater than 5.2.0 server version.


test/integration/backward_compatible/parallel/serialization/compact/CompactTest.js
```diff
@@ -78,7 +78,7 @@ describe('CompactTest', function () {
     before(async function () {
         TestUtil.markClientVersionAtLeast(this, '5.1.0');
         const comparisonValueForServerVersion520 = await TestUtil.compareServerVersionWithRC(RC, '5.2.0');
-        if ((await TestUtil.compareServerVersionWithRC(RC, '5.1.0')) < 0) {
+        if ((await TestUtil.compareServerVersionWithRC(RC, '5.2.0')) < 0) {
```
mdumandag, line 81: I think this should stay like the previous version, and we should add a check below. Something like 
```
        // Compact serialization 5.2 and newer clients are not compatible with servers older than 5.2
        if (comparisonValueForServerVersion520 < 0 && TestUtil.isClientVersionAtLeast('5.2.0')) {
            this.skip();
        }
```

srknzl:
@mdumandag This means that compact won't work with 5.1.0 server. So it is a breaking change. Is this expected?

test/integration/backward_compatible/parallel/serialization/compact/CompactTest.js
harunalpak: Fixed

srknzl:
approving, discussed via slack. It is expected that we break compatiblity

test/integration/backward_compatible/parallel/serialization/compact/CompactTest.js
```diff
@@ -78,7 +78,8 @@ describe('CompactTest', function () {
     before(async function () {
         TestUtil.markClientVersionAtLeast(this, '5.1.0');
         const comparisonValueForServerVersion520 = await TestUtil.compareServerVersionWithRC(RC, '5.2.0');
-        if ((await TestUtil.compareServerVersionWithRC(RC, '5.1.0')) < 0) {
+        // Compact serialization 5.2 and newer clients are not compatible with servers older than 5.2
+        if (comparisonValueForServerVersion520 < 0 && TestUtil.isClientVersionAtLeast('5.2.0')) {
```
srknzl, line 82: I think we should assert client and server should be >= 5.2.0, otherwise skip

test/integration/backward_compatible/parallel/serialization/compact/CompactTest.js
```diff
@@ -78,7 +78,8 @@ describe('CompactTest', function () {
     before(async function () {
```
srknzl, line 78: Should we update other failing tests like SQL tests about compact etc.?

srknzl: Can you check all failing tests one by one? You need to be careful with tests with the same suite name / file name.

srknzl (changes requested):

test/integration/backward_compatible/parallel/serialization/compact/CompactTest.js
mdumandag: That is what I had in my mind, which would simplify a lot of things. The only downside is that we won't be able to test 5.1.z server with a 5.1.z client regarding Compact serialization when releasing a new version of either of those. I believe it is totally OK, as the Compact was BETA in those versions anyways

srknzl: Then, we can assert either they are both above or equal 5.2.0, or older than 5.2.0.

test/integration/backward_compatible/serial/CompactRestartTest.js
```diff
@@ -50,11 +50,14 @@ describe('CompactRestartTest', function() {
...
-        // Compact serialization 5.2 server is not compatible with clients older than 5.2
-        if (comparisonValueForServerVersion520 >= 0 && !TestUtil.isClientVersionAtLeast('5.2.0')) {
+        const isClientVersionAtLeast520 = TestUtil.isClientVersionAtLeast('5.2.0');
+        // Compact serialization 5.2 and newer server is not compatible with clients older than 5.2
+        // Compact serialization 5.2 and newer clients are not compatible with servers older than 5.2
+        const isNotCompactCompatible =
+            (comparisonValueForServerVersion520 >= 0 && !isClientVersionAtLeast520) ||
+            (comparisonValueForServerVersion520 < 0 && isClientVersionAtLeast520);
+
+        if (isNotCompactCompatible) {
```
srknzl, line 60: Can you put this logic in TestUtil in a function? I think the name isCompactCompatible should be ok.

srknzl (approved):

PR #1442, opened 2022-12-13
Change logic around closing connections and writers API-1283 (redux)

### LINKED ISSUE #1256: The error: `Cannot add property 0, object is not extensible`

srknzl:
Clean port of #1417 

Already approved in https://github.com/hazelcast/hazelcast-nodejs-client/pull/1417

Will merge with one approval because it is already approved.

--- 

fixes #1256 

The issue was writing some data to the connection after it being closed. I changed and refactored some logic around PipelinedWriter and Connection to avoid that. 

- Moved socket closing responsibility to Writers
- Fixed a test in PipelinedWriterTest
- Ongoing and future socket.write calls will end with an error after the writer is closed. This is because close() destroys the socket and any ongoing and future socket.write calls will end with an error as described in node.js [documentation](https://nodejs.org/api/stream.html#writabledestroyerror). So we will be able to reject all deferred promises in the write queue upon close due to connection.close() or socket write error close().
- We were not closing the socket on write error before, now we do it. We were sending end packet and don't write anymore. This is the same as java. [See](https://hazelcast.slack.com/archives/C01JU7ZJYGP/p1668695419424179)


mdumandag (approved):

harunalpak (approved):
This issue has already been approved

PR #1471, opened 2023-02-02
Update Readme and package.json about hazelcast and node.js client

srknzl:
I wanted to update package.json only at first, but then decided to update README as well. The What is hazelcast part of readme is taken from current master of Hazelcast repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "hazelcast-client",
     "version": "5.2.0",
-    "description": "Hazelcast - open source In-Memory Data Grid - client for Node.js",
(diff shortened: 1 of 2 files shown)


yuce (approved):
Looks good!

package.json
```diff
@@ -72,11 +72,12 @@
     },
     "keywords": [
         "hazelcast",
-        "nodejs",
+        "real-time",
```
mdumandag, line 75: minor: is it `real-time` or `real time`, as we have used the latter above, and the former here

srknzl: I used real-time because it was real-time in hazelcast.com. I will make both real-time

mdumandag (approved):

PR #1518, opened 2024-02-16
Update certificate file

srknzl:
Updates the cert file via https://github.com/hazelcast/private-test-artifacts/blob/data/certs.jar


emreyigit (approved):

srknzl:
the link checker fails with 403 on a non-broken link, safe to ignore (retried and failed again)

PR #1523, opened 2024-05-09
Update dependencies, linting, server upgrade & workflows

srknzl:
Prepare for 5.4 release

- update dependencies 
- fix linting problems due to upgrading eslint
- update hazelcast version 5.4.0 from 5.4.0-snapshot
- Fix workflows that depend on node 10 to compile the code. We now compile in latest lts but run tests in node 10. 

**NOTE:** Tests will fail in this PR because the workflow is taken from the master branch. It fails due to upgrading our development dependencies. We need to force-merge this PR and then it'll run on master and we'll know if any tests fail with the new workflow. The tests passed on my local.


src/serialization/Portable.ts
```diff
@@ -37,6 +37,7 @@ export enum FieldType {
     FLOAT = 7,
     DOUBLE = 8,
     UTF = 9, // Defined for backwards compatibility.
+    /* eslint-disable @typescript-eslint/no-duplicate-enum-values */
```
ihsandemir, line 40: why is this comment?

srknzl:
closed as not prioritized

PR #1524, opened 2024-05-30
Use released Hazelcast `5.4.0` instead of `SNAPSHOT`

JackPGreen:
Now `5.4.0` is released, the released version should be used in preferences to a pre-release `SNAPSHOT`.


JackPGreen:
@srknzl are you able to take a look at this, please?

srknzl (approved):

srknzl (commented):
we should ideally use 5.5.0-SNAPSHOT btw, but does not matter much since no development is done for this client for now

PR #1526, opened 2024-07-19
Add&use markCPAvailable

srknzl:
5.5 OS removed CP. Exclude CP compat tests from running against 5.5 and above OS.


srknzl:
we don't need to merge this anymore because Node.js client is removed client compat suites.

